PR #19, opened 2019-12-04
Provide access to fetched signed metadata

### LINKED ISSUE #15: Provide access to the fetched metadata
Add `root`, `snapshot`, and `timestamp` methods to `Repository` to return references to the underlying signed metadata, and change `targets` to do the same, to permit consumers to make small changes to the metadata and re-sign it (such as adding a single target).

zmrow:
This commit makes an API-breaking change to the `tough` library. It adds
`root`, `snapshot`, and `timestamp` methods to `Repository` which return
references to the signed metadata. The `targets` method for `Repository`
has been updated to do the same.

`DownloadArgs.run()` in `tuftool` has been updated to reflect the
aforementioned change to `Repository.targets()`.

*Issue #, if available:*
Fixes #15 

*Description of changes:*
* Add `root`, `snapshot`, and `timestamp` methods to `Repository`
* Change `targets` method to return reference to signed metadata (`Signed<crate::schema::Targets`)
* Bump `tough` version to `0.2.0`
* Fix one breaking test
* Update `tuftool` to make use of the new `Repository.targets()` return value

*Testing done:*
* All unit tests pass
* `clippy` is only warning about things this PR doesn't touch 🤷‍♂ 

### DIFF
--- a/tough/src/lib.rs
+++ b/tough/src/lib.rs
@@ -33,7 +33,6 @@ use crate::schema::{Role, RoleType, Root, Signed, Snapshot, Timestamp};
 use chrono::{DateTime, Utc};
 use snafu::{ensure, OptionExt, ResultExt};
 use std::borrow::Cow;
-use std::collections::HashMap;
 use std::io::Read;
 use std::path::Path;
 use url::Url;
@@ -115,8 +114,11 @@ pub struct Repository<'a, T: Transport> {
     datastore: Datastore<'a>,
     earliest_expiration: DateTime<Utc>,
     earliest_expiration_role: RoleType,
+    root: Signed<Root>,
+    snapshot: Signed<Snapshot>,
+    timestamp: Signed<Timestamp>,
+    targets: Signed<crate::schema::Targets>,
     target_base_url: Url,
-    targets: HashMap<String, Target>,
 }
 
 impl<'a, T: Transport> Repository<'a, T> {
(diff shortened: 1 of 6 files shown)


tough/src/lib.rs
```diff
@@ -115,8 +115,11 @@ pub struct Repository<'a, T: Transport> {
     datastore: Datastore<'a>,
     earliest_expiration: DateTime<Utc>,
     earliest_expiration_role: RoleType,
+    root: Signed<Root>,
+    snapshot: Signed<Snapshot>,
+    timestamp: Signed<Timestamp>,
     target_base_url: Url,
-    targets: HashMap<String, Target>,
+    targets: Signed<crate::schema::Targets>,
```
iliana, line 120: nit: reorder these

tough/src/lib.rs
```diff
@@ -193,21 +196,34 @@ impl<'a, T: Transport> Repository<'a, T> {
...
+            snapshot,
+            timestamp,
             target_base_url,
-            targets: targets
-                .signed
-                .targets
-                .into_iter()
-                .map(|(key, value)| (key, value.into()))
-                .collect(),
+            targets,
```
iliana, line 201: (and here)

iliana (changes requested):
Can we get rid of `tough::Target` and just `use crate::schema::Targets` now that we've removed its use in `Repository`?

zmrow:
> Can we get rid of `tough::Target` and just `use crate::schema::Targets` now that we've removed its use in `Repository`?

Oooh - indeed we can. Will do!

zmrow:
Addressed @iliana 's comments in the above force-push

iliana (approved):

PR #55, opened 2020-02-10
bump reqwest version to 0.10.1

webern:
*Issue #, if available:*

N/A

*Description of changes:*

In another project we experienced failures using http 1.1 over TLS. This was tracked to `hyper` where we believe we had a slightly different manifestation of this issue [hyper 1439](https://github.com/hyperium/hyper/issues/1439) (in our case a closed connection was being used leading to a "Failed to fetch" error).

The `hyper` was fixed in `v0.13` which is obtained by bumping `reqwest` from `0.9` to `0.10`.

In `reqwest` 0.10 the non-async api is relegated to the `reqwest::blocking` module. This is breaking for customers of `tough` since `pub type HttpTransport = reqwest::Client;` becomes `pub type HttpTransport = reqwest::blocking::Client;`. However the code changes required to accommodate this are as simple as adding `::blocking::` to the path wherever the compiler complains.

This change was tested in an update program that uses `tough` and was found to fix the issue.


iliana (approved):

zmrow (approved):
LGTM

🏋

Issue #65, opened 2020-02-27
Rename target-update-url to targets-update-url


webern:
Using the plural 'targets' is more consistent with the TUF terminology.

PR #68, opened 2020-03-03
rename target_base_url to targets_base_url

webern:
*Issue #, if available:*

#65 

*Description of changes:*

`targets_base_url` is a better name than `target_base_url` because it is more consistent with the naming throughout tough, tuftool and the tuf spec. 

*Testing Done*
`cargo test` at the workspace level.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1969,7 +1969,7 @@ dependencies = [
 
 [[package]]
 name = "tough"
-version = "0.4.0"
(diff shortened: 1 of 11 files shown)


iliana:
Let's bump minor versions in {tough/tuftool}/Cargo.toml, just so we make sure we don't forget. :)

webern:
Pushed version bumps.

iliana (approved):

Issue #61, opened 2020-02-13
License files are not included in published crates


iliana:
The LICENSE-APACHE and LICENSE-MIT files need to be included in the published crates, but aren't because they live in the parent directory.

PR #69, opened 2020-03-03
copy license files to published crates

webern:
*Issue #, if available:*

#61 

*Description of changes:*

License files were not included in published crates, so copied the top-level license files to each of the crates.


iliana (approved):
I wish there was a better way to do this but this is fine.

Issue #12, opened 2019-11-25
HTTP retries


tjkirch:
On connection resets, the `Read`er should retry the HTTP connection for the rest of the range (similar to wget’s default behavior).

webern:
Given some recent issues, we will prioritize this. However, reqwest [doesn't offer a suitable interface for this](https://github.com/seanmonstar/reqwest/issues/316):

> retry the HTTP connection for the rest of the range (similar to wget’s default behavior).

But we can at least start over from the beginning, which is what I will implement quickly.

iliana:
We return an `impl Read` — if we start over and then continue sending those bytes to the reader, the data is corrupt.

We can keep a counter of bytes read so far and set [a `Range` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Range) pretty trivially. We would have to use a wrapper struct around a `reqwest::Response`. We could also keep a counter of bytes and skip the first `n` bytes on the resulting Read stream.

Either way, we need a wrapper struct to *correctly* impl Read while possibly re-sending HTTP requests on retriable errors.

PR #72, opened 2020-03-06
test http transport happy case

webern:
*Issue #, if available:*

Related to #12 as it establishes a method by which we can test http transport.

*Description of changes:*

Uses the [`mockito` crate](https://docs.rs/mockito/0.23.3/mockito/) to mock an http server allowing us to test a repo that uses http transport.

*Testing Done*

`cargo test` for the workspace.
`cargo test --all-features` in the `tough` create.

Unfortunately these two types of testing can not be done in one shot, so I added `cd tough && cargo test --all-features` to the CI run.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -14,3 +14,4 @@ jobs:
     - run: cargo build --locked -p tough
     - run: cargo build --locked -p tuftool
     - run: cargo test --locked
+    - run: cd tough && cargo test --all-features --locked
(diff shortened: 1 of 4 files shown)


.github/workflows/rust.yml
```diff
@@ -14,3 +14,4 @@ jobs:
     - run: cargo build --locked -p tough
     - run: cargo build --locked -p tuftool
     - run: cargo test --locked
+    - run: cd tough && cargo test --all-features --locked
```
iliana, line 17:
```suggestion
    - run: cargo test -p tough --all-features --locked
```

iliana (approved):

PR #78, opened 2020-03-16
Rusoto 0.43

### LINKED ISSUE #77: Update Rusoto to 0.43
Seems there is a breaking change based on dependebots PRs. #74 #75 #76

webern:
*Issue #, if available:*

Fixes #77 

*Description of changes:*

Rusoto 0.43 changes the async interface. Retained the behavior of tough by using `block_on` instead of `.sync()`.

*Testing*

Created a new Bottlerocket repo using `tuftool create`, then used that repo to downgrade a Bottlerocket node from 0.3.1 to 0.3.0. It worked.

Also ran all of what CI does. Discovered an issue with the imports in the `interop.rs` test file and fixed it.

### DIFF
--- a/tuftool/Cargo.toml
+++ b/tuftool/Cargo.toml
@@ -23,9 +23,9 @@ pem = "0.7.0"
 rayon = "1.2"
 reqwest = { version = "0.10.4", features = ["blocking"] }
 ring = { version = "0.16.7", features = ["std"] }
-rusoto_core = { version = "0.42", optional = true, default-features = false }
-rusoto_credential = { version = "0.42", optional = true }
-rusoto_ssm = { version = "0.42", optional = true, default-features = false }
+rusoto_core = { version = "0.43", optional = true, default-features = false }
+rusoto_credential = { version = "0.43", optional = true }
+rusoto_ssm = { version = "0.43", optional = true, default-features = false }
 serde = "1.0.99"
 serde_json = "1.0.48"
 snafu = { version = "0.6.0", features = ["backtraces-impl-backtrace-crate"] }
@@ -34,6 +34,7 @@ tempfile = "3.1.0"
 url = "2.1.0"
 walkdir = "2.2.9"
 tough = { version = "0.5.0", path = "../tough", features = ["http"] }
+tokio = "0.2.13"
 
 [dev-dependencies]
 assert_cmd = "0.12"
--- a/tuftool/src/source.rs
+++ b/tuftool/src/source.rs
@@ -16,6 +16,9 @@ use tough::schema::key::Key;
 use tough::sign::{parse_keypair, Sign};
 use url::Url;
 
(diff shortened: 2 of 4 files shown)


iliana:
Can you test signing with an `aws-ssm://` key?

tuftool/src/source.rs
```diff
@@ -58,16 +59,14 @@ impl KeySource {
...
-                    .sync()
-                    .context(error::SsmGetParameter {
-                        profile: profile.clone(),
-                        parameter_name,
-                    })?;
+                let response_fut = ssm_client.get_parameter(rusoto_ssm::GetParameterRequest {
+                    name: parameter_name.to_owned(),
+                    with_decryption: Some(true),
+                });
+                let response = block_on(response_fut).context(error::SsmGetParameter {
```
zmrow: I'm not super familiar with futures, but does this `block_on` call (or the underlying `ssm::GetParameterRequest`) have some kind of timeout?

webern:
Wow, this solution doesn't work: `thread 'main' panicked at 'not currently running on the Tokio runtime.', src/libcore/option.rs:1188:5`

Searching brought me to here: https://github.com/tokio-rs/tokio/issues/2289

Currently not clear how to proceed. Researching...

tuftool/src/source.rs
webern: No, block_on will run forever afaik.

webern:
Updated. Using an async function from within a non-async function is more involved than I thought and requires a runtime (or, i suppose you could roll-your-own polling loop, but what I've done here seems to be the suggested way to do this).

tuftool/src/source.rs
```diff
@@ -53,34 +56,38 @@ impl KeySource {
...
+        use rusoto_ssm::Ssm;
+
+        let ssm_client = crate::ssm::build_client(profile.deref_shim())?;
+        let fut = ssm_client.get_parameter(rusoto_ssm::GetParameterRequest {
+            name: parameter_name.to_owned(),
+            with_decryption: Some(true),
+        });
+        let response = tokio::runtime::Runtime::new()
+            .unwrap()
+            .block_on(async move { fut.await })
```
iliana: I think because `fut` is already a future, you can just do `block_on(fut)`.

tuftool/Cargo.toml
```diff
@@ -34,6 +34,7 @@ tempfile = "3.1.0"
 url = "2.1.0"
 walkdir = "2.2.9"
 tough = { version = "0.5.0", path = "../tough", features = ["http"] }
+tokio = { version = "0.2.13", features = ["macros"] }
```
iliana: You can probably drop `macros` here and not specify any features, since you're not using `#[tokio::main]`.

iliana (approved):

zmrow (approved):
lgtm!

💎

Issue #43, opened 2020-01-15
Move functionality out of tuftool into a library


zmrow:
When it comes to interacting and updating TUF repos, quite a bit of useful functionality is still contained in `tuftool` and isn't usable by outside code. This functionality should be put in a library.

Right off the bat I can see a few functions that should be split out:
* `tuftool/src/create.rs : write_metadata()` : This function takes a role, computes its SHA/length, signs it, and writes it to the TUF output directory
* `tuftool/src/create.rs : process_target()` : Given a `&Path`, this function computes its SHA/length, builds a `Target` object, and writes the file to the appropriate TUF output directory.
*  `tuftool/src/source.rs : KeySource.read()` : Fetching / parsing signing keys is necessary to sign a repo.
* `tuftool/src/key.rs: keys_for_root()` : Given a signing key, get the appropriate root keys.
* `tuftool/src/key.rs : sign_metadata_inner()` : This function does some crucial parsing `root.json` to get the proper keys to use to actually sign.

webern:
@iliana, does this need to wait until you are done with #80, or do you think it is something I could work on now?

iliana:
@webern this can happen whenever, so now works :)

zmrow:
I think the most logical place to start is to extract some of the basic functionality out of tuftool.

This is what I came up with to start. I think most of this can live in `tough`.
* **Create new TUF repo** (`tough/create.rs`)
  * Add a `new()` method to `tough::Repository` that takes in some kind of `CreateArgs` struct and returns a `Repository`.
  * Much of the rest can be lifted almost verbatim from `tuftool`.
    * The `create::CreateArgs` struct becomes a regular struct with sane `Default`s that gets passed to the `new()` call above.
    * The logic currently in `create::CreateProcess.run()` can be adapted for the `Repository::new()` method.

* **Update an existing TUF repo with new targets**
  * Add an `update(&self) -> tough::Repository` method to `tough::Repository`. It will take in an `UpdateArgs` struct and return a brand new `Repository`. (It could optionally write this new Repository's files at a supplied location?)
    * The `UpdateArgs` struct would contain a list of new targets, metadata versions, and expiration dates. Under the hood it would add the new targets, update and sign the metadata.

* **Refresh an existing TUF repo's timestamps**
  * Add a `refresh(&mut self)` method to `tough::Repository` that takes in a `RefreshArgs` struct. (Can be lifted almost verbatim from `tuftool`)
[...]

PR #103, opened 2020-04-17
Implement `KeySource` trait and make corresponding updates to `tough` and `tuftool`

zmrow:
*Issue #, if available:*
Related to #43 

*Description of changes:*
This is the initial step towards factoring quite a bit of logic *out* of `tuftool` and into `tough` where it can be more widely used. The `KeySource` enum was pretty central to `tuftool` and I thought it made more sense to implement this as a trait. This also allows users to add additional key sources.

I think when reviewing, it makes sense to review the commits in logical order:
```
commit 88fc5147f1a0f468cf8814e2280b792149caa4cd
Author: Zac Mrowicki <mrowicki@amazon.com>
Date:   Wed Apr 15 18:29:47 2020 +0000

    Initial commit of KeySource trait

    This commit adds the KeySource trait to tough. It is meant to
    replace the KeySource enum in tuftool and allows for users to
    implement their own sources of Keys.

    Initially we implement this trait for local files.

commit e2f13132275365009e4cf2ddafa1465dc4f2ee9e
Author: Zac Mrowicki <mrowicki@amazon.com>
Date:   Fri Apr 17 21:13:44 2020 +0000

    Adds the tough-ssm crate

    This crate is meant to implement the KeySource trait for keys
    that live in AWS SSM.

commit c838277ca06863f7c3265532e763eb6660d1891a (HEAD -> keysource-trait, origin/keysource-trait)
Author: Zac Mrowicki <mrowicki@amazon.com>
Date:   Fri Apr 17 21:20:45 2020 +0000

    tuftool: Remove KeySource enum, use tough-ssm and KeySource trait

    This is a decent set of changes to tuftool. It fully removes the
    KeySource enum in source.rs and instead uses the KeySource trait
    from tough.

    This commit also removes all SSM related code in favor of the
    tough-ssm crate (which is a copy of this code).

    It also removes the deref.rs file, as it is no longer need because
    these features have stabilized in upstream Rust. See:
    https://github.com/rust-lang/rust/issues/50264
    and
    https://github.com/rust-lang/rust/pull/64708
```

*Testing done:*
* All unit tests continue to pass

I haven't manually run all commands available in `tuftool`, but ran the below to make sure core functionality works:
* I'm able to download a repo
```
cargo run -- download ./foo --metadata-url https://updates.bottlerocket.aws/2020-02-02/aws-k8s-1.15/x86_64/ --target-url https://updates.bottlerocket.aws/targets/ --root ~/bottlerocket/bottlerocket/packages/os/root.json 
   Compiling tough v0.5.0 (/home/mrowicki/tough/tough/tough)
...
Downloading targets to "./foo"
        -> bottlerocket-aws-k8s-1.15-x86_64-0.3.1-8a0c0b3-root.ext4.lz4
        -> bottlerocket-aws-k8s-1.15-x86_64-0.3.0-faaec6e4-root.ext4.lz4
        -> bottlerocket-aws-k8s-1.15-x86_64-0.3.0-faaec6e4-boot.ext4.lz4
        -> bottlerocket-aws-k8s-1.15-x86_64-0.3.1-8a0c0b3-root.verity.lz4
        -> manifest.json
        -> bottlerocket-aws-k8s-1.15-x86_64-0.3.0-faaec6e4-root.verity.lz4
        -> bottlerocket-aws-k8s-1.15-x86_64-0.3.1-8a0c0b3-boot.ext4.lz4
```
* I'm able to create a repo (redacted the actual name of the key)
```
$ cargo run -- create ./foo ./out -k 'aws-ssm:///REDACTED' --root /home/mrowicki/bottlerocket/bottlerocket/packages/os/root.json --targets-expires 'in 2 weeks' --targets-version $(date +%s) --snapshot-expires 'in 2 weeks' --snapshot-version $(date +%s) --timestamp-expires 'in 5 days' --timestamp-version $(date +%s)
    Finished dev [unoptimized + debuginfo] target(s) in 0.10s
...

$ tree foo/
foo/
├── bottlerocket-aws-k8s-1.15-x86_64-0.3.0-faaec6e4-boot.ext4.lz4
├── bottlerocket-aws-k8s-1.15-x86_64-0.3.0-faaec6e4-root.ext4.lz4
├── bottlerocket-aws-k8s-1.15-x86_64-0.3.0-faaec6e4-root.verity.lz4
├── bottlerocket-aws-k8s-1.15-x86_64-0.3.1-8a0c0b3-boot.ext4.lz4
├── bottlerocket-aws-k8s-1.15-x86_64-0.3.1-8a0c0b3-root.ext4.lz4
├── bottlerocket-aws-k8s-1.15-x86_64-0.3.1-8a0c0b3-root.verity.lz4
└── manifest.json

0 directories, 7 files

$ tree out/
out/
├── metadata
│   ├── 1587159947.snapshot.json
│   ├── 1587159947.targets.json
│   ├── 1.root.json
│   └── timestamp.json
└── targets
    ├── 2a7def4e4c64d20af17dea3c65934be901ef6f0b01bf517428ea7f5dfa211cb1.bottlerocket-aws-k8s-1.15-x86_64-0.3.0-faaec6e4-boot.ext4.lz4
    ├── 43ca01b101a6d9b7aa241255870862419b886731051f659ccc420ca431ce1163.bottlerocket-aws-k8s-1.15-x86_64-0.3.1-8a0c0b3-root.ext4.lz4
    ├── 88ea21326e148b54e225c14a3382a1db2e3f21989a6880151eed719c79194a27.bottlerocket-aws-k8s-1.15-x86_64-0.3.0-faaec6e4-root.verity.lz4
    ├── ae0685346a531a698e67df5c999d684f6a9d2ff55a7f1120c5cd02fad98a1449.bottlerocket-aws-k8s-1.15-x86_64-0.3.1-8a0c0b3-root.verity.lz4
    ├── c44124b5199a1e17af81a3ba4cacca2a48ddab13f2bb8d93dc1f8b19c5c87c22.manifest.json
    ├── dd9646d4dd2de050bde89f73a05db8d48d88f521e54af77db0a8d7831b67d1b1.bottlerocket-aws-k8s-1.15-x86_64-0.3.0-faaec6e4-root.ext4.lz4
    └── df2fcebb46864096d7fdacea19796664935056bcc06c488b7a54153e1df784ca.bottlerocket-aws-k8s-1.15-x86_64-0.3.1-8a0c0b3-boot.ext4.lz4

2 directories, 11 files
```

* Tried a few of the `root` commands
```
$ cat root.json | grep -i '"version":'
    "version": 1,

$ cargo run -- root bump-version ./root.json 
    Finished dev [unoptimized + debuginfo] target(s) in 0.10s
 ...

$ cat root.json | grep -i '"version":'
    "version": 2,

$ cargo run -- root init ./initroot.json
    Finished dev [unoptimized + debuginfo] target(s) in 0.09s
...

$ cat initroot.json 
{
  "signed": {
    "_type": "root",
    "spec_version": "1.0.0",
    "consistent_snapshot": true,
    "version": 1,
    "expires": "2020-04-17T21:56:39Z",
    "keys": {},
    "roles": {
      "snapshot": {
        "keyids": [],
        "threshold": 1507
      },
      "root": {
        "keyids": [],
        "threshold": 1507
      },
      "targets": {
        "keyids": [],
        "threshold": 1507
      },
      "timestamp": {
        "keyids": [],
        "threshold": 1507
      }
    }
  },
  "signatures": []
}
```


Cargo.toml
```diff
@@ -2,5 +2,6 @@
 members = [
     "olpc-cjson",
     "tough",
+    "tough-ssm",
```
tjkirch, line 5: @zmrow and @iliana, what do we think are the pros and cons of a subcrate vs. a feature?

webern: I don't think they are mutually exclusive. Having a subcrate could make the feature cleaner by excluding the entire subcrate.

tuftool/src/source.rs
```diff
@@ -8,169 +8,51 @@
...
+            path: PathBuf::from(url.path()),
+        })),
+        #[cfg(any(feature = "rusoto-native-tls", feature = "rusoto-rustls"))]
+        "aws-ssm" => Ok(Box::new(SsmKeySource {
+            profile: url.host_str().and_then(|s| {
+                if s.is_empty() {
+                    None
+                } else {
+                    Some(s.to_owned())
+                }
```
webern, line 70: A possible solution would be to use a strategy pattern. A collection of `Bpx<dyn KeySource>>`s could be held. Each key source could be asked `keysrc->do_you_handle_this_url_scheme()` and the first one that says yes could be used. This would be a strategy pattern commonly used, for example, for codecs I believe. Maybe it's too OO for Rust or for our tastes, but it's an idea that would make something like this easily extensible instead of using match arms.

edit: this approach would have the benefit of only needing a `#[cfg` around the code that would `push` a specific type of key source into the collection.

webern (commented):
This seems like a move in the right direction to me, though I'm not fully convinced that we have the key_source as extensible as we like. In any event an edit is required since clippy is angry.

Cargo.toml
tjkirch: Sure, but maybe only cleaner for the developer and not the user?  The user has to add another dependency and use a different namespace to refer to that key source, when I think often that's accomplished with a feature.

tuftool/src/source.rs
zmrow: I'm happy to be told otherwise by the Rust experts, but I believe match arms are pretty idiomatic for Rust.

Given that this function performs actions that are fairly specific to `tuftool` (getting a `KeySource` from text), I'm also not convinced that it's something that belongs in the trait.  (I also come from a very OO background so I can absolutely see it your way as well 😄  )

Cargo.toml
zmrow: The strategy behind the trait is to make it easier for developers to use their own providers of keys. Implementing the SSM crate provides a good reference and example for developers and means `tough` isn't tied to any AWS-specific implementation, feature or otherwise.

zmrow:
Fix a few of `clippy`'s issues.

I'm not sure if I can get around the `&Box<...>` pedantic issue that `clippy` is angry about, so I'm allowing it for now.

tough-ssm/src/lib.rs
```diff
@@ -0,0 +1,89 @@
...
+pub struct SsmKeySource {
+    pub profile: Option<String>,
+    pub parameter_name: String,
+    pub key_id: Option<String>,
+}
+
+impl SsmKeySource {
+    /// Read a key from SSM.
+    fn read(&self) -> Result<Vec<u8>> {
+        let ssm_client = client::build_client(self.profile.as_deref())?;
```
tjkirch: `build_client` can't be called concurrently, but I don't think it's obvious to any user of `SsmKeySource` -- or of `tough` with a user-supplied `KeySource` -- that the restriction applies to them.  You'd have to dig through several levels to find it.  Is there any way we can make that more obvious so someone doesn't shoot themself in the foot?

In other words, I wonder if this move raises the priority of the rusoto_credential fix..?

tough-ssm/src/lib.rs
```diff
@@ -0,0 +1,89 @@
...
+        let fut = ssm_client.get_parameter(rusoto_ssm::GetParameterRequest {
+            name: self.parameter_name.to_owned(),
+            with_decryption: Some(true),
+        });
+        let response = tokio::runtime::Runtime::new()
+            .unwrap()
```
tjkirch: Can we make runtime failures an error instead of a panic?

tough-ssm/src/error.rs
```diff
@@ -0,0 +1,68 @@
...
+
+    #[snafu(display("Error creating AWS request dispatcher: {}", source))]
+    RusotoTls {
+        source: rusoto_core::request::TlsError,
+        backtrace: Backtrace,
+    },
+
+    #[snafu(display(
+        "Failed to get aws-ssm://{}{}: {}",
+        profile.as_deref().unwrap_or(""),
```
tjkirch, line 52: Does the profile include the region?  Region would be useful in some of these errors.

tough-ssm/src/error.rs
```diff
@@ -0,0 +1,68 @@
...
+    SsmGetParameter {
+        profile: Option<String>,
+        parameter_name: String,
+        source: rusoto_core::RusotoError<rusoto_ssm::GetParameterError>,
+        backtrace: Backtrace,
+    },
+
+    #[snafu(display("Missing field in SSM response: {}", field))]
+    SsmMissingField {
+        field: &'static str,
```
tjkirch: I think it'd help to include the region and parameter.

tough-ssm/src/lib.rs
```diff
@@ -0,0 +1,89 @@
...
+
+/// Implements the KeySource trait.
+impl KeySource for SsmKeySource {
+    fn as_sign(
+        &self,
+    ) -> std::result::Result<Box<dyn Sign>, Box<dyn std::error::Error + Send + Sync + 'static>>
+    {
+        //fn as_sign(&self) -> std::result::Result<Box<dyn Sign>, Self::Error> {
+        let data = &self.read()?;
+        //let sign: Box<dyn Sign> = Box::new(parse_keypair(&data).context(error::KeyPairParse)?);
```
tjkirch: Can we get rid of the old code, or does it serve a purpose?

[edit] Ah, it's removed in the next commit.  Can we fix that so we don't introduce it just to remove it?

tough-ssm/src/lib.rs
```diff
@@ -0,0 +1,89 @@
...
+#[derive(Debug)]
+pub struct SsmKeySource {
+    pub profile: Option<String>,
+    pub parameter_name: String,
+    pub key_id: Option<String>,
+}
+
+impl SsmKeySource {
+    /// Read a key from SSM.
+    fn read(&self) -> Result<Vec<u8>> {
```
tjkirch: Do we have a need to call these methods outside the trait?  Is there a benefit to implementing it on the struct here and just calling this from the trait?

tough-ssm/src/lib.rs
```diff
@@ -0,0 +1,89 @@
...
+        let sign = Box::new(parse_keypair(&data).context(error::KeyPairParse)?);
+        Ok(sign)
+    }
+
+    fn write(
+        &self,
+        value: &str,
+        key_id_hex: &str,
+    ) -> std::result::Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
+        Ok(self.write(value, key_id_hex)?)
```
tjkirch:
```suggestion
        self.write(value, key_id_hex)
```

tuftool/src/root.rs
```diff
@@ -51,7 +52,8 @@ pub(crate) enum Command {
         /// Path to root.json
         path: PathBuf,
         /// The new key
-        key_path: KeySource,
+        #[structopt(parse(try_from_str = parse_key_source))]
+        key_path: Box<dyn KeySource>,
```
tjkirch: Not really a path anymore?  :)  Same below.

tuftool/src/source.rs
```diff
@@ -8,169 +8,54 @@
...
+///
+/// Users are welcome to add their own sources of keys by implementing
+/// the `KeySource` trait in the `tough` library. A user can then add
+/// to this parser to support them in `tuftool`.
+pub(crate) fn parse_key_source(input: &str) -> Result<Box<dyn KeySource>> {
+    let pwd_url = Url::from_directory_path(std::env::current_dir().context(error::CurrentDir)?)
+        .expect("expected current directory to be absolute");
+    let url = Url::options()
+        .base_url(Some(&pwd_url))
+        .parse(input)
```
tjkirch, line 56: Why use `options`, `base_url`, and `parse`, if we're not changing the options?  Wouldn't `Url::parse` do the same?  Is it related to the (unrecognized?) `aws-ssm` scheme?  In any case, can we comment why we're setting the base URL to the current directory?

tuftool/src/source.rs
```diff
@@ -8,169 +8,54 @@
...
+        .context(error::UrlParse { url: input })?;
+
+    match url.scheme() {
+        "file" => Ok(Box::new(LocalKeySource {
+            path: PathBuf::from(url.path()),
+        })),
+        #[cfg(any(feature = "rusoto-native-tls", feature = "rusoto-rustls"))]
+        "aws-ssm" => Ok(Box::new(SsmKeySource {
```
tjkirch, line 64: Is it possible to separate the usage of SSM from the selection of a TLS library?  It seems odd to support SSM only if you choose one of these two TLS options.  If we/rusoto add another TLS option, we have to come back and change things like this.  Can we have an "ssm" feature that brings in SSM support, and a separate set of features for TLS library?

tuftool/src/source.rs
```diff
@@ -8,169 +8,54 @@
...
+                    Some(s.to_owned())
+                }
+            }),
+            parameter_name: url.path().to_owned(),
+            key_id: url.query_pairs().find_map(|(k, v)| {
+                if k == "kms-key-id" {
+                    Some(v.into_owned())
+                } else {
+                    None
+                }
```
tjkirch, line 80: Should we error if there's no key parameter?  When would you not want to encrypt the secure string?  (Or is there a default?  Can we comment?)

tuftool/src/source.rs
```diff
@@ -8,169 +8,54 @@
...
-                }),
-                parameter_name: url.path().to_owned(),
-                key_id: url
-                    .query_pairs()
-                    .find(|(k, _)| k == "kms-key-id")
-                    .map(|(_, v)| v.into_owned()),
+/// Parses a user-specified source of signing keys.
+/// Currently supported key sources are local files and AWS SSM.
+/// Sources are passed to `tuftool` as arguments in string format:
+/// "file:///" or "aws-ssm:///"
```
tjkirch: Can we give more exact examples of each scheme?  The aws-ssm scheme, in particular, has pieces that users will probably be unfamiliar with.

tough-ssm/src/lib.rs
zmrow: I think both make sense: I'll improve the docs and see if I can nudge (or implement) the rusoto fix.

tough-ssm/src/lib.rs
zmrow: Oh gosh - I copied that code from the original so it must have slipped past in the code review for that code.

I'll fix it.

tough-ssm/src/lib.rs
zmrow: Not really. The original code had this pattern and I didn't change it.

I'll add this logic into the trait.

tough-ssm/src/error.rs
zmrow: Profile doesn't include the region sadly (unless the user named it that way)

tough-ssm/src/error.rs
zmrow: Same as above

tjkirch: Sorry, same as what above?  If the profile doesn't include region, I think it'd be helpful to separately include region, and the parameter name.  Unless there's only one?  If it's obvious because there's only one, you can ignore me...

zmrow: Sorry if I wasn't clear, we won't have region here either.

In this context, we're telling the user that SSM didn't include something in it's response. [These are the response values](https://rusoto.github.io/rusoto/rusoto_ssm/struct.Parameter.html) I have to work with, which includes the parameter that we queried for (`Parameter.name`).  However, it's an `Option<String>` so it requires handling of its own, which gets weird in the case of `None`. It would be odd for the user to get an error that says "SSM returned a response to me that didn't include the name of the parameter we queried for". I think it's clearer to the user to get an error that says "SSM didn't return the value we queried for" (in this case `parameter` or `parameter.value`).

tuftool/src/source.rs
zmrow: The base URL is required in the case of files, otherwise the `URL` lib complains about a "relative URL without a base".  This is actually mentioned in the doc comment for this file, but I'll move it down so it's closer to the applicable code.

tuftool/src/source.rs
zmrow: This parameter gives a user the option of using their own KMS key. It's primarily only used when writing to the SSM parameter store.  If nothing is provided, [the system uses the default key associated with your AWS account](https://rusoto.github.io/rusoto/rusoto_ssm/struct.PutParameterRequest.html).

I'll add additional comments for this, along with additional "docs" on the format of the aws-ssm:// schema.

tuftool/src/source.rs
zmrow: Yes - I'll add some more docs.

tough-ssm/src/error.rs
tjkirch: Let me ask a different way - will we only have a single KeySource in a single region in any given run?  If so, then the user knows what parameter they requested, and we wouldn't need to include anything about it specifically in the error.  If not, they won't know which one failed, and it'd be helpful to include the parameter name and region in the error.

Separately, it was confusing to me to reuse the `SsmMissingField` error for two cases - we didn't get a parameter back at all, and we did get a parameter back but it was missing a value.  The first seems to indicate that the user specified the parameter incorrectly or are in the wrong region or something, where the second indicates that they specified it correctly but set the parameter incorrectly or there was an SSM service issue.

tuftool/src/source.rs
tjkirch: After looking a bit more, it seems we'd either need Cargo deps to support "OR" or to have virtual dependencies, so an "ssm" feature could depend on whatever TLS the user chose, rather than having to choose between enforcing a single choice and not reflecting the real TLS dependency that SSM has.

tough-ssm/src/error.rs
zmrow: I think the answer to your first question is "yes". The user supplies the SSM KeySource at the command line, so I think they'll know what parameter they requested.

As far as the `SsmMissingField` error is concerned, I feel it's clear enough without trying to diagnose the issue. "We got a response, but it didn't contain what we wanted".  If you feel strongly about it, I'm fine with adding additional error variants.

tjkirch: > I think the answer to your first question is "yes". The user supplies the SSM KeySource at the command line, so I think they'll know what parameter they requested.

Is this only true with our current usage in tuftool, or does it apply to all usage of tough-ssm?  If we're moving this logic to tough, we have to think about use cases other than tuftool, and if that means there can be multiple `KeySource`s, then I think we have to consider clarity issues like this more carefully.  If that doesn't apply, we can resolve this.

zmrow:
Ok - I believe I've got all of @tjkirch 's comments addressed.

The last outstanding issue is the concurrency problem with the SSM Client. I've submitted [a PR to Rusoto](https://github.com/rusoto/rusoto/pull/1741) that solves this issue and should remove any concurrency concerns since it doesn't manipulate environment variables.  We should probably wait to merge this PR until upstream is fixed.

tough-ssm/src/error.rs
zmrow: I've added the parameter name to the error to make it clearer.

zmrow:
Fix clippy's anger.

tough-ssm/src/lib.rs
```diff
@@ -0,0 +1,79 @@
...
+            name: self.parameter_name.to_owned(),
+            description: Some(key_id_hex.to_owned()),
+            key_id: self.key_id.as_ref().cloned(),
+            overwrite: Some(true),
+            type_: "SecureString".to_owned(),
+            value: value.to_owned(),
+            ..rusoto_ssm::PutParameterRequest::default()
+        });
+        tokio::runtime::Runtime::new()
+            .unwrap()
```
tjkirch:
```suggestion
        tokio::runtime::Runtime::new()
            .context(error::RuntimeCreation)? 
```

tuftool/src/source.rs
```diff
@@ -1,176 +1,88 @@
...
 //! This module parses a key source command line parameter as a URL, relative to `file://$PWD`,
 //! then matches the URL scheme against ones we understand.
+//!
+//! Currently supported key sources are local files and AWS SSM.
+//!
+//! Examples of currently supported formats:
+//!
+//! Local files may be specified using a path or "file:///" prefixed path:
+//! "./a/key/file/here"
+//! "file:///./a/key/file/here" (notice the 3 slashes after the colon)
```
tjkirch, line 18: Is this a quirk of the url library, using three slashes even for a relative path?  Would an absolute path have four?

tjkirch (approved):

zmrow:
Fixed that last lingering `unwrap()`. :)

tuftool/src/source.rs
zmrow: Sort of? Normally an URL would have a host string there, but since there isn't one for files, that portion of the "URL" is blank.

webern (approved):
⛄

iliana (approved):

zmrow:
The above force-push rebases on the top of `develop`

tjkirch (commented):
[woops, wrong PR]

zmrow:
This last commit should be the final piece of the puzzle for this PR and addresses any concurrency concerns (ha) we had. It updates `rusoto` to `0.44`, which allows us to use the `region_from_profile()` function instead of mucking around with environment variables.

tough-ssm/src/lib.rs
```diff
@@ -0,0 +1,79 @@
...
+        value: &str,
+        key_id_hex: &str,
+    ) -> std::result::Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
+        let ssm_client = client::build_client(self.profile.as_deref())?;
+        let fut = ssm_client.put_parameter(rusoto_ssm::PutParameterRequest {
+            name: self.parameter_name.to_owned(),
+            description: Some(key_id_hex.to_owned()),
+            key_id: self.key_id.as_ref().cloned(),
+            overwrite: Some(true),
+            type_: Some("SecureString".to_owned()),
```
zmrow, line 66: `rusoto` 0.44 now has this field as an `Option<>`... 🤷‍♂️

tjkirch (approved):

PR #111, opened 2020-05-06
tough: add method for caching a repo with its metadata

### LINKED ISSUE #80: Add method for saving metadata relevant to a set of targets
I'd like to fetch a target from a TUF repo, store it locally, and also store with it the metadata necessary to form the trust chain from my trusted root metadata file to that target. This would allow me to continue using the target in a trusted way when a machine has not yet configured networking, for instance.

- [ ] Store the root.json trust chain
- [ ] Add a method that returns the set of metadata required to verify a set of targets

What my brain thinks might be a good interface for this right now:

```rust
fn metadata_chain(
    &self,
    targets: impl Iterator<Item = impl AsRef<str>>
) -> impl Iterator<Item = (String, Signed<Box<dyn Role>>)>
```

This is incredibly unwieldy and needs some work :)

webern:
*Issue #, if available:*

Closes #80
Supports  https://github.com/bottlerocket-os/bottlerocket/issues/91
Supports https://github.com/bottlerocket-os/bottlerocket/issues/905

*Description of changes:*

In support of Bottlerocket's datastore migrations, we need to locally cache TUF targets along with the metadata that they were signed with.
This PR adds a function, named `cache` to the `Repository` object, which takes a list of targets, and stores them along with the repo metadata into directories for future use.

*Testing Performed*

Ran existing unit/integ tests and added integ tests for the new cache function.

### DIFF
--- a/tough/src/error.rs
+++ b/tough/src/error.rs
@@ -9,6 +9,7 @@ use crate::schema::RoleType;
 use chrono::{DateTime, Utc};
 use snafu::{Backtrace, Snafu};
 use std::path::PathBuf;
+use url::Url;
 
 /// Alias for `Result<T, Error>`.
 pub type Result<T> = std::result::Result<T, Error>;
@@ -212,6 +213,40 @@ pub enum Error {
         role: RoleType,
         fetched: u64,
         expected: u64,
+        backtrace: Backtrace,
+    },
+
+    #[snafu(display("Error reading data from '{}': {}", url, source))]
+    CacheFileRead {
+        url: Url,
+        source: std::io::Error,
+        backtrace: Backtrace,
+    },
+
+    #[snafu(display("Error writing data to '{}': {}", path.display(), source))]
+    CacheFileWrite {
+        path: PathBuf,
+        source: std::io::Error,
+        backtrace: Backtrace,
+    },
+
+    #[snafu(display("Error creating the directory '{}': {}", path.display(), source))]
+    CacheDirectoryCreate {
+        path: PathBuf,
+        source: std::io::Error,
+        backtrace: Backtrace,
+    },
+
+    #[snafu(display("Error writing target file to '{}': {}", path.display(), source))]
+    CacheTargetWrite {
+        path: PathBuf,
+        source: std::io::Error,
+        backtrace: Backtrace,
+    },
+
+    #[snafu(display("The target '{}' was not found", target_name))]
+    CacheTargetMissing {
+        target_name: String,
         backtrace: Backtrace,
     },
 }
--- a/tough/src/lib.rs
+++ b/tough/src/lib.rs
@@ -20,6 +20,7 @@
     clippy::missing_errors_doc
 )]
 
+mod cache;
 mod datastore;
 pub mod error;
 mod fetch;
@@ -124,6 +125,8 @@ pub struct Repository<'a, T: Transport> {
     snapshot: Signed<Snapshot>,
     timestamp: Signed<Timestamp>,
     targets: Signed<crate::schema::Targets>,
+    limits: Limits,
+    metadata_base_url: Url,
     targets_base_url: Url,
 }
 
@@ -205,6 +208,8 @@ impl<'a, T: Transport> Repository<'a, T> {
             snapshot,
             timestamp,
             targets,
+            limits: settings.limits,
+            metadata_base_url,
             targets_base_url,
         })
     }
@@ -266,28 +271,14 @@ impl<'a, T: Transport> Repository<'a, T> {
         //   HASH is one of the hashes of the targets file listed in the targets metadata file
         //   found earlier in step 4. In either case, the client MUST write the file to
         //   non-volatile storage as FILENAME.EXT.
-        let targets = &self.targets.signed.targets;
-        Ok(if let Some(target) = targets.get(name) {
-            let sha256 = &target.hashes.sha256.clone().into_vec();
-            let file = if self.consistent_snapshot {
-                format!("{}.{}", hex::encode(sha256), name)
+        Ok(
+            if let Some(target) = self.targets.signed.targets.get(name) {
+                let (sha256, file) = self.target_digest_and_filename(target, name);
+                Some(self.fetch_target(target, &sha256, file.as_str())?)
             } else {
-                name.to_owned()
-            };
-
-            Some(fetch_sha256(
-                self.transport,
-                self.targets_base_url.join(&file).context(error::JoinUrl {
-                    path: file,
-                    url: self.targets_base_url.to_owned(),
-                })?,
-                target.length,
-                "targets.json",
-                sha256,
-            )?)
-        } else {
-            None
-        })
+                None
+            },
+        )
     }
 }
 
--- a/tough/tests/repo_cache.rs
+++ b/tough/tests/repo_cache.rs
@@ -1,0 +1,209 @@
+// Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+use std::fs::File;
+use std::io::Read;
+use std::path::PathBuf;
+use tempfile::TempDir;
+use test_utils::{dir_url, test_data};
+use tough::{FilesystemTransport, Limits, Repository, Settings};
+
+mod test_utils;
+
+struct RepoPaths {
+    root_path: PathBuf,
+    datastore: TempDir,
+    metadata_base_url: String,
+    targets_base_url: String,
+}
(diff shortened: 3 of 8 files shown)


tough/src/error.rs
```diff
@@ -214,6 +215,40 @@ pub enum Error {
         expected: u64,
         backtrace: Backtrace,
     },
+
+    #[snafu(display("Error reading data from {}: {}", url, source))]
```
zmrow: Nit: Add quotations around the `Url`/`Pathbuf` in these errors:

`Error reading data from '{}'...`

tough/src/lib.rs
```diff
@@ -266,29 +271,157 @@ impl<'a, T: Transport> Repository<'a, T> {
...
+    ///
+    /// * `metadata_outdir` is the directory where cached metadata files will be saved.
+    /// * `targets_outdir` is the directory where cached targets files will be saved.
+    /// * `targets_subset` is the list of targets to include in the cached repo. If no subset is
+    /// specified (`None`), then *all* targets are included in the cache.
+    /// * `cache_root_chain` specifies whether or not we will cache all versions of `root.json`.
+    pub fn cache<P1, P2>(
+        &self,
+        metadata_outdir: P1,
+        targets_outdir: P2,
```
zmrow: Not disagreeing, just asking: Is there any reason to separate these directory arguments?  Is there a reason to put metadata and targets at different locations?

tough/src/lib.rs
```diff
@@ -266,29 +271,157 @@ impl<'a, T: Transport> Repository<'a, T> {
...
+                self.cache_target(&targets_outdir, target_name)?;
+            }
+        } else {
+            let targets = &self.targets.signed.targets;
+            for target_name in targets.keys() {
+                self.cache_target(&targets_outdir, target_name)?;
+            }
+        }
+
+        self.cache_file_from_transport(self.snapshot_filename().as_str(), &metadata_outdir)?;
```
zmrow: Could we add a small comment here? At first glance I wondered why you were calling the same function 3 times until I saw it was to cache all the metadata.

tough/tests/repo_cache.rs
```diff
@@ -0,0 +1,204 @@
...
+#[test]
+fn test_repo_cache_all_targets() {
+    // load the reference_impl repo
+    let mut repo_paths = RepoPaths::new();
+    let repo = load_tuf_reference_impl(&mut repo_paths);
+
+    // cache the repo for future use
+    let destination = TempDir::new().unwrap();
+    let metadata_desintaion = destination.as_ref().join("metadata");
+    let targets_desintaion = destination.as_ref().join("targets");
```
zmrow: s/desintaion/destination/g

zmrow (changes requested):
Overall, it would be super helpful to add some more comments to all the functions that were added. I was getting wrapped around the axle keeping some of them straight.

Sidebar to that thought: I wonder if this might be better implemented separately as a different data structure? Perhaps a `Cache` struct with some methods hanging off of it? Something to think about but I won't block on it. @iliana might have more to add here.

tough/src/lib.rs
webern: It’s not required to do-so, unless you happen to be a) not using consistent snapshots, and b) have a target file with the same name as a metadata file. If a and b are true then you could not save targets and metadata in the same dir.

webern:
> more comments to all the functions that were added. I was getting wrapped around the axle keeping some of them straight.

Agreed, I will liberally add comments based on your feedback. I tend to prefer small helper functions that reduce the overall length and visual complexity of a routine, but not everybody agrees on where to draw the line on such (lingering effects of reading Clean Code a long time ago).

> Perhaps a `Cache` struct with some methods hanging off of it? Something to think about but I won't block on it. @iliana might have more to add here.

I agree that `lib.rs` is growing too big. And, more specifically, that Repository is getting big. Although, the entire library culminates in a Repository object so it might be hard to avoid the god class anti-pattern there.

I would vote against a struct for Cache simply because there is no data that it needs to hold (other than a Repository). I could create a cache.rs file, and extract any/all functions that are only used for the cache function, and which do not require private access into the Repository.

zmrow:
> Agreed, I will liberally add comments based on your feedback. I tend to prefer small helper functions that reduce the overall length and visual complexity of a routine, but not everybody agrees on where to draw the line on such (lingering effects of reading Clean Code a long time ago)

😄 I have no problems with the style, I just find comments are helpful.

tough/src/lib.rs
```diff
@@ -266,28 +271,174 @@ impl<'a, T: Transport> Repository<'a, T> {
...
+    fn cache_file_from_transport<P: AsRef<Path>>(&self, filename: &str, outdir: P) -> Result<()> {
+        let mut read = fetch_max_size(
+            self.transport,
+            self.metadata_base_url
+                .join(filename)
+                .context(error::JoinUrl {
+                    path: filename,
+                    url: self.metadata_base_url.to_owned(),
+                })?,
+            self.limits.max_root_size, // TODO(webern) - this is not correct
```
webern: I discovered this bug after opening the PR. I will fix this. It should be max root size for root.json files and signed size (i think) for other metadata file. So probably max_size needs to be added to the function signature.

webern:
> I have no problems with the style

At least I deleted this function which was ~~kind of~~ ridiculous...

```rust
    fn find_target(&self, name: &str) -> Option<&Target> {
        let targets = &self.targets.signed.targets;
        targets.get(name)
    }
```

webern:
Converted to draft. Properly correcting for use of the correct max_size is harder than I thought.

webern:
Re-opened, now using the correct file size limits. I tried to find a better way of providing the `specifier` string literals, but all solutions were more obfuscating than just passing a string literal. I documented the `specifier` argument because it took me a while to understand exactly what it meant.

tough/src/error.rs
```diff
@@ -214,6 +215,40 @@ pub enum Error {
...
+    },
+
+    #[snafu(display("Error writing target file to '{}': {}", path.display(), source))]
+    CacheTargetWrite {
+        path: PathBuf,
+        source: std::io::Error,
+        backtrace: Backtrace,
+    },
+
+    #[snafu(display("Error the target '{}' was not found", target_name))]
```
webern:
```suggestion
    #[snafu(display("The target '{}' was not found", target_name))]
```

tough/src/lib.rs
```diff
@@ -266,28 +271,212 @@ impl<'a, T: Transport> Repository<'a, T> {
...
+    /// * `metadata_outdir` is the directory where cached metadata files will be saved.
+    /// * `targets_outdir` is the directory where cached targets files will be saved.
+    /// * `targets_subset` is the list of targets to include in the cached repo. If no subset is
+    /// specified (`None`), then *all* targets are included in the cache.
+    /// * `cache_root_chain` specifies whether or not we will cache all versions of `root.json`.
+    pub fn cache<P1, P2>(
+        &self,
+        metadata_outdir: P1,
+        targets_outdir: P2,
+        targets_subset: Option<&Vec<String>>,
```
iliana:
```suggestion
        targets_subset: Option<&[S]>,
```

... `where S: AsRef<str>`, to make the set of types that can be used here much wider (you can pass something like `Vec<String>` that implicitly `Deref`s to a slice, or a `&[&str]`, or anything in between).

tough/src/lib.rs
```diff
@@ -266,28 +271,212 @@ impl<'a, T: Transport> Repository<'a, T> {
...
+    }
+
+    /// Prepends the version number to the timestamp.json filename if using consistent snapshot mode.
+    fn timestamp_filename(&self) -> String {
+        if self.root.signed.consistent_snapshot {
+            format!("{}.timestamp.json", self.snapshot.signed.version)
+        } else {
+            "timestamp.json".to_owned()
+        }
+    }
```
iliana: This isn't correct -- timestamp.json doesn't have the version prepended even when consistent_snapshot is true.

iliana:
My other larger feedback is to split out the cache-specific methods. You can have two modules implement methods on the same struct, just make sure the struct is in scope and open a new `impl Repository` block. The current `impl Repository` block is getting *very* big.

Assume you have my approval if the two callouts in my review are fixed. :)

tough/src/lib.rs
webern: Nice, tip.

webern:
Revised https://github.com/awslabs/tough/compare/9d97ad7b4024c007cd7815b6e8930ab7d30617dd..5faa548fba3513c207d519306c3ea7dd268fe17b

zmrow (approved):
🕹️

PR #121, opened 2020-05-14
tough: support the loading of an expired repo in unsafe mode

### LINKED ISSUE #112: Add ability to load a tuf repo while ignoring expiration dates
In our use case for Bottlerocket signed migrations, https://github.com/bottlerocket-os/bottlerocket/issues/905, we want to ignore (or possibly extend) the expiration dates in the TUF metadata. Our reasoning is that it is better, and safe-enough, to do so then to perhaps have a Bottlerocket host fail to upgrade because an expiration occurred between the time the targets were cached locally and the time the restart occurs (when the cached repo must be used without network access).

@iliana has previously proposed using the type-system to indicate when a repo is being used in an unsafe state.

For alternatives considered, see the Bottlerocket issue linked above.

webern:
*Issue #, if available:*

Closes #112 
Supports https://github.com/bottlerocket-os/bottlerocket/issues/905
Supports https://github.com/bottlerocket-os/bottlerocket/issues/91

*Description of changes:*

Add a setting to the `Settings` object such that we can choose to load a repository with expired metadata if we want to. This supports a use case in Bottlerocket when we need to load a cached, local repository without access to networking, and where failing due to an expiration cannot be tolerated.

I used an enum for this setting, `Unsafe` vs `Safe` to make it clearer what this setting does, and thus harder to accidentally choose incorrectly.


tough/src/lib.rs
```diff
@@ -37,13 +37,21 @@ use crate::datastore::Datastore;
 use crate::error::Result;
 use crate::fetch::{fetch_max_size, fetch_sha256};
 use crate::schema::{Role, RoleType, Root, Signed, Snapshot, Timestamp};
+use crate::Enforcement::Safe;
```
zmrow: I don't think this needs to be here if `Enforcement` is in the same module.

zmrow (commented):
Looks good!

Two things:
* This is a API-breaking change so we'll want to make sure that we bump the minor version of the entire library and update the changelog.
* We'll probably want to add an argument to `tuftool` so a user can download an expired repo. This is also an API breaking change and deserves a version bump.

Let's get @iliana 's eyes on this too. :)

webern:
> * we'll want to make sure that we bump the minor version of the entire library and update the changelog.

Agreed. The versions are already bumped to 0.5.0 and I plan to write the changelog in a separate release PR.

> * We'll probably want to add an argument to `tuftool` so a user can download an expired repo. This is also an API breaking change and deserves a version bump.

Yes, that could be added separately if we want the functionality exposed.

tough/src/lib.rs
```diff
@@ -44,6 +44,13 @@ use std::io::Read;
 use std::path::Path;
 use url::Url;
 
+/// Represents whether a Repository is in a `Safe`, enforcing mode, or in an `Unsafe` mode.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Enforcement {
+    Safe,
+    Unsafe,
+}
```
iliana: 1. Should the type name be more specific, like `ExpirationEnforcement`? I worry about "enforcing mode for tough" becoming an overarching term for a number of separate things.

2. Should it `impl From<bool>` (and the reverse)?

3. Should we add:

```rust
impl Default for Enforcement {
    fn default() -> Enforcement {
        Enforcement::Safe
    }
}
```

iliana (approved):
You can move my feedback into an issue. :)

tough/src/lib.rs
webern: I did the rename and impl'ed those traits.

zmrow (approved):
Nice work!

🏎️

Issue #20, opened 2019-12-04
Figure out something better than an `_extra` struct member


iliana:
`tough::schema` struct types have an `_extra` member:

https://github.com/awslabs/tough/blob/533d8db92e53d6ec1b81be785c99310402b12e8c/tough/src/schema/mod.rs#L83-L95

This is to handle keys that tough doesn't know about (yet), while still correctly validating signatures during serialization.

It'd be nice if consumers can create these structs without having to specify `_extra: HashMap::new()` or `..Default::default()`. Not sure if there's any type magick we can perform.

iliana:
Had a sudden idea here:

```rust
#[derive(Deserialize, Serialize)]
struct Verbatim<T> {
    #[serde(flatten)]
    pub inner: T,
    #[serde(flatten)]
    pub extra: HashMap<String, Value>,
}

impl<T> From<T> for Verbatim<T> {
    ...
}
```

No idea if it's possible to have two items with `#[serde(flatten)]` but this feels pretty ideal to me for a lot of reasons.

iliana:
Apparently the `Verbatim<T>` struct idea works. https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=05c006e27cd6f4a577c7fca408477c4a

serde_derive frightens me.

ecpullen:
> Apparently the `Verbatim<T>` struct idea works. https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=05c006e27cd6f4a577c7fca408477c4a
> 
> serde_derive frightens me.

Working towards implementing this feature.

PR #125, opened 2020-05-22
Adds RepositoryEditor

zmrow:
*This PR builds on the [`KeySource` trait PR](https://github.com/awslabs/tough/pull/103) that needs a Rusoto release to be merged. Once that PR is merged we can merge this one*

*Issue #, if available:*
Related to #43 
Builds on #103 

*Description of changes:*
This commit adds a RepositoryEditor struct, which allows a user
to build a new repository from scratch or edit an existing
`tough::Repository` using a pseudo builder-style pattern.

Users supply a path to a valid `root.json`, add targets, versions, and expiration dates, and then call `sign()` with signing keys. Provided all goes well, a user is given a `SignedRepository` struct which they can then use to write to disk. Convenience methods are provided to symlink targets to a location provided they exist in the `SignedRepository` and the checksums match.

`SignedRepository` does not keep track of locations of given targets. This responsibility is left to the calling code.

*Considerations and questions:*
* `RepositoryEditor` currently provides no facilities to manipulate versions or expirations; they must be provided. I think it would be nice to do so, but it's difficult to know what users could be using for versions and how they would want to increment them. Expiration dates could (should?) be unique for each role, so I didn't want to blanket implement something when a user should either update them, or save the current date and use it in the builder methods.

* While this abstracts away most of the logic from `tuftool` it does not help a user sign a *single* role (timstamp, targets, etc). We could make this possible by moving some of the logic into a `SignedRole.from_role(<ROLE HERE>)` method? I'm not sure how much of this we want to expose. It's also a bit weird to have `SignedRole` own this logic.

*Testing:*
* All unit tests pass and clippy is happy
* Added a few simple unit tests
* Currently working on an "integration" style test that (hopefully) can "sign" a repo, and then test the symlink methods on that signed repo.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1865,6 +1865,7 @@ dependencies = [
  "tempfile 3.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "untrusted 0.7.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "url 2.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "walkdir 2.3.1 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
--- a/tough/src/error.rs
+++ b/tough/src/error.rs
@@ -8,6 +8,7 @@
 use crate::schema::RoleType;
 use chrono::{DateTime, Utc};
 use snafu::{Backtrace, Snafu};
+use std::io;
 use std::path::PathBuf;
 use url::Url;
 
@@ -18,6 +19,13 @@ pub type Result<T> = std::result::Result<T, Error>;
 #[derive(Debug, Snafu)]
 #[snafu(visibility = "pub(crate)")]
 pub enum Error {
+    #[snafu(display("Unable to canonicalize path '{}': {}", path.display(), source))]
+    AbsolutePath {
+        path: PathBuf,
+        source: std::io::Error,
+        backtrace: Backtrace,
+    },
+
     /// The library failed to create a file in the datastore.
     #[snafu(display("Failed to create file at datastore path {}: {}", path.display(), source))]
     DatastoreCreate {
@@ -51,6 +59,13 @@ pub enum Error {
         backtrace: Backtrace,
     },
 
+    #[snafu(display("Failed to create directory '{}': {}", path.display(), source))]
+    DirCreate {
+        path: PathBuf,
+        source: std::io::Error,
+        backtrace: Backtrace,
+    },
+
     /// A metadata file has expired.
     #[snafu(display("{} metadata is expired", role))]
     ExpiredMetadata {
@@ -62,6 +77,13 @@ pub enum Error {
     FileRead {
         path: PathBuf,
         source: std::io::Error,
+        backtrace: Backtrace,
+    },
+
+    #[snafu(display("Failed to parse {}: {}", path.display(), source))]
+    FileParseJson {
+        path: PathBuf,
+        source: serde_json::Error,
         backtrace: Backtrace,
     },
 
@@ -96,14 +118,30 @@ pub enum Error {
         backtrace: Backtrace,
     },
 
+    #[snafu(display("Unable to parse keypair: {}", source))]
+    KeyPairFromKeySource {
+        source: Box<dyn std::error::Error + Send + Sync + 'static>,
+        backtrace: Backtrace,
+    },
+
     #[snafu(display("Private key rejected: {}", source))]
     KeyRejected {
         source: ring::error::KeyRejected,
         backtrace: Backtrace,
     },
 
+    #[snafu(display("Unable to match any of the provided keys with root.json"))]
+    KeysNotFoundInRoot { backtrace: Backtrace },
+
     #[snafu(display("Unrecognized private key format"))]
     KeyUnrecognized { backtrace: Backtrace },
+
+    #[snafu(display("Failed to create symlink at '{}': {}", path.display(), source))]
+    LinkCreate {
+        path: PathBuf,
+        source: io::Error,
+        backtrace: Backtrace,
+    },
 
     /// A file's maximum size exceeded a limit set by the consumer of this library or the metadata.
     #[snafu(display("Maximum size {} (specified by {}) exceeded", max_size, specifier))]
@@ -127,6 +165,16 @@ pub enum Error {
         role: RoleType,
         backtrace: Backtrace,
     },
+
+    #[snafu(display("Missing '{}' when building repo from RepositoryEditor", field))]
+    Missing { field: String, backtrace: Backtrace },
+
+    /// Unable to determine file name (path ends in '..' or is '/')
+    #[snafu(display("Unable to determine file name from path: '{}'", path.display()))]
+    NoFileName { path: PathBuf, backtrace: Backtrace },
+
+    #[snafu(display("Key for role '{}' doesn't exist in root.json", role))]
+    NoRoleKeysinRoot { role: String },
 
     /// A downloaded metadata file has an older version than a previously downloaded metadata file.
     #[snafu(display(
@@ -176,9 +224,41 @@ pub enum Error {
         backtrace: Backtrace,
     },
 
+    /// Path isn't a valid UTF8 string
+    #[snafu(display("Path {} is not valid UTF-8", path.display()))]
+    PathUtf8 { path: PathBuf, backtrace: Backtrace },
+
+    #[snafu(display("Failed to serialize role '{}' for signing: {}", role, source))]
+    SerializeRole {
+        role: String,
+        source: serde_json::Error,
+        backtrace: Backtrace,
+    },
+
+    #[snafu(display("Failed to serialize signed role '{}': {}", role, source))]
+    SerializeSignedRole {
+        role: String,
+        source: serde_json::Error,
+        backtrace: Backtrace,
+    },
+
     #[snafu(display("Failed to sign message"))]
     Sign {
         source: ring::error::Unspecified,
+        backtrace: Backtrace,
+    },
+
+    #[snafu(display("Unable to find signing keys for role '{}'", role))]
+    SigningKeysNotFound { role: String },
+
+    #[snafu(display(
+        "Tried to use role metadata with spec version '{}', version '{}' is supported",
+        given,
+        supported
+    ))]
+    SpecVersion {
+        given: String,
+        supported: String,
         backtrace: Backtrace,
     },
 
@@ -191,6 +271,13 @@ pub enum Error {
     SystemTimeSteppedBackward {
         sys_time: DateTime<Utc>,
         latest_known_time: DateTime<Utc>,
+    },
+
+    #[snafu(display("Unable to create Target from path '{}': {}", path.display(), source))]
+    TargetFromPath {
+        path: PathBuf,
+        source: crate::schema::Error,
+        backtrace: Backtrace,
     },
 
     /// A transport error occurred while fetching a URL.
@@ -261,6 +348,13 @@ pub enum Error {
     #[snafu(display("The target '{}' was not found", target_name))]
     CacheTargetMissing {
         target_name: String,
+        backtrace: Backtrace,
+    },
+
+    #[snafu(display("Failed to walk directory tree '{}': {}", directory.display(), source))]
+    WalkDir {
+        directory: PathBuf,
+        source: walkdir::Error,
         backtrace: Backtrace,
     },
 }
--- a/tough/src/schema/mod.rs
+++ b/tough/src/schema/mod.rs
@@ -13,6 +13,7 @@ pub use crate::schema::error::{Error, Result};
 use crate::schema::decoded::{Decoded, Hex};
 use crate::schema::iter::KeysIter;
 use crate::schema::key::Key;
+use crate::sign::Sign;
 use chrono::{DateTime, Utc};
 use olpc_cjson::CanonicalFormatter;
 use ring::digest::{Context, SHA256};
@@ -44,6 +45,8 @@ pub trait Role: Serialize {
     const TYPE: RoleType;
 
     fn expires(&self) -> DateTime<Utc>;
+
+    fn version(&self) -> NonZeroU64;
 
     fn canonical_form(&self) -> Result<Vec<u8>> {
         let mut data = Vec::new();
@@ -120,6 +123,17 @@ impl Root {
             keys: &self.keys,
         }
     }
+
+    /// Given an object/key that impls Sign, return the corresponding
+    /// key ID from Root
+    pub fn key_id(&self, key_pair: &dyn Sign) -> Option<Decoded<Hex>> {
+        for (key_id, key) in &self.keys {
+            if key_pair.tuf_key() == *key {
+                return Some(key_id.clone());
+            }
+        }
+        None
+    }
 }
 
 impl Role for Root {
@@ -127,6 +141,10 @@ impl Role for Root {
 
     fn expires(&self) -> DateTime<Utc> {
         self.expires
+    }
+
+    fn version(&self) -> NonZeroU64 {
+        self.version
     }
 }
 
@@ -181,11 +199,26 @@ pub struct Hashes {
     pub _extra: HashMap<String, Value>,
 }
 
+impl Snapshot {
+    pub fn new(spec_version: String, version: NonZeroU64, expires: DateTime<Utc>) -> Self {
+        Snapshot {
+            spec_version,
+            version,
+            expires,
+            meta: HashMap::new(),
+            _extra: HashMap::new(),
+        }
+    }
+}
 impl Role for Snapshot {
     const TYPE: RoleType = RoleType::Snapshot;
 
     fn expires(&self) -> DateTime<Utc> {
         self.expires
+    }
+
+    fn version(&self) -> NonZeroU64 {
+        self.version
     }
 }
 
@@ -267,11 +300,27 @@ impl Target {
     }
 }
 
+impl Targets {
+    pub fn new(spec_version: String, version: NonZeroU64, expires: DateTime<Utc>) -> Self {
+        Targets {
+            spec_version,
+            version,
+            expires,
+            targets: HashMap::new(),
+            _extra: HashMap::new(),
+        }
+    }
+}
+
 impl Role for Targets {
     const TYPE: RoleType = RoleType::Targets;
 
     fn expires(&self) -> DateTime<Utc> {
         self.expires
+    }
+
+    fn version(&self) -> NonZeroU64 {
+        self.version
     }
 }
 
@@ -311,10 +360,26 @@ pub struct TimestampMeta {
     pub _extra: HashMap<String, Value>,
 }
 
+impl Timestamp {
+    pub fn new(spec_version: String, version: NonZeroU64, expires: DateTime<Utc>) -> Self {
+        Timestamp {
+            spec_version,
+            version,
+            expires,
+            meta: HashMap::new(),
+            _extra: HashMap::new(),
+        }
+    }
+}
+
 impl Role for Timestamp {
     const TYPE: RoleType = RoleType::Timestamp;
 
     fn expires(&self) -> DateTime<Utc> {
         self.expires
+    }
+
+    fn version(&self) -> NonZeroU64 {
+        self.version
     }
 }
--- a/tough/tests/repo_editor.rs
+++ b/tough/tests/repo_editor.rs
@@ -1,0 +1,152 @@
+// Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+use chrono::{Duration, Utc};
+use std::fs::File;
+use std::io::Read;
+use std::num::NonZeroU64;
+use std::path::PathBuf;
+use tempfile::TempDir;
+use test_utils::{dir_url, test_data};
+use tough::editor::RepositoryEditor;
+use tough::key_source::LocalKeySource;
+use tough::{ExpirationEnforcement, FilesystemTransport, Limits, Repository, Settings};
+
+mod test_utils;
+
+struct RepoPaths {
+    root_path: PathBuf,
+    datastore: TempDir,
+    metadata_base_url: String,
+    targets_base_url: String,
+}
+
+impl RepoPaths {
+    fn new() -> Self {
+        let base = test_data().join("tuf-reference-impl");
+        RepoPaths {
+            root_path: base.join("metadata").join("1.root.json"),
+            datastore: TempDir::new().unwrap(),
+            metadata_base_url: dir_url(base.join("metadata")),
+            targets_base_url: dir_url(base.join("targets")),
+        }
+    }
+
+    fn root(&self) -> File {
+        File::open(&self.root_path).unwrap()
+    }
+}
+
+fn load_tuf_reference_impl<'a>(paths: &'a mut RepoPaths) -> Repository<'a, FilesystemTransport> {
+    Repository::load(
+        &tough::FilesystemTransport,
+        Settings {
+            root: &mut paths.root(),
+            datastore: paths.datastore.as_ref(),
+            metadata_base_url: paths.metadata_base_url.as_str(),
+            targets_base_url: paths.targets_base_url.as_str(),
+            limits: Limits::default(),
+            expiration_enforcement: ExpirationEnforcement::Safe,
+        },
+    )
+    .unwrap()
+}
+
+// Test a RepositoryEditor can be created from an existing Repo
+#[test]
+fn repository_editor_from_repository() {
+    // Load the reference_impl repo
+    let mut repo_paths = RepoPaths::new();
+    let root = repo_paths.root_path.clone();
+    let repo = load_tuf_reference_impl(&mut repo_paths);
+
+    assert!(RepositoryEditor::from_repo(&root, repo).is_ok());
+}
+
+// Load a repository, edit it, and write it to disk. Ensure it loads correctly
+// and attempt to read a target
+#[test]
+fn repo_load_edit_write_load() {
+    let mut repo_paths = RepoPaths::new();
+    let repo = load_tuf_reference_impl(&mut repo_paths);
+
+    let root = test_data().join("simple-rsa").join("root.json");
+    let root_key = test_data().join("snakeoil.pem");
+    let key_source = LocalKeySource { path: root_key };
+    let timestamp_expiration = Utc::now().checked_add_signed(Duration::days(3)).unwrap();
+    let timestamp_version = NonZeroU64::new(1234).unwrap();
+    let snapshot_expiration = Utc::now().checked_add_signed(Duration::days(21)).unwrap();
+    let snapshot_version = NonZeroU64::new(5432).unwrap();
+    let targets_expiration = Utc::now().checked_add_signed(Duration::days(13)).unwrap();
+    let targets_version = NonZeroU64::new(789).unwrap();
+    let targets_location = test_data().join("tuf-reference-impl").join("targets");
+    let target3 = targets_location.join("file3.txt");
+
+    // Load the reference_impl repo
+    let mut editor = RepositoryEditor::from_repo(&root, repo).unwrap();
+
+    // Add the required data and a new target
+    // We clear the targets first because the reference implementation includes
+    // "file1.txt" and "file2.txt". The reference implementation's "targets"
+    // directory includes all 3 targets. We want to explicitly add "file3.txt"
+    // as a target, and later ensure that "file3" is the only target in the
+    // new repo and the only target that gets symlinked. Doing so tests the
+    // implementation of `SignedRepository.link_targets()`.
+    editor
+        .targets_expires(targets_expiration)
+        .targets_version(targets_version)
+        .snapshot_expires(snapshot_expiration)
+        .snapshot_version(snapshot_version)
+        .timestamp_expires(timestamp_expiration)
+        .timestamp_version(timestamp_version)
+        .clear_targets()
+        .add_target(target3)
+        .unwrap();
+
+    // Sign the newly updated repo
+    let signed_repo = editor.sign(&[Box::new(key_source)]).unwrap();
+
+    // Create the directories and write the repo to disk
+    let destination = TempDir::new().unwrap();
+    let metadata_destination = destination.as_ref().join("metadata");
+    let targets_destination = destination.as_ref().join("targets");
+    assert!(signed_repo.write(&metadata_destination).is_ok());
+    assert!(signed_repo
+        .link_targets(&targets_location, &targets_destination)
+        .is_ok());
+
+    // Load the repo we just created
+    let datastore = TempDir::new().unwrap();
+    let metadata_base_url = dir_url(&metadata_destination);
+    let targets_base_url = dir_url(&targets_destination);
+    let new_repo = Repository::load(
+        &tough::FilesystemTransport,
+        Settings {
+            root: File::open(&root).unwrap(),
+            datastore: datastore.as_ref(),
+            metadata_base_url: metadata_base_url.as_str(),
+            targets_base_url: targets_base_url.as_str(),
+            limits: Limits::default(),
+            expiration_enforcement: ExpirationEnforcement::Safe,
+        },
+    )
+    .unwrap();
+
+    // Ensure the new repo only has the single target
+    assert_eq!(new_repo.targets().signed.targets.len(), 1);
+
+    // The repo shouldn't contain file1 or file2
+    // `read_target()` returns a Result(Option<>) which is why we unwrap
+    assert!(new_repo.read_target("file1.txt").unwrap().is_none());
+    assert!(new_repo.read_target("file2.txt").unwrap().is_none());
+
+    // Read file3.txt
+    let mut file_data = Vec::new();
+    let file_size = new_repo
+        .read_target("file3.txt")
+        .unwrap()
+        .unwrap()
+        .read_to_end(&mut file_data)
+        .unwrap();
+    assert_eq!(28, file_size);
+}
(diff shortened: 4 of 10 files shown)


tough/src/editor.rs
```diff
@@ -0,0 +1,660 @@
...
+    }
+
+    /// Crawls a given directory and symlinks any targets found to the given
+    /// "out" directory.
+    ///
+    /// For each file found in the `indir`, the method gets the filename and
+    /// if the filename exists in `Targets`, the file's sha256 is compared
+    /// against the data in `Targets`. If this data does not match, the
+    /// method will fail. If all is well, the taret is symlinked.
+    pub fn link_targets<P1, P2>(&self, indir: P1, outdir: P2) -> Result<()>
```
zmrow: I realize this method is fairly long and I spent entirely too long trying to break it up.  Given the data we need to keep track of and the order of operations here, I left the logic as is since it is straightforward and linear.

Cargo.lock
```diff
@@ -1865,6 +1865,7 @@ dependencies = [
  "tempfile 3.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "untrusted 0.7.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "url 2.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "walkdir 2.3.1 (registry+https://github.com/rust-lang/crates.io-index)",
```
tjkirch, line 1868: This seems odd; you have walkdir 2.2.9 in Cargo.toml, but we're getting 2.3.x here?  I would have expected it to pull in 2.2.latest by Cargo's default caret requirement rules.  Any idea why?

I think it'd make sense to use 2.2.x to match existing crates, or update everything to 2.3.x (probably separately) so they don't have to build twice, but I just wanted to raise this question...

tough/src/schema/mod.rs
```diff
@@ -120,6 +123,18 @@ impl Root {
...
+    /// Given an object/key that impls Sign, return the corresponding
+    /// key ID from Root
+    #[allow(clippy::unnecessary_operation)]
+    pub fn key_id(&self, key_pair: &dyn Sign) -> Option<Decoded<Hex>> {
+        for (key_id, key) in &self.keys {
+            if key_pair.tuf_key() == *key {
+                Some(key_id.clone());
+            }
+        }
+        None
```
tjkirch: Is this clippy warning because we're not actually `return`ing the cloned `key_id`?  It seems like it'll always return `None`.

tough/tests/repo_editor.rs
```diff
@@ -0,0 +1,60 @@
...
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+use std::fs::File;
+use std::path::PathBuf;
+use tempfile::TempDir;
+use test_utils::{dir_url, test_data};
+use tough::editor::RepositoryEditor;
+use tough::{ExpirationEnforcement, FilesystemTransport, Limits, Repository, Settings};
+
+mod test_utils;
```
tjkirch, line 15: It might be clearer to have a `tough/tests/mod.rs` with `mod test_utils;` so each tests file doesn't have to declare it?

tough/src/editor.rs
```diff
@@ -0,0 +1,660 @@
...
+            sha256,
+            length,
+        };
+
+        Ok(signed_role)
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
tjkirch: Why split the tests between this file and the separate tests file?

tough/tests/repo_editor.rs
```diff
@@ -0,0 +1,60 @@
...
+
+/// Test a RepositoryEditor can be created from an existing Repo
+#[test]
+fn repository_editor_from_repository() {
+    // load the reference_impl repo
+    let mut repo_paths = RepoPaths::new();
+    let root = repo_paths.root_path.clone();
+    let repo = load_tuf_reference_impl(&mut repo_paths);
+
+    assert!(RepositoryEditor::from_repo(&root, repo).is_ok());
```
tjkirch, line 63: It seems like it'd be good to extend the test (or add a test) to change a field, `write` the repo, and then make sure the written repo also loads correctly through the standard `Repository::load`.

tough/src/error.rs
```diff
@@ -18,6 +19,14 @@ pub type Result<T> = std::result::Result<T, Error>;
 #[derive(Debug, Snafu)]
 #[snafu(visibility = "pub(crate)")]
 pub enum Error {
+    /// The library failed to canonicalize the given path
```
tjkirch:
```suggestion
```
I think the `display` is pretty clear - does this help in some other way?  I thought docstrings were mainly used if you don't need any variables in the `display`.  (Same with below docstrings)

tough/src/editor.rs
```diff
@@ -0,0 +1,660 @@
...
+    /// For each file found in the `indir`, the method gets the filename and
+    /// if the filename exists in `Targets`, the file's sha256 is compared
+    /// against the data in `Targets`. If this data does not match, the
+    /// method will fail. If all is well, the taret is symlinked.
+    pub fn link_targets<P1, P2>(&self, indir: P1, outdir: P2) -> Result<()>
+    where
+        P1: AsRef<Path>,
+        P2: AsRef<Path>,
+    {
+        std::fs::create_dir_all(&outdir).context(error::DirCreate {
```
tjkirch:
```suggestion
        let outdir = outdir.as_ref();
        std::fs::create_dir_all(&outdir).context(error::DirCreate {
```
`outdir` is used a number of times here; this will let you remove the `&` and `.as_ref()` when using it.

tough/src/editor.rs
```diff
@@ -0,0 +1,660 @@
...
+            path: &outdir.as_ref(),
+        })?;
+        // Get the absolute path of the indir and outdir
+        let abs_indir = std::fs::canonicalize(&indir).context(error::AbsolutePath {
+            path: indir.as_ref().to_owned(),
+        })?;
+        let abs_outdir = std::fs::canonicalize(&outdir).context(error::AbsolutePath {
+            path: outdir.as_ref().to_owned(),
```
tjkirch:
```suggestion
            path: indir.as_ref(),
        })?;
        let abs_outdir = std::fs::canonicalize(&outdir).context(error::AbsolutePath {
            path: outdir.as_ref(),
```
Snafu should handle cloning for you, and only in the error case; this would clone even in the success case.  (If you need to do it manually for some reason I'm missing, `.with_context(|| bla)` will only run in the error case, which is good if the error creation takes a few cycles...)

tough/src/error.rs
```diff
@@ -128,6 +167,21 @@ pub enum Error {
         backtrace: Backtrace,
     },
 
+    /// A required reference to a metadata file is missing from a metadata file.
+    #[snafu(display("Missing field: {}", field))]
```
tjkirch:
```suggestion
    #[snafu(display("Missing '{}' when building repo from RepositoryEditor", field))]
```

tough/src/error.rs
```diff
@@ -128,6 +167,21 @@ pub enum Error {
...
+    Missing { field: String, backtrace: Backtrace },
+
+    /// Unable to determine file name (path ends in '..' or is '/')
+    #[snafu(display("Unable to determine file name from path: '{}'", path.display()))]
+    NoFileName { path: PathBuf, backtrace: Backtrace },
+
+    /// Can't create a repo with no targets
+    #[snafu(display("Unable to create a TUF repo without targets"))]
+    NoTargets {},
+
```
tjkirch:
```suggestion
```
Wound up not using this.

tough/src/editor.rs
```diff
@@ -0,0 +1,660 @@
...
+        // If it's important to have a non-empty targets, the object can be
+        // inspected by the calling code.
+        let mut targets: HashMap<String, Target> = HashMap::new();
+        if let Some(ref existing_targets) = self.existing_targets {
+            targets.extend(existing_targets.clone());
+        }
+        if let Some(ref new_targets) = self.new_targets {
+            targets.extend(new_targets.clone());
+        }
+        //ensure!(!targets.is_empty(), error::NoTargets);
```
tjkirch:
```suggestion
```

tough/src/error.rs
```diff
@@ -128,6 +167,21 @@ pub enum Error {
...
+    #[snafu(display("Key for role '{}' do not exist in root.json", role))]
```
tjkirch:
```suggestion
    #[snafu(display("Key for role '{}' doesn't exist in root.json", role))]
```

tough/src/editor.rs
```diff
@@ -0,0 +1,660 @@
...
+        let mut role = Signed {
+            signed: role,
+            signatures: Vec::new(),
+        };
+
+        let mut data = Vec::new();
+        let mut ser = serde_json::Serializer::with_formatter(&mut data, CanonicalFormatter::new());
+        role.signed
+            .serialize(&mut ser)
+            .context(error::SerializeRole)?;
```
tjkirch: Can we include the `T::TYPE` name in this error?  Same for SerializeSignedRole just below?

tough/src/editor.rs
```diff
@@ -0,0 +1,660 @@
...
+        let abs_outdir = std::fs::canonicalize(&outdir).context(error::AbsolutePath {
+            path: outdir.as_ref().to_owned(),
+        })?;
+        let repo_targets = &self.targets.signed.signed.targets;
+
+        // Walk the absolute path of the indir. Using the absolute path here
+        // means that `entry.path()` call will return its absolute path.
+        let walker = WalkDir::new(abs_indir).follow_links(false);
+        for entry in walker {
+            let entry = entry.context(error::WalkDir)?;
```
tjkirch: Can we include the path we failed to walk in the error?

tjkirch (commented):
Going to review in a couple batches; this is everything but editor.rs.

tough/src/editor.rs
```diff
@@ -0,0 +1,660 @@
...
+    where
+        P: AsRef<Path>,
+    {
+        let outdir = outdir.as_ref();
+        std::fs::create_dir_all(&outdir).context(error::DirCreate { path: &outdir })?;
+
+        let filename = match T::TYPE {
+            RoleType::Targets => {
+                if consistent_snapshot {
+                    format!("{}.targets.json", self.signed.signed.version())
```
tjkirch: We have this naming logic in 4 places now, maybe we should do it in one place?  (tuftool/tests/refresh_command, Repository::*_filename, lib::load_targets, and now here)

tough/src/editor.rs
```diff
@@ -0,0 +1,660 @@
...
+                } else {
+                    "targets.json".to_string()
+                }
+            }
+            RoleType::Snapshot => {
+                if consistent_snapshot {
+                    format!("{}.snapshot.json", self.signed.signed.version())
+                } else {
+                    "targets.json".to_string()
```
tjkirch:
```suggestion
                    "snapshot.json".to_string()
```

tough/src/editor.rs
```diff
@@ -0,0 +1,660 @@
...
+
+        let path = outdir.join(filename);
+        std::fs::write(&path, &self.buffer).context(error::FileWrite { path })
+    }
+}
+
+/// A set of signed TUF Repository metadata.
+///
+/// This metadata represents a signed TUF repository, but without the actual
+/// targets, the repository cannot be used. It is up to ther user to ensure
```
tjkirch:
```suggestion
/// targets, the repository cannot be used. It is up to the user to ensure
```

tough/src/editor.rs
```diff
@@ -0,0 +1,660 @@
...
+/// all the target files are in the proper directories along with the
+/// contained metadata. There are convenience methods provided to help with
+/// this.
```
tjkirch:
```suggestion
/// This metadata represents a signed TUF repository and provides the ability to write the metadata to disk.
///
/// Note: without the target files, the repository cannot be used. It is up to the user to ensure
/// all target files referenced by the metadata are available. There are convenience methods to help with this.
```
Slight simplification, and not presenting a "negative" first.

tough/src/editor.rs
```diff
@@ -0,0 +1,660 @@
...
+            path: indir.as_ref().to_owned(),
+        })?;
+        let abs_outdir = std::fs::canonicalize(&outdir).context(error::AbsolutePath {
+            path: outdir.as_ref().to_owned(),
+        })?;
+        let repo_targets = &self.targets.signed.signed.targets;
+
+        // Walk the absolute path of the indir. Using the absolute path here
+        // means that `entry.path()` call will return its absolute path.
+        let walker = WalkDir::new(abs_indir).follow_links(false);
```
tjkirch: Why not follow links?  I'm not sure we should dictate how someone lays out their directories.

tough/src/editor.rs
```diff
@@ -0,0 +1,660 @@
...
+        for entry in walker {
+            let entry = entry.context(error::WalkDir)?;
+
+            // If the entry is a file...
+            if entry.file_type().is_file() {
```
tjkirch:
```suggestion
            if !entry.file_type().is_file() {
                continue;
            }
```
Rather than the `else` way below.

tough/src/editor.rs
```diff
@@ -0,0 +1,660 @@
...
+                // get the filename...
+                let file_name = entry
+                    .file_name()
+                    .to_str()
+                    .context(error::PathUtf8 { path: entry.path() })?;
+
+                // use the file name to see if a target exists in the repo
+                // with that name. If so...
+                if let Some(repo_target) = repo_targets.get(file_name) {
```
tjkirch: I wouldn't always recommend this, but there's so much nesting in the function that I think this would be clearer overall:
```suggestion
                let repo_target = match repo_targets.get(file_name) {
                    Some(repo_target) => repo_target,
                    None => continue;
                };
```

tough/src/editor.rs
```diff
@@ -0,0 +1,660 @@
...
+                    // create a Target object using the entry's path, and then...
+                    let target_from_path = Target::from_path(entry.path())
+                        .context(error::TargetFromPath { path: entry.path() })?;
+
+                    // compare the hashes of the target from the repo and the
+                    // target we just created. If they are the same, this must
+                    // be the same file, symlink it.
+                    if target_from_path.hashes.sha256 == repo_target.hashes.sha256 {
```
tjkirch: Similar to above suggestions, remove some nesting:
```suggestion
                    ensure!(target_from_path.hashes.sha256 == repo_target.hashes.sha256,
                        error::HashMismatch { bla });
```

tough/src/editor.rs
```diff
@@ -0,0 +1,660 @@
...
+                                file_name
+                            ))
+                        } else {
+                            abs_outdir.join(&file_name)
+                        };
+
+                        symlink(entry.path(), &dest).context(error::LinkCreate { path: &dest })?;
+                    } else {
+                        return error::HashMismatch {
+                            context: "target".to_string(),
```
tjkirch:
```suggestion
                            context: "target",
```
Snafu should do this, and only when it needs to (though this is currently in an error branch so it'd only happen on error... still, no need to do it manually)

tough/src/editor.rs
```diff
@@ -0,0 +1,660 @@
...
+                    let target_from_path = Target::from_path(entry.path())
+                        .context(error::TargetFromPath { path: entry.path() })?;
+
+                    // compare the hashes of the target from the repo and the
+                    // target we just created. If they are the same, this must
+                    // be the same file, symlink it.
+                    if target_from_path.hashes.sha256 == repo_target.hashes.sha256 {
+                        let dest = if self.root.signed.signed.consistent_snapshot {
+                            abs_outdir.join(format!(
+                                "{}.{}",
```
tjkirch: Similar to the json file naming, we're now doing this name generation in 4 places, should we standardize?  (Repository::target_digest_and_filename, tuftool::metadata::write_metadata, tuftool::create::process_target, and now this)

tough/src/editor.rs
```diff
@@ -0,0 +1,660 @@
...
+        }
+
+        Ok(())
+    }
+}
+
+/// `RepositoryEditor` contains the various bits of data needed to construct
+/// or edit a TUF repository.
+///
+/// A new repository may be created using the `new()` method.
```
tjkirch:
```suggestion
/// A new repository may be started using the `new()` method.
```

tough/src/editor.rs
```diff
@@ -0,0 +1,660 @@
...
+///
+/// An existing `tough::Repository` may be loaded and edited using the
+/// `from_repo()` method. When a repo is loaded in this way, versions and
+/// expirations are discarded. It is good practice to update these whenever
+/// a repo is changed.
+///
+/// Targets, versions, and expirations may be added to their respective roles
+/// via the provided "setter" methods. The final step in the process is the
+/// `sign()` method, which takes a given set of signing keys, builds each of
+/// the roles using the data provided, and signs the roles.
```
tjkirch:
```suggestion
/// the roles using the data provided, and signs the roles.
/// This results in a `SignedRepository` which can be used to write the repo to disk.
```

tough/src/error.rs
```diff
@@ -176,12 +230,42 @@ pub enum Error {
...
     Sign {
         source: ring::error::Unspecified,
         backtrace: Backtrace,
     },
 
+    #[snafu(display("Unable to find signing keys for role '{}'", role))]
+    SigningKeysNotFound { role: String },
+
+    #[snafu(display(
+        "Tried to load repository with spec version '{}', version '{}' is supported",
```
tjkirch: Not necessarily a repository; we also use this error when we try to set targets, which you can do in a fresh editor.

tough/src/editor.rs
```diff
@@ -0,0 +1,660 @@
...
+        let mut editor = RepositoryEditor::new(root_path)?;
+        editor.targets(repo.targets.signed)?;
+        editor.snapshot(repo.snapshot.signed)?;
+        editor.timestamp(repo.timestamp.signed)?;
+
+        Ok(editor)
+    }
+
+    /// Add an existing `Targets` struct to the repository. Any existing
+    /// targets are preserved.
```
tjkirch: Not sure I'm understanding this - existing targets aren't preserved, we're fully replacing `self.existing_targets`.  Should this be called something like `set_targets` or `replace_targets`?  Or were you intending to merge the given targets with `existing_targets`?

tough/src/editor.rs
```diff
@@ -0,0 +1,660 @@
...
+        );
+
+        // Hold on to the existing targets
+        self.existing_targets = Some(targets.targets);
+        self.targets_extra = Some(targets._extra);
+        Ok(self)
+    }
+
+    /// Add an existing `Snapshot` to the repository. Only the `_extra` data
+    /// is preserved
```
tjkirch: If the intent is just to allow setting `snapshot_extra`, and we always construct `snapshot`, should we call this `snapshot_extra`?

(Same with `fn timestamp`)

tough/src/editor.rs
tjkirch: Another option is having `add_targets` and `clear_targets` methods, so a user can work with `Targets` objects if desired (merging, replacing, etc.), and changing the path-based methods below to `add_target(s)_path` if they want to work with paths..?  One use case we've discussed is merging repos, so this distinction seems useful.

tough/src/editor.rs
```diff
@@ -0,0 +1,660 @@
...
+            targets.spec_version == SPEC_VERSION,
+            error::SpecVersion {
+                given: targets.spec_version,
+                supported: SPEC_VERSION
+            }
+        );
+
+        // Hold on to the existing targets
+        self.existing_targets = Some(targets.targets);
+        self.targets_extra = Some(targets._extra);
```
tjkirch: Do we need methods to set *_extra fields?  Currently these methods that fully replace exiting targets/bla are the only way to set it.

tough/src/editor.rs
```diff
@@ -0,0 +1,660 @@
...
+    fn build_snapshot(&self, signed_targets: &SignedRole<Targets>) -> Result<Snapshot> {
+        let version = self.snapshot_version.context(error::Missing {
+            field: "snapshot version",
+        })?;
+        let expires = self.snapshot_expires.context(error::Missing {
+            field: "snapshot expiration",
+        })?;
+        let _extra = self.snapshot_extra.clone().unwrap_or_else(HashMap::new);
+
+        let targets_meta = Self::snapshot_meta(signed_targets);
```
tjkirch: Sorry if this is a dumb question... why do we use `snapshot_meta` to make a variable called `targets_meta` associated with `targets.json`, and below we use `timestamp_meta` to make a variable called `snapshot_meta` associated with `snapshot.json`?  Naively they seem to be used backwards... :)

I think we could use a little more documentation on/in these build methods.  It's complicated stuff.

tough/src/editor.rs
```diff
@@ -0,0 +1,660 @@
...
+                _extra: HashMap::new(),
+            },
+            length: role.length,
+            version: role.signed.signed.version(),
+            _extra: HashMap::new(),
+        }
+    }
+
+    // Signs a given role. This function finds the appropriate keys to sign
+    // given role, signs the role, and builds a `SignedRole<T>`.
```
tjkirch: Can we use docstrings `///` instead of comments?

tough/src/editor.rs
```diff
@@ -0,0 +1,660 @@
...
+            .targets(targets.signed)
+            .unwrap()
+            .snapshot(snapshot.signed)
+            .unwrap()
+            .timestamp(timestamp.signed)
+            .unwrap();
+
+        let targets = editor.existing_targets.as_ref().unwrap();
+
+        assert!(editor.targets_version.is_none());
```
tjkirch: I'm not sure of the end goal of this test function, other than the unwraps not failing - these `is_none` checks don't seem to be helping us much since that'd be the default :)  I guess it checks that file1/file2 are in targets?  Is that enough to have confidence?

tjkirch (commented):
I had a hard time following the build logic, so I think that could use some documentation, but the code structure seems clear overall.

tough/src/editor.rs
webern: I agree. It's probably also time to put those strings somewhere as consts.

tough/src/editor.rs
```diff
@@ -0,0 +1,660 @@
...
+pub struct SignedRepository {
+    root: SignedRole<Root>,
+    targets: SignedRole<Targets>,
+    snapshot: SignedRole<Snapshot>,
+    timestamp: SignedRole<Timestamp>,
+}
+
+impl SignedRepository {
+    /// Writes the metadata to the given directory. If consistent snapshots
+    /// are used, the appropriate files are prefixed with their `sha256`.
```
webern: slightly confusing. in the snapshot.json file these sha256 prefixes should appear, but the metadata files get a version prefix.

tough/src/editor.rs
webern: > I'm not sure of the end goal of this test function, other than the unwraps not failing - these `is_none` checks don't seem to be helping us much since that'd be the default :) I guess it checks that file1/file2 are in targets? Is that enough to have confidence?

The issue might be that we are using the simplistic 'reference implementation' tuf repo. We should consider adding a more complex tuf repo to the test data. One that has consistent snapshots and multiple keys, for example. In other words, the assertions are as boring as the tuf repo is.

Further: We should develop the test utils to build repos on-the-fly during testing to start doing more varied testing with edge cases and such.

tough/src/schema/mod.rs
```diff
@@ -267,12 +301,28 @@ impl Target {
     }
 }
 
+impl Targets {
+    pub fn new(spec_version: String, version: NonZeroU64, expires: DateTime<Utc>) -> Self {
```
webern, line 304: These `new` functions probably address @iliana's concern in #20

tough/tests/repo_editor.rs
webern: > It seems like it'd be good to extend the test (or add a test) to change a field, `write` the repo, and then make sure the written repo also loads correctly through the standard `Repository::load`.

I agree. If not this PR then maybe in the future, but more tests that serve the dual purpose of use case examples would be nice.

webern (approved):
🍿

tough/src/schema/mod.rs
zmrow: :)

Yes - that little `return` fixed all my testing issues as well. :)

tough/src/error.rs
zmrow: Honestly I only added these to stay consistent with the style in the file.

tough/src/schema/mod.rs
webern: Clippy can be pretty helpful at times.
Another one I've found is if I'm calling `fn do_something() -> Result<()>` like this `do_something();` clippy helpfully tells me I'm ignoring the error!

tough/src/editor.rs
zmrow: I guess I'm drawing a line, albeit a squiggly one, between the unit tests and tests that include other parts of the library.

I was thinking that the tests that create/write/load would go here since they use more than just the `Editor` library. Tests that specifically test the `Editor` library would go in the lib.

tough/tests/repo_editor.rs
zmrow: Yep - I have a separate integration test waiting in the wings for just that.

tough/src/editor.rs
zmrow: Good point. I'll clarify the comment as well.

One method I completely missed is a `remove_target` method as well.

tough/src/editor.rs
zmrow: I see your point. The `*_extra` member of the role structs only exist to capture extra arguments found during deserialization and not something I'd think a user would probably interact with in most cases.

I really see these `snapshot` and `timestamp` setters as convenience methods so a user can just pass the entire object. They keep a consistent API, allow us to check the spec version, and give us the opportunity to do more in the future if the TUF spec changes.

tough/src/editor.rs
zmrow: I'm not opposed to it, but I'm wondering how much of a use case it will really have.

tjkirch: I agree, I'd rather leave it out.  I hope no one would rely on those extra fields just because they can set them when deserializing.  (I hope we can eliminate them at some point...)

tough/src/editor.rs
zmrow: Not a dumb question at all. I think the confusion comes from the variable name (suggestions welcome). Snapshot contains the filename and hashes of the signed `Targets` and `Root` data. Further details below.

The `Snapshot` struct has a member `meta`. `Snapshot.meta` contains a `HashMap<String, SnapshotMeta>`.  This `Hashmap` is where the data about the `Targets/Root` is stored. In practice it looks something like:

```
{
   ...,
   "meta":{
      "root.json":{
         "length":1,
         "hashes":{
            "sha256":"1234"
         },
         "version":1
      },
      "targets.json":{
         "length":2,
         "hashes":{
            "sha256":"5678"
         },
         "version":1588449897
      }
   }
}
```

tjkirch: Maybe a comment above this like `// Snapshot stores metadata about targets and root.`  And below in the other function, `// Targets stores metadata about snapshot.` or something..?

tough/src/editor.rs
zmrow: I guess I saw this as more of a "unit" test for the methods that add data from existing roles. I just want to make sure that we only keep the data we want to.

tough/src/editor.rs
zmrow: Oops - you're right. Will fix.

tough/src/schema/mod.rs
zmrow: I don't currently implement `new()` for all of these types unfortunately. :(

Cargo.lock
zmrow: No idea what's happening here. I even went so far as to recreate the `Cargo.lock` file but it still wants to pull in 2.3.1?

tjkirch: What if you use an equals requirement instead of a (default) caret requirement?  (We don't actually want to do that, but it'd be nice to understand.  This isn't a blocker.)

tough/src/editor.rs
zmrow: Oh, wasn't thinking when I answered initially. When a user passes a `Targets` struct to this method, we preserve any `Target` objects in `self.existing_targets`.  When a user adds targets using the `add_target(s)()` methods, they are saved to `self.new_targets`. We combine the two maps when creating the final repository.

I agree that a `clear_targets()` method would be a good idea.

I'm wondering how we best support a `remove_target()` method. The user would need to supply a target filename, since we store this information as `HashMap<filename, Target_object>`

Cargo.lock
tjkirch: I misunderstood caret requirements - `x.y.z` allows `x.*.*` higher versions, not just `x.y.*` higher versions, if the dependency is past 1.0.  I guess I'm used to 0.x dependencies :)

https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#caret-requirements

tough/src/editor.rs
tjkirch: Could wait for a use case / request, but I'd assume they probably have a filename/path, so it could build a Target based on that and remove it from existing/new lists...  doesn't seem urgent if we have clear.

zmrow:
Addressed almost all of @tjk's and @webern's comments.

Open items:

- [ ] Refactor a bit and split out the signed* types to better emphasize the main method of editor (`sign()`)

- [ ] Further editing of targets:
  * `clear_targets()`: Clear both targets HashMaps.
  * ~~What is the best interface for "removing" a single target from the list? By path? By filename? (preferred)~~ (Skipping for now)

- [ ] Better integration test. Create/edit/write/read. (Currently working on this one)

- [x] Unifying the file naming logic - outside the scope of this PR. Will create an issue and link here. --> https://github.com/awslabs/tough/issues/126

tjkirch:
> What is the behavior of `clear_targets()`? Will it clear the targets added from the Targets object? or targets added by path? Differentiating sounds weird and has a crappy interface.

I assumed it would set new and existing to `Some(HashMap::new())`.  Really clear all targets.  No?

> What is the best interface for "removing" a single target from the list? By path? By filename? (preferred)

I think we could/should skip this for now; we don't have a use case or a clear choice.

> Open items:

I'm also hoping we can split up editor.rs a bit, as we discussed offline - the signed* types seem like an easy split, and we could emphasize the "main" method of the editor, `sign`, more easily then.

tough/src/editor.rs
```diff
@@ -0,0 +1,744 @@
...
+    pub fn link_targets<P1, P2>(&self, indir: P1, outdir: P2) -> Result<()>
+    where
+        P1: AsRef<Path>,
+        P2: AsRef<Path>,
+    {
+        let outdir = outdir.as_ref();
+        std::fs::create_dir_all(outdir).context(error::DirCreate { path: outdir })?;
+        // Get the absolute path of the indir and outdir
+        let abs_indir = std::fs::canonicalize(&indir).context(error::AbsolutePath {
+            path: indir.as_ref().to_owned(),
```
tjkirch:
```suggestion
            path: indir.as_ref(),
```
Shouldn't need that, same as outdir, right?

tough/src/editor.rs
```diff
@@ -0,0 +1,744 @@
...
+        let walker = WalkDir::new(&abs_indir).follow_links(true);
+        for entry in walker {
+            let entry = entry.context(error::WalkDir {
+                directory: &abs_indir,
+            })?;
+
+            // If the entry is a file...
+            if !entry.file_type().is_file() {
+                continue;
+            } else {
```
tjkirch:
```suggestion
                continue;
            }
```
You don't need all the code below to be indented for this.  The `continue` will move to the next iteration and skip it all.

zmrow:
You're right about the split, I was gonna introduce that later. :) I edited my comment to reflect the work.

tough/src/editor.rs
zmrow: Good catch - I missed that one.

tough/src/editor.rs
zmrow: 😳 

Sorry about that... This is a silly mistake.

tough/src/editor.rs
zmrow: https://github.com/awslabs/tough/issues/126

tough/src/editor.rs
zmrow: https://github.com/awslabs/tough/issues/126

tough/src/editor.rs
zmrow: Agreed. I'll implement `clear_targets()` for now.

zmrow:
This push addresses the remaining open items except for the refactor into separate files. I'll do that in another push to make it easier to review.

- [ ] Refactor a bit and split out the signed* types to better emphasize the main method of editor (`sign()`)

- [x] Further editing of targets:
  * `clear_targets()`: Clear both targets HashMaps.
  * ~~What is the best interface for "removing" a single target from the list? By path? By filename? (preferred)~~ (Skipping for now)

- [x] Better integration test. Create/edit/write/read.

- [x] Unifying the file naming logic - outside the scope of this PR. Will create an issue and link here. --> https://github.com/awslabs/tough/issues/126

zmrow:
Appeased clippy... angry little fellow he is...  :)

tjkirch (approved):
LGTM - hoping to get the file split discussed above so that the editor code stands out clearly, but I'm sure that'll work fine.

zmrow:
This push is a refactor into separate modules as per @tjkirch 's request.  It's much nicer now. The only change to the code is that I moved the `RepositoryEditor::sign_role()` method to `SignedRole::new()` which feels a bit more ergonomic. 

To wrap close the loop:
- [x] Refactor a bit and split out the signed* types to better emphasize the main method of editor (`sign()`)

- [x] Further editing of targets:
  * `clear_targets()`: Clear both targets HashMaps.
  * ~~What is the best interface for "removing" a single target from the list? By path? By filename? (preferred)~~ (Skipping for now)

- [x] Better integration test. Create/edit/write/read.

- [x] Unifying the file naming logic - outside the scope of this PR. Will create an issue and link here. --> https://github.com/awslabs/tough/issues/126

tjkirch (approved):

webern (approved):
🎪

zmrow:
No-op push to kick the checks again

PR #135, opened 2020-06-03
Refactor `tuftool` to use `RepositoryEditor`

### LINKED ISSUE #43: Move functionality out of tuftool into a library

zmrow:
*Issue #, if available:*
Fixes #43 

*Description of changes:*
The tl;dr is: "simplify `tuftool` by using `RepositoryEditor`."

The actual list of changes is:
* Add "getters" to `SignedRole` so a user can access, but not change data.
* Use `RepositoryEditor` anywhere a repository is being manipulated inside `tuftool`. This includes the `create` and `refresh` commands.
* **Breaking change**: Move the `sign` command to a subcommand of `tuftool root`. This functionality is primarily meant of creation of the `root.json` anyway.
* **Breaking change**: `tuftool create` now *symlinks* targets to the given directory rather than *copying* them. This is a better use of disk space as repositories or targets get larger. *If we still want to keep the copy functionality we can provide a flag the user can pass?*
* Remove any unused files
* Remove any unused error variants

*Testing done:*
* All unit and integration tests pass (which gives me pretty good confidence that we're creating repos the right way)
* Manually create a repository locally; targets are named correctly and properly symlinked.
* Manually refresh a repository locally; everything looks good here as well.


tuftool/src/error.rs
```diff
@@ -1,6 +1,8 @@
 // Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
+// Not really worried about the memory penalty of large enum variants here
+#![allow(clippy::large_enum_variant)]
```
tjkirch, line 5: What makes it large?  Nothing particularly stands out to me, and it seems smaller than it did before...

tuftool/src/root.rs
```diff
@@ -85,6 +87,13 @@ pub(crate) enum Command {
         #[structopt(short = "r", long = "role")]
         roles: Vec<RoleType>,
     },
+    /// Set the signature count threshold for a role
+    Sign {
```
tjkirch: I think the comment was copypasta :)

tuftool/src/root.rs
```diff
@@ -262,6 +272,21 @@ impl Command {
...
+
+        let signed_root = SignedRole::new(
+            root.signed.clone(),
+            &root.signed,
+            &[key_source],
+            &SystemRandom::new(),
+        )
+        .context(error::SignRoot { path })?;
+
+        std::fs::write(&path, signed_root.buffer()).context(error::FileWrite { path: &path })?;
```
tjkirch: Would be pretty funny to partially overwrite someone's root key with half of the signed data :D  Perhaps we should write to a temp file and move it over the original path?  (nit)

tuftool/src/create.rs
```diff
@@ -77,202 +63,49 @@ pub(crate) struct CreateArgs {
...
+            .timestamp_expires(self.timestamp_expires);
+
+        let signed_repo = editor.sign(&self.keys).context(error::SignRepo)?;
+
+        let metadata_dir = &self.outdir.join("metadata");
+        let targets_dir = &self.outdir.join("targets");
+        signed_repo.write(metadata_dir).context(error::WriteRepo)?;
+        signed_repo
+            .link_targets(&self.indir, targets_dir)
+            .context(error::LinkTarget {
```
tjkirch:
```suggestion
            .context(error::LinkTargets {
```

tuftool/src/error.rs
```diff
@@ -253,6 +254,12 @@ pub(crate) enum Error {
         backtrace: Backtrace,
     },
 
+    #[snafu(display("Failed writing repo data to disk: {}", source))]
+    WriteRepo {
```
tjkirch, line 284: Can we include the directory where we were trying to write the repo?

tuftool/src/create.rs
```diff
@@ -77,202 +63,49 @@ pub(crate) struct CreateArgs {
 
 impl CreateArgs {
     pub(crate) fn run(&self) -> Result<()> {
-        if let Some(jobs) = self.jobs {
-            rayon::ThreadPoolBuilder::new()
```
tjkirch, line 81: rayon didn't seem to actually be used in refresh.rs, so I understand that removal, but I think it's used here in create to parallelize building targets, and I think we're losing that with the move to the repo editor... right?  Does it take a lot longer to create repos?  I wonder if we should include that as an improvement to repo editor in another commit for this PR?

tuftool/src/create.rs
```diff
@@ -77,202 +63,49 @@ pub(crate) struct CreateArgs {
...
+            .snapshot_expires(self.snapshot_expires)
+            .timestamp_version(self.timestamp_version)
+            .timestamp_expires(self.timestamp_expires);
+
+        let signed_repo = editor.sign(&self.keys).context(error::SignRepo)?;
+
+        let metadata_dir = &self.outdir.join("metadata");
+        let targets_dir = &self.outdir.join("targets");
+        signed_repo.write(metadata_dir).context(error::WriteRepo)?;
+        signed_repo
```
tjkirch: Should we link targets before we write metadata so the metadata is valid/usable as soon as it's there?

tuftool/src/create.rs
```diff
@@ -77,202 +63,49 @@ pub(crate) struct CreateArgs {
...
-        let mut length = 0;
-        loop {
-            match file.read(&mut buf).context(error::FileRead { path })? {
-                0 => break,
-                n => {
-                    digest.update(&buf[..n]);
-                    length += n as u64;
-                }
+    // Build a list of files from the given directory
+    fn build_targets(&self) -> Result<Vec<PathBuf>> {
```
tjkirch:
```suggestion
    fn find_targets(&self) -> Result<Vec<PathBuf>> {
```
Not really building anymore...

tjkirch (commented):
Look at all that beautiful red!

tuftool/src/error.rs
zmrow: Clippy is angry that some of the enum variants are larger than others. I really think it's overly pedantic in this case and for a CLI tool I don't think we care about this.
```
warning: large size difference between variants
   --> tuftool/src/error.rs:145:5
    |
145 | /     LinkTarget {
146 | |         indir: PathBuf,
147 | |         outdir: PathBuf,
148 | |         source: tough::error::Error,
149 | |         backtrace: Backtrace,
150 | |     },
    | |_____^
```

tuftool/src/root.rs
zmrow: Oops

tuftool/src/root.rs
zmrow: Good point. :) I'll fix that.

tuftool/src/create.rs
zmrow: I'll change it as you're right, but I can't see there being a race condition of any kind here unless a user is using their local machine as a "real" repo and polling for updates constantly. That being said... we have no idea what a user is going to do. :)

zmrow:
Fixes @tjkirch 's comments except for the multithreading question and the clippy comment.

zmrow:
@tjkirch @iliana : This PR symlinks the targets to the given outdir rather than copying them. How do you feel about that? Should we provide a flag so the user can choose to copy the targets?

tjkirch:
> @tjkirch @iliana : This PR symlinks the targets to the given outdir rather than copying them. How do you feel about that? Should we provide a flag so the user can choose to copy the targets?

You know I like it, I requested it :)  A flag would be nice, I guess, but I'm not sure of a reason for it yet.

tuftool/src/create.rs
tjkirch: Per offline discussion, tough shouldn't own the parallelism, but tuftool could/should.  If we can add `Target`s individually to the editor, rather than just paths, then we can build `Target`s in parallel in tuftool and add them to the editor as they're ready.  I don't think it has to be done here, but it does feel unfortunate to lose the parallelism with the change...

zmrow:
> > @tjkirch @iliana : This PR symlinks the targets to the given outdir rather than copying them. How do you feel about that? Should we provide a flag so the user can choose to copy the targets?
> 
> You know I like it, I requested it :) A flag would be nice, I guess, but I'm not sure of a reason for it yet.

Unless @iliana has different feelings, I'll leave it out for now.

tuftool/src/create.rs
webern: We should probably add Windows to the build matrix. Easy to do, seems relevant with respect to symlinks, unless they 'just work' the same over there.

webern (approved):
Wow, now I see why the editor is a good thing!

zmrow:
The above commit adds the threading support back into `tuftool create`. I ended up being able to use a decent amount of the original code, so I got that going for me too, which is nice.

zmrow:
Rebased onto `develop`

tough/src/editor/mod.rs
```diff
@@ -189,8 +189,21 @@ impl RepositoryEditor {
         Ok(self)
     }
 
+    /// Add a `Target` to the repository
+    pub fn add_target(&mut self, name: String, target: Target) -> Result<&mut Self> {
```
tjkirch: It won't let me comment below, but we should be able to use this at the end of `add_target_path` now.

tough/src/editor/test.rs
```diff
@@ -58,19 +58,23 @@ mod tests {
...
     #[test]
-    fn add_targets() {
+    fn add_targets_from_multiple_sources() {
         let targets: Signed<Targets> = serde_json::from_str(include_str!(
             "../../tests/data/tuf-reference-impl/metadata/targets.json"
         ))
         .unwrap();
-        let target3 = targets_path().join("file3.txt");
+        let target3_path = targets_path().join("file3.txt");
+        let target2_path = targets_path().join("file2.txt");
```
tjkirch, line 67: Hmm, should we name this something other than what's in the existing targets?  Or maybe test both, replacing an existing target and adding a new one from a `Target`?

tuftool/src/refresh.rs
```diff
@@ -75,124 +68,43 @@ pub(crate) struct RefreshArgs {
 
 impl RefreshArgs {
     pub(crate) fn run(&self) -> Result<()> {
-        if let Some(jobs) = self.jobs {
```
tjkirch, line 78: It won't let me comment on the new location of this, but it made me think - the comment on `jobs` says it defaults to the number of cores, but there's no structopt-level default, and here we're checking if the argument was given... are we actually getting parallelism?  Do we need a default in the args struct?

tjkirch (approved):

tough/src/editor/mod.rs
zmrow: ha good catch - I'll fix it

tuftool/src/refresh.rs
zmrow: Good question. Per the [`rayon` FAQ](https://github.com/rayon-rs/rayon/blob/master/FAQ.md), it will spawn the same number of threads as cores. It does this by default when you use any of the parallel iterator methods it provides, which we do below in `build_targets()`.

If a user specifies a specific number of jobs, we use `rayon::ThreadPoolBuilder` to override the default.

tjkirch: Ah, sorry, I was silly and forgot that rayon creates a global thread pool by default.  This is just configuring it, if the user wanted a different number of threads.

zmrow: I'll add a comment for future readers.

tough/src/editor/test.rs
zmrow: I'll rename this to something else like "file4.txt"

zmrow:
Addressed @tjkirch 's comments

tough/src/editor/mod.rs
tjkirch: That change made me realize that this function doesn't have to return a result - we can always insert into the hash, and so we can remove the burden of error checking from callers.

zmrow: You're right... tunnel vision over here. I'll fix that before I merge.

zmrow:
Removed the `Result<>` from the `add_target()` method. Good catch @tjkirch !

zmrow:
Fixed the clippy issue.

tjkirch (approved):

PR #139, opened 2020-06-08
Add "update" command to tuftool

### LINKED ISSUE #136: tuftool should be able to update an existing repository
`tuftool` can already refresh metadata for an existing repository. It should have the ability to add targets and re-sign the metadata for an existing repository.

I see this as having mostly the same interface as `refresh`, except with an extra argument `--new-targets` which is where we'll find any new targets.

zmrow:
`tuftool update` will allow a user to update an existing repository,
adding new targets, versions, and expiration dates. It will update
the metadata and sign it. Then it will symlink the targets and write
the updated metadata to the given directory.

*Issue #, if available:*
Fixes #136 

*Description of changes:*
* Add `update` module and unit test
* Add additional sample targets separate from the tuf reference impl.
* Move `build_targets()` and `process_target()` to main so `create` and `update` can share them

*Testing done:*
* Unit and integration tests all pass
* Manually update a repo on disk. Verify the new repo gets the correct metadata and the *only* new targets are symlinked.

### DIFF
--- a/tuftool/src/main.rs
+++ b/tuftool/src/main.rs
@@ -18,14 +18,19 @@ mod error;
 mod refresh;
 mod root;
 mod source;
+mod update;
 
 use crate::error::Result;
+use rayon::prelude::*;
 use snafu::{ErrorCompat, OptionExt, ResultExt};
+use std::collections::HashMap;
 use std::fs::File;
 use std::io::Write;
 use std::path::Path;
 use structopt::StructOpt;
 use tempfile::NamedTempFile;
+use tough::schema::Target;
+use walkdir::WalkDir;
 
 static SPEC_VERSION: &str = "1.0.0";
 
@@ -39,6 +44,8 @@ enum Command {
     Refresh(refresh::RefreshArgs),
     /// Download a TUF repository's resources
     Download(download::DownloadArgs),
+    /// Update a TUF repository's metadata and optionally add targets
+    Update(update::UpdateArgs),
 }
 
 impl Command {
@@ -48,6 +55,7 @@ impl Command {
             Command::Root(root_subcommand) => root_subcommand.run(),
             Command::Refresh(args) => args.run(),
             Command::Download(args) => args.run(),
+            Command::Update(args) => args.run(),
         }
     }
 }
@@ -72,6 +80,44 @@ where
     writer.write_all(b"\n").context(error::FileWrite { path })?;
     writer.persist(path).context(error::FilePersist { path })?;
     Ok(())
+}
+
+// Walk the directory specified, building a map of filename to Target structs.
+// Hashing of the targets is done in parallel
+fn build_targets<P>(indir: P, follow_links: bool) -> Result<HashMap<String, Target>>
+where
+    P: AsRef<Path>,
+{
+    let indir = indir.as_ref();
+    WalkDir::new(indir)
+        .follow_links(follow_links)
+        .into_iter()
+        .par_bridge()
+        .filter_map(|entry| match entry {
+            Ok(entry) => {
+                if entry.file_type().is_file() {
+                    Some(process_target(entry.path()))
+                } else {
+                    None
+                }
+            }
+            Err(err) => Some(Err(err).context(error::WalkDir { directory: indir })),
+        })
+        .collect()
+}
+
+fn process_target(path: &Path) -> Result<(String, Target)> {
+    // Build a Target from the path given. If it is not a file, this will fail
+    let target = Target::from_path(path).context(error::TargetFromPath { path })?;
+
+    // Get the file name as a string
+    let target_name = path
+        .file_name()
+        .context(error::NoFileName { path })?
+        .to_str()
+        .context(error::PathUtf8 { path })?
+        .to_owned();
+    Ok((target_name, target))
 }
 
 fn main() -> ! {
--- a/tuftool/src/update.rs
+++ b/tuftool/src/update.rs
@@ -7,14 +7,18 @@ use crate::error::{self, Result};
 use crate::source::parse_key_source;
 use chrono::{DateTime, Utc};
 use snafu::ResultExt;
+use std::fs::File;
 use std::num::{NonZeroU64, NonZeroUsize};
 use std::path::PathBuf;
 use structopt::StructOpt;
+use tempfile::tempdir;
 use tough::editor::RepositoryEditor;
 use tough::key_source::KeySource;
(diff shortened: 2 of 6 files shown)


tuftool/src/main.rs
```diff
@@ -39,6 +44,8 @@ enum Command {
     Refresh(refresh::RefreshArgs),
     /// Download a TUF repository's resources
     Download(download::DownloadArgs),
+    /// Download a TUF repository's resources
+    Update(update::UpdateArgs),
```
tjkirch: Copypasta doc comment

tuftool/src/update.rs
```diff
@@ -0,0 +1,142 @@
...
+    /// Number of target hashing threads to run (default: number of cores)
+    // No default is specified in structopt here. This is because rayon
+    // automatically spawns the same number of threads as cores when any
+    // of its parallel methods are called.
+    #[structopt(short = "j", long = "jobs")]
+    jobs: Option<NonZeroUsize>,
+
+    /// Key files to sign with
+    #[structopt(short = "k", long = "key", required = true, parse(try_from_str = parse_key_source))]
+    keys: Vec<Box<dyn KeySource>>,
```
tjkirch, line 35: Why is it `keys` and "files" and `Vec` when `parse_key_source` returns a single `Box<dyn KeySource>`?  I feel like I'm missing something, does structopt automatically handle multiple arguments?

tuftool/src/update.rs
```diff
@@ -0,0 +1,142 @@
...
+                .context(error::InitializeThreadPool)?;
+        }
+
+        let new_targets = build_targets(&self.indir, self.follow)?;
+
+        let settings = tough::Settings {
+            root: File::open(&self.root).context(error::FileOpen { path: &self.root })?,
+            datastore: self.workdir.as_path(),
+            metadata_base_url: self.metadata_base_url.as_str(),
+            targets_base_url: self.metadata_base_url.as_str(),
```
tjkirch: Are these intentionally the same?  I'm missing something.  Are we assuming "metadata" and "targets" are subdirectories of the given URL, the way they'd come out from the `create` command?  I don't understand how the repo loads successfully.

tuftool/src/update.rs
```diff
@@ -0,0 +1,142 @@
...
+    #[structopt(short = "r", long = "root")]
+    root: PathBuf,
+
+    /// TUF repository metadata base URL
+    #[structopt(short = "m", long = "metadata-url")]
+    metadata_base_url: Url,
+
+    /// The working directory where the current metadata files will be written.
+    #[structopt(short = "w", long = "workdir", default_value = ".")]
+    workdir: PathBuf,
```
tjkirch: It looks like this is just a temporary location for tough's data store?  Can we use a tempdir instead of making the user specify?

tuftool/src/update.rs
zmrow: I followed the lead of the other modules that include this argument. Structopt allows a user to pass [multiple identical arguments](https://docs.rs/structopt/0.3.14/structopt/index.html#type-magic) by parsing the type; `Vec` in this case.

I don't think it will be very common, but using a `Vec` here allows a user to pass multiple `-k` arguments with different key sources.

tuftool/src/update.rs
zmrow: In this case, since we aren't ever loading targets from the repo, we don't really care what the `targets_base_url` is. `tough::Settings` requires a value, so rather than asking the user for something we're never going to use, we use a placeholder of `metadata_base_url`.

I'll add a comment about this the clarify.

tuftool/src/update.rs
zmrow: I could see situations where a user may want to inspect those files as they were.

How about keeping the argument, but using a tempdir if the user *doesn't* pass the arg?

zmrow:
Address @tjkirch 's comments

tuftool/src/main.rs
```diff
@@ -39,6 +44,8 @@ enum Command {
     Refresh(refresh::RefreshArgs),
     /// Download a TUF repository's resources
     Download(download::DownloadArgs),
+    /// Update a TUF repository's metadata and targets
```
tjkirch:
```suggestion
    /// Update a TUF repository's metadata and optionally add targets
```
Feels like it's good to clarify that adding targets is the thing they can do; don't want someone to think it's a full replacement of targets.

tuftool/src/update.rs
tjkirch: I think wanting to inspect those files is... odd?  They can inspect the original repo and the resulting repo; what's left to care about?

If you think it's important, how about clarifying the flag name to show that it persists the data store to a given path?  It's just a rename of workdir I think - `--persist-tough-data-store PATH` or something?

zmrow: True - I guess a user can just inspect that actual repo. I'll remove the arg.

tjkirch (commented):
After looking at this again, I have to ask - are we sure this should be a separate subcommand?  It's basically `refresh` plus adding targets.  I think it'd make sense to me to remove `refresh`, and have the targets in `update` be optional... (refresh and update mean basically the same here...)

It's also quite similar to `create`, just with a `from_repo` instead of a `new`...

iliana:
I agree with @tjkirch but also think that deprecating/removing the `refresh` command can be done in a separate pull request.

tjkirch:
How about:
<code>tuftool repo --output-dir bar --input-dir foo --add-targets</code> = `update`
<code>tuftool repo --output-dir bar &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; --add-targets</code> = `create`
<code>tuftool repo --output-dir bar --input-dir foo &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;</code> = `refresh`

Obviously output directory is always required, so it could be positional, but I guess I'm biased toward "options".

zmrow:
> I agree with @tjkirch but also think that deprecating/removing the `refresh` command can be done in a separate pull request.

Opened #140 to track this work

zmrow:
Addressed @tjkirch 's comments for this PR

zmrow:
Fixed the broken test

tjkirch (approved):

webern (approved):
Nice work!

PR #145, opened 2020-06-11
Update dependencies

zmrow:
*Issue #, if available:*
N/A

*Description of changes:*
* ran `cargo update`

All tests continue to pass and project continues to build just fine.


webern (changes requested):
This doesn't seem right. Only the lock file changed? Maybe cargo update isn't the right command, the Cargo.toml files should be updated to the latest minor versions.

webern (approved):
Per our discussion, this is apparently all that's needed.

PR #155, opened 2020-06-18
Integration of delegated targets described in the TUF spec for repo loading

ecpullen:
**This is the first part of the integration of delegated targets with tough, featuring the loading of delegated targets of an existing repository. An upcoming PR will contain the additions for creating/delegating targets and signing delegated targets.**

*Issue #, if available:*
Related to #5 

*Description of changes:*
* Delegations Struct for serialization and deserialization
* DelegatedRole manages the metadata that was delegated and the targets struct that was delegated
* Delegations describe paths using PathSet, which is either a set of shell style paths or path hash prefixes
* Delegated targets are loaded in a pre-order as specified by the TUF spec

*Testing Done*
* Integration with current tests

**Thorough testing requires complex delegations structures that need to be signed properly. This will be completed after the write and sign portion is completed**

### DIFF
--- a/tough/src/cache.rs
+++ b/tough/src/cache.rs
@@ -69,7 +69,7 @@ impl<'a, T: Transport> Repository<'a, T> {
             &metadata_outdir,
         )?;
 
-        for name in self.targets.signed.roles_str() {
+        for name in self.targets.signed.role_names() {
             self.cache_file_from_transport(
                 self.delegated_filename(name).as_str(),
                 self.limits.max_targets_size,
--- a/tough/src/cache.rs
+++ b/tough/src/cache.rs
@@ -42,7 +42,7 @@ impl<'a, T: Transport> Repository<'a, T> {
                 self.cache_target(&targets_outdir, target_name.as_ref())?;
             }
         } else {
-            let targets = &self.targets.signed.get_targets_map();
+            let targets = &self.targets.signed.targets_map();
             for target_name in targets.keys() {
                 self.cache_target(&targets_outdir, target_name)?;
             }
@@ -69,9 +69,9 @@ impl<'a, T: Transport> Repository<'a, T> {
             &metadata_outdir,
         )?;
 
-        for name in self.targets.signed.get_roles_str() {
+        for name in self.targets.signed.roles_str() {
             self.cache_file_from_transport(
-                &format!("{}.json", name),
+                self.delegated_filename(name).as_str(),
                 self.limits.max_targets_size,
                 "max_targets_size argument",
                 &metadata_outdir,
@@ -110,6 +110,15 @@ impl<'a, T: Transport> Repository<'a, T> {
         }
     }
 
+    /// Prepends the version number to the role.json filename if using consistent snapshot mode.
+    fn delegated_filename(&self, name: &str) -> String {
+        if self.root.signed.consistent_snapshot {
+            format!("{}.{}.json", self.targets.signed.version, name)
+        } else {
+            format!("{}.json", name)
+        }
+    }
+
     /// Copies a file using `Transport` to `outdir`.
     fn cache_file_from_transport<P: AsRef<Path>>(
         &self,
@@ -145,7 +154,13 @@ impl<'a, T: Transport> Repository<'a, T> {
     /// Saves a signed target to the specified `outdir`. Retains the digest-prepended filename if
     /// consistent snapshots are used.
     fn cache_target<P: AsRef<Path>>(&self, outdir: P, name: &str) -> Result<()> {
-        let t = self.targets.signed.find_target(name).unwrap();
+        let t = self
+            .targets
+            .signed
+            .find_target(name)
+            .context(error::CacheTargetMissing {
+                target_name: name.to_owned(),
+            })?;
         let (sha, filename) = self.target_digest_and_filename(&t, name);
         let mut reader = self.fetch_target(t, &sha, filename.as_str())?;
         let path = outdir.as_ref().join(filename);
--- a/tough/src/cache.rs
+++ b/tough/src/cache.rs
@@ -49,7 +49,6 @@ impl<'a, T: Transport> Repository<'a, T> {
         }
 
         // Save the snapshot, targets and timestamp metadata files, and (optionally) the root files.
-        // TODO - support role delegation https://github.com/awslabs/tough/issues/5
         self.cache_file_from_transport(
             self.snapshot_filename().as_str(),
             self.max_snapshot_size()?,
@@ -70,13 +69,16 @@ impl<'a, T: Transport> Repository<'a, T> {
         )?;
 
         for name in self.targets.signed.role_names() {
-            self.cache_file_from_transport(
-                self.delegated_filename(name).as_str(),
-                self.limits.max_targets_size,
-                "max_targets_size argument",
-                &metadata_outdir,
-            )?;
+            if let Some(filename) = self.delegated_filename(name) {
+                self.cache_file_from_transport(
+                    filename.as_str(),
+                    self.limits.max_targets_size,
+                    "max_targets_size argument",
+                    &metadata_outdir,
+                )?;
+            }
         }
+
         if cache_root_chain {
             // Copy all versions of root.json less than or equal to the current version.
             for ver in (1..=self.root.signed.version.get()).rev() {
@@ -111,11 +113,19 @@ impl<'a, T: Transport> Repository<'a, T> {
     }
 
     /// Prepends the version number to the role.json filename if using consistent snapshot mode.
-    fn delegated_filename(&self, name: &str) -> String {
+    fn delegated_filename(&self, name: &str) -> Option<String> {
         if self.root.signed.consistent_snapshot {
-            format!("{}.{}.json", self.targets.signed.version, name)
+            Some(format!(
+                "{}.{}.json",
+                self.snapshot
+                    .signed
+                    .meta
+                    .get(&format!("{}.json", name))?
+                    .version,
+                name
+            ))
         } else {
-            format!("{}.json", name)
+            Some(format!("{}.json", name))
         }
     }
 
--- a/tough/src/cache.rs
+++ b/tough/src/cache.rs
@@ -42,7 +42,7 @@ impl<'a, T: Transport> Repository<'a, T> {
                 self.cache_target(&targets_outdir, target_name.as_ref())?;
             }
         } else {
-            let targets = &self.targets.signed.targets;
+            let targets = &self.targets.signed.get_targets_map();
             for target_name in targets.keys() {
                 self.cache_target(&targets_outdir, target_name)?;
             }
@@ -68,6 +68,15 @@ impl<'a, T: Transport> Repository<'a, T> {
             "max_timestamp_size argument",
             &metadata_outdir,
         )?;
+
+        for name in self.targets.signed.get_roles_str() {
+            self.cache_file_from_transport(
+                &format!("{}.json", name),
+                self.limits.max_targets_size,
+                "max_targets_size argument",
+                &metadata_outdir,
+            )?;
+        }
         if cache_root_chain {
             // Copy all versions of root.json less than or equal to the current version.
             for ver in (1..=self.root.signed.version.get()).rev() {
@@ -136,14 +145,7 @@ impl<'a, T: Transport> Repository<'a, T> {
     /// Saves a signed target to the specified `outdir`. Retains the digest-prepended filename if
     /// consistent snapshots are used.
     fn cache_target<P: AsRef<Path>>(&self, outdir: P, name: &str) -> Result<()> {
-        let t = self
-            .targets
-            .signed
-            .targets
-            .get(name)
-            .context(error::CacheTargetMissing {
-                target_name: name.to_owned(),
-            })?;
+        let t = self.targets.signed.find_target(name).unwrap();
         let (sha, filename) = self.target_digest_and_filename(&t, name);
         let mut reader = self.fetch_target(t, &sha, filename.as_str())?;
         let path = outdir.as_ref().join(filename);
--- a/tough/src/error.rs
+++ b/tough/src/error.rs
@@ -348,6 +348,7 @@ pub enum Error {
     #[snafu(display("The target '{}' was not found", target_name))]
     CacheTargetMissing {
         target_name: String,
+        source: crate::schema::Error,
         backtrace: Backtrace,
     },
 
@@ -360,6 +361,25 @@ pub enum Error {
 
     #[snafu(display("Target File not Delegated: {}", target_url))]
     TargetNotFound { target_url: String },
+
+    #[snafu(display("Delegated role not found: {}", name))]
+    DelegateNotFound { name: String },
+
+    #[snafu(display("Delegation didn't contain targets field"))]
+    NoTargets {},
+
+    #[snafu(display("Targets didn't contain delegations field"))]
+    NoDelegations {},
+
+    #[snafu(display("Delegated roles are not consistent for {}", name))]
+    DelegatedRolesNotConsistent { name: String },
+
+    /// Target doesn't have proper permissions from parent delegations
+    #[snafu(display("Invalid file permissions"))]
+    InvalidPath { source: crate::schema::Error },
+
+    #[snafu(display("Role Missing from snapshot meta: {}", name))]
+    RoleNotInMeta { name: String },
 }
 
 // used in `std::io::Read` implementations
--- a/tough/src/error.rs
+++ b/tough/src/error.rs
@@ -369,10 +369,10 @@ pub enum Error {
     },
 
     #[snafu(display("Delegation doesn't contain targets field"))]
-    NoTargets {},
+    NoTargets,
 
     #[snafu(display("Targets doesn't contain delegations field"))]
-    NoDelegations {},
+    NoDelegations,
 
     #[snafu(display("Delegated roles are not consistent for {}", name))]
     DelegatedRolesNotConsistent { name: String },
--- a/tough/src/error.rs
+++ b/tough/src/error.rs
@@ -357,6 +357,9 @@ pub enum Error {
         source: walkdir::Error,
         backtrace: Backtrace,
     },
+
+    #[snafu(display("Target File not Delegated: {}", target_url))]
+    TargetNotFound { target_url: String },
 }
 
 // used in `std::io::Read` implementations
--- a/tough/src/error.rs
+++ b/tough/src/error.rs
@@ -359,16 +359,19 @@ pub enum Error {
         backtrace: Backtrace,
     },
 
-    #[snafu(display("Target File not Delegated: {}", target_url))]
+    #[snafu(display("Target file not delegated: {}", target_url))]
     TargetNotFound { target_url: String },
 
     #[snafu(display("Delegated role not found: {}", name))]
-    DelegateNotFound { name: String },
+    DelegateNotFound {
+        name: String,
+        source: crate::schema::Error,
+    },
 
-    #[snafu(display("Delegation didn't contain targets field"))]
+    #[snafu(display("Delegation doesn't contain targets field"))]
     NoTargets {},
 
-    #[snafu(display("Targets didn't contain delegations field"))]
+    #[snafu(display("Targets doesn't contain delegations field"))]
     NoDelegations {},
 
     #[snafu(display("Delegated roles are not consistent for {}", name))]
@@ -378,7 +381,7 @@ pub enum Error {
     #[snafu(display("Invalid file permissions"))]
     InvalidPath { source: crate::schema::Error },
 
-    #[snafu(display("Role Missing from snapshot meta: {}", name))]
+    #[snafu(display("Role missing from snapshot meta: {}", name))]
     RoleNotInMeta { name: String },
 }
 
--- a/tough/src/lib.rs
+++ b/tough/src/lib.rs
@@ -31,7 +31,7 @@ pub mod schema;
 pub mod sign;
 mod transport;
 
-use crate::schema::{DelegatedRole, Delegations, Target};
+use crate::schema::{DelegatedRole, Delegations};
 #[cfg(feature = "http")]
 pub use crate::transport::HttpTransport;
 pub use crate::transport::{FilesystemTransport, Transport};
@@ -292,8 +292,8 @@ impl<'a, T: Transport> Repository<'a, T> {
     }
 
     ///return a vec of all targets including all target files delegated by targets
-    pub fn all_targets(&self) -> Vec<&Target> {
-        self.targets.signed.targets()
+    pub fn all_targets<'b>(&'b self) -> impl Iterator + 'b {
+        self.targets.signed.targets_iter()
     }
 
     /// Fetches a target from the repository.
@@ -343,14 +343,8 @@ impl<'a, T: Transport> Repository<'a, T> {
         })
     }
 
-    pub fn role(&self, name: &str) -> Result<&DelegatedRole> {
-        Ok(self
-            .targets
-            .signed
-            .delegated_role(name)
-            .context(error::DelegateNotFound {
-                name: name.to_string(),
-            })?)
+    pub fn delegated_role(&self, name: &str) -> Option<&DelegatedRole> {
+        self.targets.signed.delegated_role(name).ok()
     }
 }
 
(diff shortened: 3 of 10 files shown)


tough/src/cache.rs
```diff
@@ -42,7 +42,7 @@ impl<'a, T: Transport> Repository<'a, T> {
                 self.cache_target(&targets_outdir, target_name.as_ref())?;
             }
         } else {
-            let targets = &self.targets.signed.targets;
+            let targets = &self.targets.signed.get_targets_map();
```
webern: Stylisticly in keeping with what I see in Rust land, they don't seem to have the `get_` prefix.

```suggestion
            let targets = &self.targets.signed.targets_map();
```

tough/src/cache.rs
```diff
@@ -136,14 +145,7 @@ impl<'a, T: Transport> Repository<'a, T> {
...
     fn cache_target<P: AsRef<Path>>(&self, outdir: P, name: &str) -> Result<()> {
-        let t = self
-            .targets
-            .signed
-            .targets
-            .get(name)
-            .context(error::CacheTargetMissing {
-                target_name: name.to_owned(),
-            })?;
+        let t = self.targets.signed.find_target(name).unwrap();
```
webern: Need to retain the context(error::CacheTargetMissing), we don't use `unwrap()` in production code (test code is fine).

```suggestion
        let t = self.targets.signed.find_target(name)..context(error::CacheTargetMissing {
                target_name: name.to_owned(),
            })?;
```

tough/src/cache.rs
```diff
@@ -68,6 +68,15 @@ impl<'a, T: Transport> Repository<'a, T> {
             "max_timestamp_size argument",
             &metadata_outdir,
         )?;
+
+        for name in self.targets.signed.get_roles_str() {
+            self.cache_file_from_transport(
+                &format!("{}.json", name),
```
webern: Is consistent snapshot prefixing needed here? I notice above that the targets file comes from a function: `targets_filename()`

```rust
                &format!("{}.json", name),
```

tough/src/lib.rs
```diff
@@ -328,14 +334,21 @@ impl<'a, T: Transport> Repository<'a, T> {
...
-        )
+        Ok(if let Ok(target) = self.targets.signed.find_target(name) {
+            let (sha256, file) = self.target_digest_and_filename(target, name);
+            Some(self.fetch_target(target, &sha256, file.as_str())?)
+        } else {
+            None
+        })
+    }
+
+    pub fn get_role(&self, name: &str) -> Result<&DelegatedRole> {
```
webern: Since root.json, targets.json, etc are all "Roles", maybe this should be called `pub fn delegate(&self, name: &str)`? And the error `DelegateNotFound`?

tough/src/lib.rs
```diff
@@ -874,10 +887,105 @@ fn load_targets<T: Transport>(
...
+        HashMap::new();
+    for del_role in &delegation.roles {
+        //find the role file metadata
+        let role_meta = snapshot
+            .signed
+            .meta
+            .get(&format!("{}.json", &del_role.name))
+            .unwrap();
+
+        let path = format!("{}.json", &del_role.name);
```
webern: Same question as earlier: does consistent snapshots affect the filename by prepending a version for these?

```rust
        let path = format!("{}.json", &del_role.name);
```

tough/src/lib.rs
```diff
@@ -874,10 +887,105 @@ fn load_targets<T: Transport>(
...
+        del_role.targets = delegated_roles.remove(&del_role.name).unwrap();
+        load_delegations(
+            transport,
+            snapshot,
+            metadata_base_url,
+            max_targets_size,
+            &mut del_role
+                .targets
+                .as_mut()
+                .unwrap()
```
webern: As mentioned before, we will need to use Errors and Results instead of panics.

```rust
                .unwrap()
```

tough/src/schema/mod.rs
```diff
@@ -308,7 +311,69 @@ impl Targets {
...
+                    target_url: target_url.to_string(),
+                }),
+                Some(delegations) => delegations.find_target(target_url),
+            },
+        }
+    }
+
+    ///Given the name of a delegated role, return the delegated role
+    pub fn get_del_role(&self, name: &str) -> Result<&DelegatedRole> {
+        self.delegations.as_ref().unwrap().get_del_role(name)
```
webern: Need to do a match case on the Option instead of `unwrap()`. Also something to think about, the return could be an option instead of result, here and in other places. In the standard library, e.g. HashMap `get` returns an `Option` instead of a `Result`. Doesn't seem like a big deal either way though.

```rust
        self.delegations.as_ref().unwrap().get_del_role(name)
```

webern (commented):
@iliana and @tjkirch will hold the most weight. Here are some things I noticed, though I don't yet have a complete grasp on correctness.

tough/src/lib.rs
ecpullen: That is true, and will be updated in an incoming PR along with write and sign capabilities.

tough/src/lib.rs
ecpullen: I don't know if delegate is the term we want to use for the roles that have been delegated as they are still roles and with 2 meanings of the word delegate the naming could create confusion.

tough/src/error.rs
```diff
@@ -357,6 +358,28 @@ pub enum Error {
         source: walkdir::Error,
         backtrace: Backtrace,
     },
+
+    #[snafu(display("Target File not Delegated: {}", target_url))]
```
zmrow: nit: these error messages don't need to be title cased.

tough/src/error.rs
```diff
@@ -357,6 +358,28 @@ pub enum Error {
...
+    TargetNotFound { target_url: String },
+
+    #[snafu(display("Delegated role not found: {}", name))]
+    DelegateNotFound { name: String },
+
+    #[snafu(display("Delegation didn't contain targets field"))]
```
zmrow: nit: This error message's tense isn't consistent with the others. Suggestion: 

`Delegation doesn't contain targets field`

tough/src/error.rs
```diff
@@ -357,6 +358,28 @@ pub enum Error {
...
+    NoTargets {},
+
+    #[snafu(display("Targets didn't contain delegations field"))]
```
zmrow: Same here - message should be present tense.

tough/src/error.rs
```diff
@@ -357,6 +358,28 @@ pub enum Error {
...
+    NoDelegations {},
+
+    #[snafu(display("Delegated roles are not consistent for {}", name))]
+    DelegatedRolesNotConsistent { name: String },
+
+    /// Target doesn't have proper permissions from parent delegations
+    #[snafu(display("Invalid file permissions"))]
+    InvalidPath { source: crate::schema::Error },
+
+    #[snafu(display("Role Missing from snapshot meta: {}", name))]
```
zmrow: nit: don't need to capital "M"

tough/src/schema/mod.rs
```diff
@@ -308,7 +311,72 @@ impl Targets {
             expires,
             targets: HashMap::new(),
             _extra: HashMap::new(),
+            delegations: None,
+        }
+    }
+
+    ///Given a target url, returns a reference to the Target struct or error if the target is unreachable
+    pub fn find_target(&self, target_url: &str) -> Result<&Target> {
```
zmrow: `Targets.targets` is a map of target name (a string) to `Target`. We probably want to change `target_url` to `target_name` or something similar.

tough/src/schema/mod.rs
```diff
@@ -308,7 +311,72 @@ impl Targets {
...
+                None => Err(Error::TargetNotFound {
+                    target_url: target_url.to_string(),
+                }),
+                Some(delegations) => delegations.find_target(target_url),
+            },
+        }
+    }
+
+    ///Given the name of a delegated role, return the delegated role
+    pub fn del_role(&self, name: &str) -> Result<&DelegatedRole> {
```
zmrow: I'd probably skip the abbreviation of "delegated" here and just name the function `get_delegated_role()` or, since this is a sort of "getter", `delegated_role()`.

When I first read the function name, I immediately thought "delete role". :)

tough/src/schema/mod.rs
```diff
@@ -308,7 +311,72 @@ impl Targets {
...
+        Err(Error::NoDelegations {})
+    }
+
+    ///Returns a vec of all targets and all delegated targets recursively
+    pub fn targets(&self) -> Vec<&Target> {
+        let mut targets = Vec::new();
+        for target in &self.targets {
+            targets.push(target.1);
+        }
+        if let Some(del) = &self.delegations {
```
zmrow: Same here - I'd use `delegations` rather than `del`. Yeah it's longer but its a bit more clear to the reader.

tough/src/schema/mod.rs
```diff
@@ -308,7 +311,72 @@ impl Targets {
...
+            for t in del.targets() {
+                targets.push(t);
+            }
         }
+
+        targets
+    }
+
+    ///Returns a vec of all targets and all delegated targets recursively
```
zmrow: This function returns a HashMap. :)

tough/src/schema/mod.rs
```diff
@@ -308,7 +311,72 @@ impl Targets {
...
+        }
+        if let Some(del) = &self.delegations {
+            targets.extend(del.targets_map());
+        }
+
+        targets
+    }
+
+    ///Returns a vec of all rolenames
+    pub fn roles_str(&self) -> Vec<&String> {
```
zmrow: I might name this `role_names()`

tough/src/schema/mod.rs
```diff
@@ -324,6 +392,215 @@ impl Role for Targets {
...
+            re.is_match(&target)
+        } else {
+            false
+        }
+    }
+}
+
+impl Delegations {
+    ///Determines if target passes pathset specific matching
+    pub fn check_target(&self, target: &str) -> bool {
```
zmrow: `check_target()` feels a bit generic to me and not really descriptive of what the function does.  Since we're return a `bool`, perhaps `target_matches_path()` or something similar?

tough/src/schema/mod.rs
```diff
@@ -324,6 +392,215 @@ impl Role for Targets {
...
+                return true;
+            }
+        }
+        false
+    }
+
+    ///Ensures that all delegated paths are allowed to be delegated
+    pub fn verify_paths(&self) -> Result<()> {
+        for sub_role in &self.roles {
+            for path in match &sub_role.paths {
```
zmrow: nit: To make this a little bit clearer, I might just assign the output of the match to a variable and then iterate through that rather than doing it all at once.

tough/src/schema/mod.rs
```diff
@@ -324,6 +392,215 @@ impl Role for Targets {
...
+
+    ///verifies that roles matches contain valid keys
+    pub fn verify_role(&self, role: &Signed<Targets>, name: &str) -> Result<()> {
+        let role_keys = self.role(name).expect("Role not found");
+        let mut valid = 0;
+
+        let mut data = Vec::new();
+        let mut ser = serde_json::Serializer::with_formatter(&mut data, CanonicalFormatter::new());
+        role.signed
+            .serialize(&mut ser)
```
zmrow, line 518: I'm a bit confused. Why are we serializing this data?

tough/src/schema/mod.rs
```diff
@@ -324,6 +392,215 @@ impl Role for Targets {
...
+                threshold: role_keys.threshold,
+                valid,
+            }
+        );
+        Ok(())
+    }
+
+    ///Finds target using pre ordered search given `target_url` or error if the target is not found
+    pub fn find_target(&self, target_url: &str) -> Result<&Target> {
+        for del_role in &self.roles {
```
zmrow: I'd avoid the abbreviation here again: `delegated_role` rather than `del_role`.

tough/src/schema/mod.rs
```diff
@@ -324,6 +392,215 @@ impl Role for Targets {
...
+            match &del_role.targets {
+                Some(targets) => match &targets.signed.find_target(target_url) {
+                    Ok(target) => return Ok(target),
+                    _ => continue,
+                },
+                None => continue,
```
zmrow: This can probably be simplified a bit and avoid some nested matching by doing:
```
if let Some(targets) = &delegated_role.targets {
...
}
```
You can then eliminate the `None => continue` and the loop will continue.

tough/src/schema/mod.rs
```diff
@@ -324,6 +392,215 @@ impl Role for Targets {
...
+            }
+        }
+        Err(Error::TargetNotFound {
+            target_url: target_url.to_string(),
+        })
+    }
+
+    ///Given a role name recursively searches for the delegated role
+    pub fn del_role(&self, name: &str) -> Result<&DelegatedRole> {
```
zmrow: s/del_role/delegated_role/g

:)

zmrow (commented):
Nice work.

From a code perspective theres a few simplifications to make and some things to tweak to make it a bit clearer. @iliana may be the authority on the implementation correctness; I'll also try to run back through to verify it.

tough/src/schema/mod.rs
ecpullen: Now that you say that, I'm not sure. I thought it was to make sure it was loaded properly, but serde wouldn't allow it to load wrong anyway.

Edited:
We serialize so that we can use the data to verify the keys.

tough/src/lib.rs
```diff
@@ -289,6 +291,10 @@ impl<'a, T: Transport> Repository<'a, T> {
         &self.timestamp
     }
 
+    pub fn delegated_targets(&self) -> Vec<&Target> {
```
zmrow: I think this function should probably be renamed to `all_targets()` or something similar since it returns all the targets, not just targets that have been delegated.

tough/src/lib.rs
```diff
@@ -328,14 +334,21 @@ impl<'a, T: Transport> Repository<'a, T> {
...
+        } else {
+            None
+        })
+    }
+
+    pub fn role(&self, name: &str) -> Result<&DelegatedRole> {
+        match self.targets.signed.delegated_role(name) {
+            Ok(delegated_role) => Ok(delegated_role),
+            _ => Err(error::Error::DelegateNotFound {
+                name: name.to_string(),
```
zmrow: I think we can simplify this a little bit and avoid the `match`:

```
self.targets.signed.delegated_role(name).context(error...)?
```

zmrow (commented):
Just a few additional comments.

tough/src/cache.rs
```diff
@@ -101,6 +110,15 @@ impl<'a, T: Transport> Repository<'a, T> {
         }
     }
 
+    /// Prepends the version number to the role.json filename if using consistent snapshot mode.
+    fn delegated_filename(&self, name: &str) -> String {
+        if self.root.signed.consistent_snapshot {
+            format!("{}.{}.json", self.targets.signed.version, name)
```
iliana: Using the targets role's version is incorrect, it should be the version of the delegated role itself. That can be found in the snapshot role.

Maybe this function needs to return `Option<String>`, then...

```suggestion
            format!("{}.{}.json", self.snapshot.signed.meta.get(format!("{}.json", name))?, name)
```

Might be better to always return `None` if the file isn't present in snapshot.json, though.

tough/src/error.rs
```diff
@@ -357,6 +358,31 @@ pub enum Error {
...
+    DelegateNotFound {
+        name: String,
+        source: crate::schema::Error,
+    },
+
+    #[snafu(display("Delegation doesn't contain targets field"))]
+    NoTargets {},
+
+    #[snafu(display("Targets doesn't contain delegations field"))]
+    NoDelegations {},
```
iliana: Drop the `{}` in these cases if there are no members?

tough/src/lib.rs
```diff
@@ -289,6 +291,11 @@ impl<'a, T: Transport> Repository<'a, T> {
         &self.timestamp
     }
 
+    ///return a vec of all targets including all target files delegated by targets
+    pub fn all_targets(&self) -> Vec<&Target> {
```
iliana: I'm admittedly not thrilled about this API, it's too confusing at first glance compared to `targets()`. I also think maybe it should be an iterator instead of allocating a new `Vec<_>`.

An earlier version of tough didn't even have the `root`, `snapshot`, etc methods.

Don't really know how we ought to do this though! Because a generic `fn role(&self, role: Role) -> impl Role` doesn't let you interact with the contents of the role, you'd have to downcast it or some other cursed thing.

Overall, I think maybe it's time for some API reworking, but that's far, far out of scope for this PR.

tough/src/lib.rs
```diff
@@ -328,14 +335,22 @@ impl<'a, T: Transport> Repository<'a, T> {
...
-        )
+        Ok(if let Ok(target) = self.targets.signed.find_target(name) {
+            let (sha256, file) = self.target_digest_and_filename(target, name);
+            Some(self.fetch_target(target, &sha256, file.as_str())?)
+        } else {
+            None
+        })
+    }
+
+    pub fn role(&self, name: &str) -> Result<&DelegatedRole> {
```
iliana: Would prefer to call this `fn delegated_role`.

tough/src/lib.rs
```diff
@@ -874,10 +889,105 @@ fn load_targets<T: Transport>(
...
+            max_targets_size,
+            delegations,
+            &datastore,
+        )?;
+    }
 
     Ok(targets)
 }
 
+//Follow the paths of delegations starting with the top level targets.json delegation
```
iliana: Style nit throughout: comments markers should have a trailing space.

```suggestion
// Follow the paths of delegations starting with the top level targets.json delegation
```

tough/src/lib.rs
```diff
@@ -874,10 +889,105 @@ fn load_targets<T: Transport>(
 
     // Now that everything seems okay, write the targets file to the datastore.
     datastore.create("targets.json", &targets)?;
```
iliana, line 879: Move these two lines above the `// 4.5` comment, and remove the now-incorrect `does not yet handle delegated roles` comment.

tough/src/lib.rs
```diff
@@ -328,14 +335,22 @@ impl<'a, T: Transport> Repository<'a, T> {
...
+        })
+    }
+
+    pub fn role(&self, name: &str) -> Result<&DelegatedRole> {
+        Ok(self
+            .targets
+            .signed
+            .delegated_role(name)
+            .context(error::DelegateNotFound {
+                name: name.to_string(),
```
iliana: I'd prefer it if this method returned `Option<_>` and we returned None here instead of erroring. This is similar to other "get thing by name" methods like [`HashMap::get`](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.get).

tough/src/lib.rs
```diff
@@ -874,10 +889,105 @@ fn load_targets<T: Transport>(
...
     Ok(targets)
 }
 
+//Follow the paths of delegations starting with the top level targets.json delegation
+fn load_delegations<T: Transport>(
+    transport: &T,
+    snapshot: &Signed<Snapshot>,
+    metadata_base_url: &Url,
+    max_targets_size: u64,
+    delegation: &mut Delegations,
```
iliana, line 903: Marking this to come back and think about later.

tough/src/schema/error.rs
```diff
@@ -18,6 +18,10 @@ pub enum Error {
     #[snafu(display("Duplicate key ID: {}", keyid))]
     DuplicateKeyId { keyid: String },
 
+    /// A duplicate key ID was present in the root metadata.
```
iliana: Docstring seems wrong

tough/src/schema/error.rs
```diff
@@ -93,6 +97,23 @@ pub enum Error {
     /// Unable to create a TUF target from anything but a file
     #[snafu(display("TUF targets must be files, given: '{}'", path.display()))]
     TargetNotAFile { path: PathBuf, backtrace: Backtrace },
+
+    /// Target doesn't have proper permissions from parent delegations
+    #[snafu(display("Invalid file permissions: {}", child))]
```
iliana: At first read, display string seems wrong

tough/src/schema/error.rs
```diff
@@ -93,6 +97,23 @@ pub enum Error {
...
+    UnmatchedPath { child: String },
+
+    /// No valid targets claims target_url
+    #[snafu(display("Target file not delegated: {}", target_url))]
+    TargetNotFound { target_url: String },
```
iliana: At first read, the docstring, display string, and variant name all disagree with each other

tough/src/schema/error.rs
```diff
@@ -93,6 +97,23 @@ pub enum Error {
...
+
+    #[snafu(display("Delegation didn't contain targets field"))]
+    NoTargets {},
+
+    #[snafu(display("Targets didn't contain delegations field"))]
+    NoDelegations {},
+
+    #[snafu(display("Invalid Regex: {}", regex))]
+    BadRegex { regex: String },
```
iliana: `source: regex::Error`?

tough/src/schema/mod.rs
iliana: Clarify -- to verify the key ID, which is a digest of the canonical JSON form of the key representation?

tough/src/schema/mod.rs
```diff
@@ -324,6 +392,214 @@ impl Role for Targets {
...
+        regex_string = regex_string.replace("*", "[^/]*");
+        regex_string = regex_string.replace("?", ".");
+        let re = Regex::new(&regex_string);
+        if let Ok(re) = re {
+            re.is_match(&target)
+        } else {
+            false
+        }
+    }
+}
```
iliana: `wildcardpath` might contain other characters which are special in regexes, we would need to escape those.

Because the spec says `PATHPATTERN can include shell-style wildcards and supports the Unix filename pattern matching convention`, it might be more straightforward to use a higher-level matching library like [globset](https://docs.rs/globset), especially since it efficiently matches over a list of globs. globset's support of Unix-style glob syntax is wider than listed in the TUF spec's examples, but it's impossible to read between the lines of specification vagueness.

tough/src/lib.rs
iliana: (Maybe I should codify this in the readme, but personally one of tough's goals is to make it as simple as possible for a consumer to use a TUF repo *without* having to already know anything about TUF.)

iliana:
GitHub doesn't let me comment on code you didn't change (a sensible choice), but there are a number of comments like this one...

https://github.com/awslabs/tough/blob/7e865747732c6e7f160aad2b1866b531941fc3b6/tough/src/cache.rs#L52

that should probably be found and removed. :)

tough/src/lib.rs
zmrow: Yeah I have to agree on the API concern... but somehow (some way) we need to be able to interact with the roles... Some refactoring is in order but I think it's going to need a decent amount of design

zmrow (approved):
🐙 

I'm good with this barring the few things that @iliana brought up. I'll let her be the final word before we ship this.

tough/src/lib.rs
iliana: Now is later!

I think my initial reaction to this recursive function with a `&mut Delegations` call is that I'm not happy about it, but it's something to throw in an issue, not require a redesign for merging.

As a general rule I would like to avoid recursive calls in tough so that a particularly gnarly TUF repository doesn't result in stack overflow.

iliana (approved):

webern (approved):
👍

tough/src/cache.rs
```diff
@@ -68,6 +67,18 @@ impl<'a, T: Transport> Repository<'a, T> {
             "max_timestamp_size argument",
             &metadata_outdir,
         )?;
+
+        for name in self.targets.signed.role_names() {
+            if let Some(filename) = self.delegated_filename(name) {
+                self.cache_file_from_transport(
+                    filename.as_str(),
+                    self.limits.max_targets_size,
```
webern, line 75: Is this the correct size limit?

```rust
                    self.limits.max_targets_size,
```

iliana: Good catch. This should be the size in `snapshot.json` for the file if it's present, and if not fall back to max_targets_size.

This also brings to my attention that `fn cache_file_from_transport` should probably just determine the appropriate max size (and hash!) if caching from the transport, although this function may as well serialize the in-memory representations of these roles.

The larger scope of my questions can be punted to an issue.

webern: The problem with serializing is that any chang would render a different hash. Fortunately when I was working on caching, serde was kind enough to make a small perturbation so that I saw the peril.

iliana: Oh, right, we put lengths / hashes in the repo metadata. sigh

PR #158, opened 2020-06-26
[editor] Adds the ability to copy or link individual targets

zmrow:
*Issue #, if available:*
N/A

*Description of changes:*
* Refactored `link_targets` to separate the walking from linking
* Added `copy_targets`, which does the same except copies rather than symlinks
* Added `link_target` and `copy_target`, which are used by the above functions and allow for handling single files with custom filenames

*Testing:*
* Unit tests pass
* Built multiple repos with copied and linked targets
* Added `copy_targets` to the integration test

### DIFF
--- a/tough/src/editor/signed.rs
+++ b/tough/src/editor/signed.rs
@@ -12,8 +12,9 @@ use ring::digest::{digest, SHA256, SHA256_OUTPUT_LEN};
 use ring::rand::SecureRandom;
 use serde::Serialize;
 use snafu::{ensure, OptionExt, ResultExt};
+use std::fs;
 use std::os::unix::fs::symlink;
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use walkdir::WalkDir;
 
 /// A signed role, including its serialized form (`buffer`) which is meant to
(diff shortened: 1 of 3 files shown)


tough/src/editor/signed.rs
```diff
@@ -219,22 +209,43 @@ impl SignedRepository {
...
                 .to_str()
-                .context(error::PathUtf8 { path: entry.path() })?;
+                .context(error::PathUtf8 { path: input })?
+        };
 
-            // Use the file name to see if a target exists in the repo
-            // with that name. If so...
-            let repo_target = match repo_targets.get(file_name) {
-                Some(repo_target) => repo_target,
-                None => continue,
```
tjkirch, line 232: I think I missed a change in behavior here.

Before, if the target's filename didn't exist in the repo, we'd `continue` here and not symlink.  With the new code, we of course skip the hash check, but we do proceed to call the given copy or link function.

I think this is a problem because calling `link_targets` / `copy_targets` on a directory shouldn't link/copy targets that weren't in the repo, or we're effectively enlarging the target directory for no reason because the extra files aren't *in* the repo.

zmrow:
Updated per @tjkirch 's comments

tjkirch (approved):
Will require a minor version bump in tough.

webern (approved):

PR #169, opened 2020-07-08
tough and tuftool: setup logging

### LINKED ISSUE #9: Add logging
With logging output, some types of issues would be easier to debug, and some outstanding events can be noted, for example removal of datastore files upon rotation.

webern:
*Issue #, if available:*

Closes #9

*Description of changes:*

Add `log` to `tough` and `simplelogger` to `tuftool`. 
I didn't add a lot of logging because I expect these to be added organically over time. I added a couple that I know will be useful to `http`. Also, from #9: 

> some outstanding events can be noted, for example removal of datastore files upon rotation.

So I added a trace in the spot that I think was being referred.

*Testing*

I used a `tuftool` download command and confirmed that it worked both with and without adding `--log-level trace`. I confirmed that `tuftool` and `tough` traces showed up, but `hyper` and `reqwest` traces did not (I think this is the desired behavior).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1562,6 +1562,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "simplelog"
+version = "0.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "chrono 0.4.13 (registry+https://github.com/rust-lang/crates.io-index)",
+ "log 0.4.8 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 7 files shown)


tjkirch (approved):
> tuftool and tough traces showed up, but hyper and reqwest traces did not (I think this is the desired behavior)

I usually leave everything because it's easy to filter things out but occasionally it's hard to reproduce / rerun to be able to add things back in.  I think the way you have it is probably what's expected, though.  I don't feel strongly.

zmrow (approved):
🥧

webern:
> webern force-pushed the webern:pr/logging branch from 9ed8e5d to 73b3024

Rebase with no conflicts.

PR #174, opened 2020-07-10
Integration of write signing in tough library and tuftool commands to use delegated targets

ecpullen:
**This is the second portion of delegated targets implementation completing the tough library. ~~A third upcoming pr will include updates to tuftool for delegated targets.~~**

*Issue #, if available:*
Related to #5

*Description of changes:*
* Use RepositoryEditor to add delegated targets
* Sign delegated targets separately
* Update snapshot and timestamp metadata
* Load updated role metadata and updated targets
* Use more types of key sources 

*Testing*

* Created repo with delegate targets and delegated roles
* Updated snapshot and timestamp metadata
* Created delegated role as delegatee
* Loaded delegated role as delegator
* Loaded delegator and delegatee role metadata as repo owner
* Signed targets metadata as targets key holder and repo owner
* Updated and added targets to delegated role


tough/src/editor/mod.rs
```diff
@@ -187,16 +225,69 @@ impl RepositoryEditor {
...
+    pub fn add_target(&mut self, name: &str, target: Target) -> Result<&mut Self> {
+        self.targets_struct
+            .as_mut()
+            .ok_or_else(|| error::Error::NoTargets)?
+            .add_target(&name, target);
+
+        Ok(self)
+    }
+
+    /// Remove a `Target` to the repository
```
webern:
```suggestion
    /// Remove a `Target` from the repository
```

tough/src/editor/mod.rs
```diff
@@ -223,30 +349,139 @@ impl RepositoryEditor {
...
-            self.add_target_path(target)?;
+    /// Remove all targets from this repo
+    pub fn clear_targets(&mut self) -> &mut Self {
+        if let Some(targets) = self.targets_struct.as_mut() {
+            targets.clear_targets();
         }
+        self
+    }
+
+    ///Delegate target with name. If `key_source` is given, new keys are given to the role if not parent keys are used
```
webern:
```suggestion
    /// Delegate target with name. If `key_source` is given, new keys are given to the role if not parent keys are used
```

tough/src/editor/mod.rs
```diff
@@ -223,30 +349,139 @@ impl RepositoryEditor {
...
+    pub fn add_delegate(
+        &mut self,
+        from: &str,
+        name: String,
+        key_source: Option<&[Box<dyn KeySource>]>,
```
webern:
```suggestion
        key_sources: Option<&[Box<dyn KeySource>]>,
```

tough/src/editor/mod.rs
```diff
@@ -223,30 +349,139 @@ impl RepositoryEditor {
...
+        // Find the parent targets for the role we are creating
+        let mut parent = if from == "targets" {
+            targets
+        } else {
+            targets
+                .targets_by_name(from)
+                .context(error::DelegateMissing { name: from })?
+        };
+        // Get the keys used to sign the new role
+        let (keyids, key_pairs) = if let Some(key) = key_source {
```
webern:
```suggestion
        let (keyids, key_pairs) = if let Some(keys) = key_sources {
```

tough/src/editor/mod.rs
```diff
@@ -223,30 +349,139 @@ impl RepositoryEditor {
...
+                    .as_sign()
+                    .context(error::KeyPairFromKeySource)?
+                    .tuf_key();
+                let key_id = RepositoryEditor::add_key(&mut parent, key_pair.clone())?;
+                key_pairs.insert(key_id.clone(), key_pair);
+                keyids.push(key_id);
+            }
+            (keyids, key_pairs)
+        } else {
+            // If we weren't given a new key source create a role using parents keys
```
webern:
```suggestion
            // If we weren't given a new key source create a role using parent keys
```

tough/src/editor/mod.rs
```diff
@@ -396,4 +623,127 @@ impl RepositoryEditor {
             _extra: HashMap::new(),
         }
     }
+
+    /// Refreshes the timestamp and snapshot
+    /// Allows for expiration to be moved sooner
```
webern:
```suggestion
    /// Allows for expirations to be changed
```

tough/src/editor/mod.rs
```diff
@@ -396,4 +623,127 @@ impl RepositoryEditor {
...
+        let signed_targets = SignedRole::from_signed(repo.targets)?;
+        let signed_root = SignedRole::from_signed(repo.root)?;
+
+        let mut snapshot = repo.snapshot.signed;
+        // Update snapshot version and expiration
+        snapshot.expires = snapshot_expiration;
+        snapshot.version = NonZeroU64::new(
+            u64::from(snapshot.version)
+                .checked_add(1)
+                .unwrap_or(1 as u64),
```
webern: Should error on overflow.

```rust
                .unwrap_or(1 as u64),
```

tough/src/editor/mod.rs
```diff
@@ -396,4 +623,127 @@ impl RepositoryEditor {
...
+        )
+        .unwrap_or_else(|| NonZeroU64::new(1).unwrap());
```
webern: Should error on overflow.

```rust
        .unwrap_or_else(|| NonZeroU64::new(1).unwrap());
```

tough/src/editor/mod.rs
```diff
@@ -396,4 +623,127 @@ impl RepositoryEditor {
...
+
+        let signed_snapshot = SignedRole::new(snapshot, &signed_root.signed.signed, keys, &rng)?;
+
+        // Update timestamp to reflect changes to snapshot
+        let mut timestamp = repo.timestamp.signed;
+        timestamp.expires = timestamp_expiration;
+        timestamp.version = NonZeroU64::new(
+            u64::from(timestamp.version)
+                .checked_add(1)
+                .unwrap_or(1 as u64),
```
webern: Should error on overflow.

```rust
                .unwrap_or(1 as u64),
```

tough/src/editor/mod.rs
```diff
@@ -396,4 +623,127 @@ impl RepositoryEditor {
...
+        )
+        .unwrap_or_else(|| NonZeroU64::new(1).unwrap());
```
webern: Should error on overflow.

```rust
        .unwrap_or_else(|| NonZeroU64::new(1).unwrap());
```

tough/src/editor/signed.rs
```diff
@@ -106,6 +108,134 @@ where
...
+            buffer,
+            sha256,
+            length,
+        };
+
+        Ok(signed_role)
+    }
+
+    /// creates a map of all signed targets roles excluding the toplevel Targets
+    ///  if `include_all`, throw error if needed keys are not present if not just ignore
```
webern:
```suggestion
    /// if `include_all`, throw error if needed keys are not present if not just ignore
```

tough/src/editor/signed.rs
```diff
@@ -106,6 +108,134 @@ where
...
+        // Serialize the role, and calculate its length and
+        // sha256.
+        let mut buffer = serde_json::to_vec_pretty(&role).context(error::SerializeSignedRole {
+            role: T::TYPE.to_string(),
+        })?;
+        buffer.push(b'\n');
+        let length = buffer.len() as u64;
+
+        let mut sha256 = [0; SHA256_OUTPUT_LEN];
+        sha256.copy_from_slice(digest(&SHA256, &buffer).as_ref());
```
webern: I find this hard to understand. Maybe you could say more about what you are doing. Like, why is serializing the role into a vector and taking its hash the right thing to do? Why does a newline need to be added?

```rust
        // Serialize the role, and calculate its length and
        // sha256.
        let mut buffer = serde_json::to_vec_pretty(&role).context(error::SerializeSignedRole {
            role: T::TYPE.to_string(),
        })?;
        buffer.push(b'\n');
        let length = buffer.len() as u64;

        let mut sha256 = [0; SHA256_OUTPUT_LEN];
        sha256.copy_from_slice(digest(&SHA256, &buffer).as_ref());
```

tough/src/editor/signed.rs
```diff
@@ -106,6 +108,134 @@ where
...
+                role.signatures.push(Signature {
+                    keyid: signing_key_id.clone(),
+                    sig: sig.into(),
+                });
+
+                role
+            } else if include_all {
+                // Make sure the signature is valid targets
+                //only sign targets that we have keys for without throwing an error
+                //delegations allow a key to sign some roles without having to sign them all
```
webern: See if this can be clearer/cleaner

```rust
                // Make sure the signature is valid targets
                //only sign targets that we have keys for without throwing an error
                //delegations allow a key to sign some roles without having to sign them all
```

tough/src/editor/signed.rs
```diff
@@ -106,6 +108,134 @@ where
...
+            // sha256.
+            let mut buffer =
+                serde_json::to_vec_pretty(&role).context(error::SerializeSignedRole {
+                    role: T::TYPE.to_string(),
+                })?;
+            buffer.push(b'\n');
+            let length = buffer.len() as u64;
+
+            let mut sha256 = [0; SHA256_OUTPUT_LEN];
+            sha256.copy_from_slice(digest(&SHA256, &buffer).as_ref());
```
webern: If this it a 'thing' then maybe a function for it (repeated code).

```rust
            let mut buffer =
                serde_json::to_vec_pretty(&role).context(error::SerializeSignedRole {
                    role: T::TYPE.to_string(),
                })?;
            buffer.push(b'\n');
            let length = buffer.len() as u64;

            let mut sha256 = [0; SHA256_OUTPUT_LEN];
            sha256.copy_from_slice(digest(&SHA256, &buffer).as_ref());
```

tuftool/src/update_targets.rs
```diff
@@ -0,0 +1,155 @@
...
+    #[structopt(short = "r", long = "root")]
+    root: PathBuf,
+
+    /// TUF repository metadata base URL
+    #[structopt(short = "m", long = "metadata-url")]
+    metadata_base_url: Url,
+
+    /// Directory of targets
+    #[structopt(short = "t", long = "add-targets")]
+    targets_indir: PathBuf,
```
webern: Inconsistent argument name? Not clear.
```rust
    /// Directory of targets
    #[structopt(short = "t", long = "add-targets")]
    targets_indir: PathBuf,
```

tuftool/src/update_targets.rs
```diff
@@ -0,0 +1,155 @@
...
+
+    /// The directory where the repository will be written
+    #[structopt(short = "o", long = "outdir")]
+    outdir: PathBuf,
+
+    /// Follow symbolic links in the given directory when adding targets
+    #[structopt(short = "f", long = "follow")]
+    follow: bool,
+
+    /// Determins if entire repo should be signed
```
webern:
```suggestion
    /// Determines if entire repo should be signed
```

tuftool/src/update_targets.rs
```diff
@@ -0,0 +1,155 @@
...
+        let datastore = tempdir().context(error::TempDir)?;
+        // We don't do anything with targets so we will use metadata url
+        let settings = tough::Settings {
+            root: File::open(&self.root).unwrap(),
+            datastore: &datastore.path(),
+            metadata_base_url: self.metadata_base_url.as_str(),
+            targets_base_url: self.metadata_base_url.as_str(),
+            limits: Limits::default(),
+            expiration_enforcement: ExpirationEnforcement::Safe,
+        };
```
webern: Messy diff, just suggesting the comment be moved to the site of the thing it's explaining.

```suggestion
        let settings = tough::Settings {
            root: File::open(&self.root).unwrap(),
            datastore: &datastore.path(),
            metadata_base_url: self.metadata_base_url.as_str(),
            // We don't do anything with targets so we will use metadata url
            targets_base_url: self.metadata_base_url.as_str(),
            limits: Limits::default(),
            expiration_enforcement: ExpirationEnforcement::Safe,
        };
```

tuftool/src/update_targets.rs
```diff
@@ -0,0 +1,155 @@
...
+            let signed_repo = editor.sign(&self.keys).context(error::SignRepo)?;
+            let metadata_dir = &self.outdir.join("metadata");
+            signed_repo.write(metadata_dir).context(error::WriteRepo {
+                directory: metadata_dir,
+            })?;
+
+            return Ok(());
+        }
+        // if not, write updated role to outdir/metadata
+        // sign the updated role and recieve SignedRole for the new role
```
webern:
```suggestion
        // sign the updated role and receive SignedRole for the new role
```

tuftool/src/update_targets.rs
```diff
@@ -0,0 +1,155 @@
...
+
+        for (filename, target) in new_targets {
+            editor
+                .add_target_to_delegatee(&filename, target, &self.role)
+                .context(error::DelegateeNotFound {
+                    role: self.role.clone(),
+                })?;
+        }
+
+        // if signall is included sign and write entire repo
```
webern:
```suggestion
        // if sign_all is requested, sign and write entire repo
```

tuftool/tests/delegation_commands.rs
```diff
@@ -0,0 +1,661 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+mod utl;
```
webern: I think I renamed `utl` to `test_utils` to match `tough`, you might need to rebase. Also, can you look at the tuftool `download_command.rs` test. It doesn't download `file3.txt` but I think it should (`file3.txt` is the delegated target in the reference repo)... (this can be separate from the current PR...)

```rust
mod utl;
```

tuftool/src/main.rs
```diff
@@ -71,9 +74,15 @@ enum Command {
...
-    Update(update::UpdateArgs),
+    Update(Box<update::UpdateArgs>),
     /// Manipulate a root.json metadata file
     Root(root::Command),
+    //Create delegated role
+    CreateRole(Box<create_role::CreateRoleArgs>),
+    //Add delegated role
+    AddRole(Box<add_role::AddRoleArgs>),
+    //Update Delegated targets
+    UpdateDelegatedTargets(Box<update_targets::UpdateTargetsArgs>),
```
webern: These should be doc comments `/// blah blah`, not `//blah blah`.
```rust
    //Create delegated role
    CreateRole(Box<create_role::CreateRoleArgs>),
    //Add delegated role
    AddRole(Box<add_role::AddRoleArgs>),
    //Update Delegated targets
    UpdateDelegatedTargets(Box<update_targets::UpdateTargetsArgs>),
```

tuftool/src/error.rs
```diff
@@ -163,8 +176,11 @@ pub(crate) enum Error {
     },
 
     #[snafu(display("Unable to initialize logger: {}", source))]
-    Logger {
-        source: simplelog::TermLogError,
+    Logger { source: simplelog::TermLogError },
+
+    #[snafu(display("Unable to load incoming metadata"))]
```
webern: Display the source error for this and all errors that have a `source`.

```suggestion
    #[snafu(display("Unable to load incoming metadata: {}", source))]
```

tuftool/src/add_role.rs
```diff
@@ -0,0 +1,187 @@
...
+
+    /// Path to root.json file for the repository
+    #[structopt(short = "r", long = "root")]
+    root: PathBuf,
+
+    /// TUF repository metadata base URL
+    #[structopt(short = "m", long = "metadata-url")]
+    metadata_base_url: Url,
+
+    /// incoming metadata
```
webern: Match casing of other arguments (these are displayed to the user during --help).

```suggestion
    /// Incoming metadata
```

tough/src/error.rs
```diff
@@ -420,7 +429,16 @@ pub enum Error {
     TargetNotFound { target_url: String },
 
     #[snafu(display("Delegated role not found: {}", name))]
-    DelegateNotFound {
+    DelegateNotFound { name: String },
+
+    #[snafu(display("Targets role not found: {}", name))]
```
webern: Errors that have an underlying `source` should display it (unless it's just going to say the exact same thing I guess).

```suggestion
    #[snafu(display("Targets role '{}' not found: {}", name, source))]
```

tough/tests/repo_editor.rs
```diff
@@ -37,6 +41,34 @@ impl RepoPaths {
...
+fn targets_key_path1() -> PathBuf {
+    test_data().join("targetskey-1")
+}
+
+// Path to fake targets in the reference implementation
+fn targets_path() -> PathBuf {
+    test_data().join("tuf-reference-impl").join("targets")
+}
+
+fn read_to_end<R: Read>(mut reader: R) -> Vec<u8> {
```
webern: I think this one is available in test_utils

```rust
fn read_to_end<R: Read>(mut reader: R) -> Vec<u8> {
```

tough/tests/repo_editor.rs
```diff
@@ -63,9 +121,510 @@ fn repository_editor_from_repository() {
     assert!(RepositoryEditor::from_repo(&root, repo).is_ok());
 }
 
+//Create sign write and reload repo
```
webern: All existing comments have a space after the slashes. New comments should as well.

```suggestion
// Create sign write and reload repo
```

tough/tests/repo_editor.rs
```diff
@@ -63,9 +121,510 @@ fn repository_editor_from_repository() {
...
+#[test]
+fn cre_sig_wri_rel() {
```
webern: An overly long function name is probably better than a cryptic one.
```rust
fn cre_sig_wri_rel() {
```

webern (commented):
Nice work! I think the original authors, Zack, Tom and iliana, will need to approve from a design and correctness standpoint (LGTM). As such, most of my comments are nits.

tough/src/editor/mod.rs
ecpullen: Since the rest of the library uses `key_source` I think it makes sense to keep the naming consistent.

tough/src/editor/mod.rs
ecpullen: This was also designed to be consistent with other parts of the library, but I'm happy to change it if that's what is best.

tuftool/src/update_targets.rs
ecpullen: Another tuftool command uses this same naming scheme, so I kept it to keep consistency.

tough/src/http.rs
```diff
@@ -172,7 +172,7 @@ impl RetryState {
                 Ordering::Greater => {
                     self.wait = settings.max_backoff;
                 }
-                _ => {}
+                std::cmp::Ordering::Equal => {}
```
ecpullen: This has nothing to do with the content in this pr, but had to be changed to pass clippy check.

webern: Yeah, clippy updates from time-to-time adding new lints. No problem.

tough/src/editor/mod.rs
webern: OK, no problem.

tough/src/editor/mod.rs
webern: OK, no problem.

tuftool/src/update_targets.rs
webern: OK.

tough/src/editor/signed.rs
webern: I'm trying to understand the diff. This went away in favor of `SignedRole::from_signed` which already existed? Is that a correct reading of your change?

webern (approved):
LGTM

tough/src/editor/signed.rs
ecpullen: Yes that's exactly what changed.

tough/src/editor/keys.rs
```diff
@@ -30,3 +30,23 @@ pub(crate) fn get_root_keys(root: &Root, keys: &[Box<dyn KeySource>]) -> Result<
     ensure!(!root_keys.is_empty(), error::KeysNotFoundInRoot);
     Ok(root_keys)
 }
+
+/// Gets the corresponding keys from delegations for the given `KeySource`s.
+/// This is a convenience function that wraps `Delegations.key_id()` for multiple
+/// `KeySource`s.
+pub(crate) fn get_targets_keys(
```
zmrow, line 100: This returns a `RootKeys`, but that's not really accurate given what this is returning. 

For future readers it may make sense to create another type that's a bit more descriptive.  Perhaps `DelegationsKeys` or something?

tough/src/editor/mod.rs
```diff
@@ -187,16 +225,69 @@ impl RepositoryEditor {
...
+    }
+
+    /// Add a `Target` to the delegatee `role`
+    pub fn add_target_to_delegatee(
+        &mut self,
+        name: &str,
+        target: Target,
+        role: &str,
+    ) -> Result<&mut Self> {
+        if role == "targets" {
```
zmrow: The method name here seems to indicate that its purpose is adding targets to *delegatee*s, not the `Targets` role, which is decidedly *not* a delegatee.  

Let's make this method specific to delegatee's only; this will also avoid the magic string matching. :)

tough/src/editor/mod.rs
```diff
@@ -173,6 +209,8 @@ impl RepositoryEditor {
                 supported: SPEC_VERSION
             }
         );
+        self.snapshot_version(snapshot.version);
+        self.snapshot_expires(snapshot.expires);
```
zmrow: Why are we holding on to the version and expiration? I think we want to encourage the behavior that these should be updated each time you change a repository.

tough/src/editor/mod.rs
```diff
@@ -187,16 +225,69 @@ impl RepositoryEditor {
                 supported: SPEC_VERSION
             }
         );
+        self.timestamp_expires(timestamp.expires());
+        self.timestamp_version(timestamp.version);
```
zmrow: Same comment here for the version/expiration: why are we holding on to them?

tough/src/editor/mod.rs
```diff
@@ -223,30 +349,139 @@ impl RepositoryEditor {
...
+            targets.clear_targets();
         }
+        self
+    }
+
+    /// Delegate target with name. If `key_source` is given, new keys are given to the role if not parent keys are used
+    pub fn add_delegate(
+        &mut self,
+        from: &str,
+        name: String,
```
zmrow: It feels  a little odd to have multiple string arguments with different types. Could we use something like `AsRef<str>` here for both?

tough/src/editor/mod.rs
```diff
@@ -187,16 +225,69 @@ impl RepositoryEditor {
...
+                name: name.to_string(),
+            })?
+            .add_target(&name, target);
+
+        Ok(self)
+    }
+
+    /// Remove a `Target` from the delegatee `role`
+    pub fn remove_target_from_delegatee(&mut self, name: &str, role: &str) -> Result<&mut Self> {
+        if role == "targets" {
```
zmrow: Same comment here as above: let's remove any references to the `Targets` role, which is *not* a delegatee.

tough/src/editor/mod.rs
```diff
@@ -206,6 +297,41 @@ impl RepositoryEditor {
...
+    {
+        let (target_name, target) = RepositoryEditor::build_target(target_path)?;
+        self.add_target(&target_name, target)?;
+        Ok(self)
+    }
+
+    /// Add a list of target paths to the repository
+    ///
+    /// See the note on `add_target_path()` regarding performance.
+    pub fn add_target_paths<P>(&mut self, targets: Vec<P>, role: &str) -> Result<&mut Self>
```
zmrow: Perhaps this is related to my comments on the "delegatee" methods.  I'm confused by this method's signature.  Do I need to supply the string "`targets`" to add a target to the `Targets` role?

I think we should keep `delegatee` and `Targets` role methods separate so there is a clear interface boundary and no magic strings are being passed around.  Users can then be totally clear on what method to use. 

The delegatee and targets methods *may* be able to do some of the same things under the hood and we can abstract those.

tough/src/editor/mod.rs
```diff
@@ -223,30 +349,139 @@ impl RepositoryEditor {
...
-    /// See the note on `add_target_path()` regarding performance.
-    pub fn add_target_paths<P>(&mut self, targets: Vec<P>) -> Result<&mut Self>
-    where
-        P: AsRef<Path>,
-    {
-        for target in targets {
-            self.add_target_path(target)?;
+    /// Remove all targets from this repo
+    pub fn clear_targets(&mut self) -> &mut Self {
+        if let Some(targets) = self.targets_struct.as_mut() {
```
zmrow: Would we ever have a situation where we didn't have a `Targets` struct? Don't we create one in the `::new()` method?

tough/src/editor/mod.rs
```diff
@@ -223,30 +349,139 @@ impl RepositoryEditor {
...
+        &mut self,
+        from: &str,
+        name: String,
+        key_source: Option<&[Box<dyn KeySource>]>,
+        paths: PathSet,
+        expiration: DateTime<Utc>,
+        version: NonZeroU64,
+    ) -> Result<&mut Self> {
+        // If we are delegating from targets all paths are valid
+        if from != "targets" {
```
zmrow: I'd rather not use string matching here again.

Is there a sane default here? Does it make sense to have `from` be an `Option<String>`, and if it's `None` we'll make `Targets` the delegated role? At least then we could document that behavior.

tough/src/editor/mod.rs
```diff
@@ -223,30 +349,139 @@ impl RepositoryEditor {
...
+            .delegations
+            .as_mut()
+            .ok_or_else(|| error::Error::NoDelegations)?;
+        // Create new targets for `role`
+        let mut new_targets = Targets::new(SPEC_VERSION.to_string(), version, expiration);
+        let mut new_delegations = Delegations::new();
+        new_delegations.keys.extend(key_pairs);
+        // Create a new delegations for `role`
+        new_targets.delegations = Some(new_delegations);
+        let threshold = NonZeroU64::new(keyids.len().try_into().context(error::InvalidInto {})?)
```
zmrow: Do we want to make threshold a configurable value rather than inferring it?

tough/src/editor/mod.rs
```diff
@@ -223,30 +349,139 @@ impl RepositoryEditor {
...
 
-    /// Remove all targets from this repo
-    pub fn clear_targets(&mut self) -> &mut Self {
-        self.existing_targets
-            .get_or_insert_with(HashMap::new)
-            .clear();
-        self.new_targets.get_or_insert_with(HashMap::new).clear();
-        self
+    /// Adds a key to the targets delegation role if not already present, and returns a result with the key id.
+    fn add_key(targets: &mut Targets, key: Key) -> Result<Decoded<Hex>> {
```
zmrow: It seems like this method is specific to adding a key for delegations? If so, let's name it as such.

Also, do keys need to be added to a specific delegated role? Is there a missing parameter to specify the correct role?

tough/src/editor/mod.rs
```diff
@@ -289,38 +530,19 @@ impl RepositoryEditor {
...
-        let _extra = self.targets_extra.clone().unwrap_or_else(HashMap::new);
-        Ok(Targets {
-            spec_version: SPEC_VERSION.to_string(),
-            version,
-            expires,
-            targets,
-            _extra,
-            delegations: None,
-        })
+        Ok(self
```
zmrow: I'm really confused by this method as it's not really building the `Targets` struct. Is it necessary anymore?

tough/src/editor/mod.rs
```diff
@@ -394,4 +621,127 @@ impl RepositoryEditor {
             _extra: HashMap::new(),
         }
     }
+
+    /// Refreshes the timestamp and snapshot
+    /// Allows for expirations to be changed
+    pub fn update_snapshot<T>(
```
zmrow: Perhaps I haven't read far enough yet, but why would we need this? The reason we discard the expiration and version when reading a repository is so that both get updated. It seems kinda odd to offer a method to update snapshot and timestamp together as we offer "setter" methods for both.

tough/src/editor/mod.rs
```diff
@@ -394,4 +621,127 @@ impl RepositoryEditor {
...
+
+    /// Crawls a given directory and symlinks any targets found to the given
+    /// "out" directory. If consistent snapshots are used, the target files
+    /// are prefixed with their `sha256`.
+    ///
+    /// For each file found in the `indir`, the method gets the filename and
+    /// if the filename exists in `Targets`, the file's sha256 is compared
+    /// against the data in `Targets`. If this data does not match, the
+    /// method will fail.
+    pub fn link_targets<P1, P2>(
```
zmrow: Did these get moved over from `SignedRepository`?  How come?  If they're needed here, let's make sure it's clear why you would use these rather than the methods from `SignedRepository`.

zmrow (commented):
Publishing what I have so far - more to come.

tough/src/editor/keys.rs
iliana: Or we rename `RootKeys` to `KeyList`?

tough/src/editor/mod.rs
iliana: `authorize_target`, `permit_target`, ...?

tough/src/editor/mod.rs
iliana: If it's an `Option<Targets>`, it should be handled like this :)

tuftool/src/error.rs
```diff
@@ -163,8 +176,11 @@ pub(crate) enum Error {
...
     #[snafu(display("Unable to initialize logger: {}", source))]
-    Logger {
-        source: simplelog::TermLogError,
+    Logger { source: simplelog::TermLogError },
+
+    #[snafu(display("Unable to load incoming metadata: {}", source))]
+    LoadMetadata {
+        source: tough::error::Error,
         backtrace: Backtrace,
     },
```
iliana: `Logger` lost its backtrace?

iliana (approved):
I'm happy with this but am deferring to @zmrow's comments.

ecpullen:
This PR is being refactored and moved to #200.

PR #177, opened 2020-07-13
tuftool: truncate targets when downloading

### LINKED ISSUE #173: `tuftool download` does not truncate existing files
If you `tuftool download` from one repo into a directory, then `tuftool download` another repo into the same directory, and those repos share a target filename, then the second file will be written over the first *without* the truncate flag.

This means you can see the full contents of the file from the second repo, then the remainder of the contents of the file from the first repo, starting at the byte number matching the length of the file from the second repo.  I saw exactly this when downloading two Bottlerocket repos, since "manifest.json" is common to each.

webern:
*Issue #, if available:*

Fixes #173 

*Description of changes:*

Use `File::create` instead of `OpenOptions` in `tuftool download`. `File::create` includes `create(true).open(true).truncate(true)` which is how we want to open our files for writing. By including `truncate(true)` the buggy behaviour seen in #173 is fixed.

*Testing*

I added an integ test of the `download` command. It adds some text to the end of one of the targets in the `outdir` after download. It then downloads the repo again and asserts the the file is correct.

This test was observed to fail before implementing the fix:

```
thread 'download_command_truncates' panicked at 'assertion failed: `(left == right)`
  left: `"This is an example target file.bloop
"`,
 right: `"This is an example target file."`: file1.txt contents do not match.', tuftool/tests/download_command.rs:34:5
```

After implementing the fix, the test passes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1931,6 +1931,7 @@ dependencies = [
  "hex 0.4.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "log 0.4.8 (registry+https://github.com/rust-lang/crates.io-index)",
  "maplit 1.0.2 (registry+https://github.com/rust-lang/crates.io-index)",
+ "mockito 0.26.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "olpc-cjson 0.1.0",
  "pem 0.8.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "rayon 1.3.1 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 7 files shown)


webern:
> webern force-pushed the webern:pr/download-bug branch from c8a1c0c to 73dab47

Rebase with no conflicts.

zmrow (approved):
🤗

tjkirch (approved):
Thanks!

webern:
> webern force-pushed the webern:pr/download-bug branch from 73dab47 to b44413c

Rebase without conflicts.

PR #184, opened 2020-07-16
Create an error if the proposed root.json is unstable

### LINKED ISSUE #123: tuftool: present an error when creating an un-loadable tuf repo
This has similar intent to #119, which is that we could make it harder to use `tuftool` incorrectly.

It is possible to use `tuftool create` with a `root.json` that can never be satisfied. For example, if I have zero keys for the `targets` role while specifying a threshold for the same role greater than zero.

Maybe `tuftool create` should inspect the sanity of `root.json` before proceeding, and error out if `root.json` would make the created repo unusable.

ecpullen:
*Issue #, if available:*
Closes #123 
*Description of changes:*
The pr adds a sanity check to `RepositoryEditor::new()` to make sure the number of keyids is more than the threshold for each role.
*Testing*
Created a root.json with a high threshold than keyids for targets and calling tuftool create reported that root was unstable
All other tests work as expected

### DIFF
--- a/tough/src/editor/mod.rs
+++ b/tough/src/editor/mod.rs
@@ -78,7 +78,11 @@ impl RepositoryEditor {
         // Sanity check of root
         for (roletype, rolekeys) in &root.signed.roles {
             if rolekeys.threshold.get() > rolekeys.keyids.len() as u64 {
-                return Err(error::Error::UnstableRoot { role: *roletype });
+                return Err(error::Error::UnstableRoot {
+                    role: *roletype,
+                    threshold: rolekeys.threshold.get(),
+                    actual: rolekeys.keyids.len(),
+                });
             }
         }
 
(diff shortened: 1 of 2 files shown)


webern (approved):
LGTM.

tough/src/error.rs
```diff
@@ -440,6 +440,10 @@ pub enum Error {
 
     #[snafu(display("Role missing from snapshot meta: {}", name))]
     RoleNotInMeta { name: String },
+
+    /// Root creates an unloadable repo
+    #[snafu(display("Unstable root, less keys then threshold for: {}", role))]
```
zmrow: Since we know how many keys we have (and the threshold), let's include it in the error message.  Suggestion:

`"Unstable root; found {} keys for role {}, threshold is {}"`

zmrow (approved):
💴

tough/src/editor/mod.rs
```diff
@@ -74,6 +74,18 @@ impl RepositoryEditor {
         let root_buf_len = root_buf.len() as u64;
         let root = serde_json::from_slice::<Signed<Root>>(&root_buf)
             .context(error::FileParseJson { path: root_path })?;
+
+        // Sanity check of root
```
iliana: https://gist.github.com/seanmhanson/fe370c2d8bd2b3228680e38899baf5cc

```suggestion
        // Quick check that root is signed by enough key IDs
```

iliana (approved):

PR #186, opened 2020-07-16
Verbatim

### LINKED ISSUE #20: Figure out something better than an `_extra` struct member

ecpullen:
**I will rebase this so that it doesn't get clippied once #185 is merged**

*Issue #, if available:*
Closes #20 
*Description of changes:*
This pr adds a verbatim type to eliminate the `_extra` field found in most metadata structs.

*Testing*
All tests perform as expected.

### DIFF
--- a/tough/src/cache.rs
+++ b/tough/src/cache.rs
@@ -42,7 +42,7 @@ impl<'a, T: Transport> Repository<'a, T> {
                 self.cache_target(&targets_outdir, target_name.as_ref())?;
             }
         } else {
-            let targets = &self.targets.signed.as_ref().targets_map();
+            let targets = (*self.targets.signed).targets_map();
             for target_name in targets.keys() {
                 self.cache_target(&targets_outdir, target_name)?;
             }
@@ -68,7 +68,7 @@ impl<'a, T: Transport> Repository<'a, T> {
             &metadata_outdir,
         )?;
 
-        for name in self.targets.signed.as_ref().role_names() {
+        for name in (*self.targets.signed).role_names() {
             if let Some(filename) = self.delegated_filename(name) {
                 self.cache_file_from_transport(
                     filename.as_str(),
@@ -81,7 +81,7 @@ impl<'a, T: Transport> Repository<'a, T> {
 
         if cache_root_chain {
             // Copy all versions of root.json less than or equal to the current version.
-            for ver in (1..=self.root.signed.as_ref().version.get()).rev() {
+            for ver in (1..=(*self.root.signed).version.get()).rev() {
                 let root_json_filename = format!("{}.root.json", ver);
                 self.cache_file_from_transport(
                     root_json_filename.as_str(),
@@ -96,8 +96,8 @@ impl<'a, T: Transport> Repository<'a, T> {
 
     /// Prepends the version number to the snapshot.json filename if using consistent snapshot mode.
     fn snapshot_filename(&self) -> String {
-        if self.root.signed.as_ref().consistent_snapshot {
-            format!("{}.snapshot.json", self.snapshot.signed.as_ref().version)
+        if (*self.root.signed).consistent_snapshot {
+            format!("{}.snapshot.json", (*self.snapshot.signed).version)
         } else {
             "snapshot.json".to_owned()
         }
@@ -105,8 +105,8 @@ impl<'a, T: Transport> Repository<'a, T> {
 
     /// Prepends the version number to the targets.json filename if using consistent snapshot mode.
     fn targets_filename(&self) -> String {
-        if self.root.signed.as_ref().consistent_snapshot {
-            format!("{}.targets.json", self.targets.signed.as_ref().version)
+        if (*self.root.signed).consistent_snapshot {
+            format!("{}.targets.json", (*self.targets.signed).version)
         } else {
             "targets.json".to_owned()
         }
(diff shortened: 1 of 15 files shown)


tough/src/editor/mod.rs
```diff
@@ -309,18 +310,20 @@ impl RepositoryEditor {
         }
 
         let _extra = self.targets_extra.clone().unwrap_or_else(HashMap::new);
-        Ok(Targets {
+        Ok(From::from(Targets {
```
iliana: `.into()` is easier to read instead of using the trait.

tough/src/cache.rs
```diff
@@ -42,7 +42,7 @@ impl<'a, T: Transport> Repository<'a, T> {
                 self.cache_target(&targets_outdir, target_name.as_ref())?;
             }
         } else {
-            let targets = &self.targets.signed.targets_map();
+            let targets = &self.targets.signed.as_ref().targets_map();
```
iliana: I'm tempted to suggest `impl<T> Deref for Verbatim<T>` (and `DerefMut`) so that we don't need to `as_ref()`/`as_mut()` everywhere, including in downstream crates. However, I worry about how that works if the inner type has a `inner` or `extra` member or method.

ecpullen:
The newest commit implements Deref as suggested by @iliana.

zmrow (commented):
Nice work implementing this @ecpullen !

While I do think this is a *very* clever way to get rid of the `_extra` struct member, now that I see the effects of the `Verbatim` type, I'm not sure it's the right direction for consumers of the library and readability of code.

`<IMHO>`
This is *very* IMHO, but in my usage of the library we're most often accessing struct members that aren't the `_extra` member. Before this change we could access them like `repo.targets().signed.targets`. After this change we have to do things like `(*(*repo.targets().signed).targets...)` or `(*root.signed)` everywhere which is a tough pill to swallow.

I do agree it's slightly inconvenient to specify `_extra: HashMap::new()` or `..Default::default()` when *creating* the structs, but we're *accessing* them far more often. I'll happily write `_extra: HashMap::new()` in one place vs. dereferencing things every time I want to use them.
`</IMHO>`

@iliana , @webern, @tjkirch  what are you thoughts?

ecpullen:
The other option is the version back one commit. That uses as_ref() and as_mut() which makes the code easier for me to read.

tjkirch:
> The other option is the version back one commit. That uses as_ref() and as_mut() which makes the code easier for me to read.

I agree with Zac, and I think requiring `as_ref()` everywhere is just as bad.

Can we instead remove the `_extra` field and cause unknown (invalid!) fields in a role to fail deserialization?  (`#[serde(deny_unknown_fields)]`)  I don't understand the use case for it, and if we're implementing the TUF spec then we should know all of the possible fields in a role, right?

iliana:
> Can we instead remove the `_extra` field and cause unknown (invalid!) fields in a role to fail deserialization?

I think it's much more correct behavior to ignore extra fields but still be able to verify the signature. In fact, if we drop the _extra field today, we will fail to verify repos used as test data in the TUF reference implementation:

https://github.com/awslabs/tough/blob/e3be8115198c1c8f8ec443bd59c88d4ed6546a2a/tough/tests/data/tuf-reference-impl/metadata/root.json#L12-L23

`keyid_hash_algorithms` is defined in [TAP 12](https://github.com/theupdateframework/taps/blob/master/tap12.md), implemented by default in the reference implementation.

> https://github.com/awslabs/tough/pull/186#pullrequestreview-450746264

Need to read the Deref version of the diff and see what you're talking about, one moment :)

tjkirch:
> I think it's much more correct behavior to ignore extra fields but still be able to verify the signature.

If we remove `_extra` and *don't* add `deny_unknown_fields`, then it should ignore extra fields, we just wouldn't have access to them afterward... but we don't need access, right?  Would that work?

iliana:
Well we need to know what the fields are to calculate the same signature the signer did :)

tjkirch:
> Well we need to know what the fields are to calculate the same signature the signer did :)

The reference implementation has tests using fields that aren't in the spec?  :(

I vote for leaving `_extra` - it's not worth hurting the user experience to the degree done by that the options we've discussed so far.

tough/src/cache.rs
```diff
@@ -42,7 +42,7 @@ impl<'a, T: Transport> Repository<'a, T> {
                 self.cache_target(&targets_outdir, target_name.as_ref())?;
             }
         } else {
-            let targets = &self.targets.signed.targets_map();
+            let targets = (*self.targets.signed).targets_map();
```
iliana, line 45: Okay yeah this is weird, I need to play with this. I don't think I expected Deref coercion to not work...

webern: In your mind it would be transparent to all the call sites? Is that the issue?

webern:
> I vote for leaving `_extra` - it's not worth hurting the user experience to the degree done by that the options we've discussed so far.

I tend to agree, though the Rust magicians may yet prevail at making this transparent.

tough/src/cache.rs
ecpullen: Deref coercion only works on methods, not fields. This means we could create a `signed` method of `Signed` to dereference, or we could use getters for the fields.

iliana:
> Deref coercion only works on methods, not fields.

This is what I did not know that made me think this might work. :(

I'm going to agree with @tjkirch that `_extra` is the lesser of all three evils, and if folks generally agree we should close this.

I'd like to keep #20 open in case we can come up with a better way prior to 1.0.

PR #187, opened 2020-07-17
Allow control of link/copy behavior for existing paths

### LINKED ISSUE #180: SignedRepository: link/copy targets without consistent snapshot can overwrite
The `target_path` function in `SignedRepository` handles consistency checks when the target is known to the repo -- if the hash doesn't match, we refuse to continue.  However, if you're using `link/copy_targets` to write targets for multiple repos into a single directory, and they have a target with the same name but different contents, *and* you're not using consistent snapshots, then the later-written file will overwrite the earlier one.

Ideally everyone uses consistent snapshots, or if they don't then they know not to share target names, but this is still a sharp corner if you do meet those conditions.  We have the information we'd need to cause an error in this case.  If the target path exists, checksum it - on match, we don't need to do anything; on failure, error out.  (Note: the `ring` code we use to sha256 the role probably shouldn't be used for this; it would have to read the whole target into memory, and they could be any size.)

zmrow:
*Issue #, if available:*
Fixes #180 

*Description of changes:*
If you used SignedRepository to link/copy targets from multiple repos into a
shared directory, these problems could occur:
* Without consistent snapshot, targets with the same name but different
  contents from different repos can overwrite each other.
* Targets with the same name and same contents from different repos can
  fail to copy/link if the file does not have write permissions.

This change addresses these related failure cases by:
* If not using consistent snapshots, we confirm that the checksum of any
  existing file matches the expected target checksum.
* Adds a `PathExists` enum that's used as a parameter to the copy/link
  functions that can be used to specify what to do with an existing
  path - skip the target, replace the existing file, or fail.

Due to the change in signature of public functions, this is a breaking
change.

*Testing done:*
As part of [this Bottlerocket PR](https://github.com/bottlerocket-os/bottlerocket/pull/964), we've tested building multiple repositories.  Without this change, we had file conflicts that caused failed builds; with this change, we're able to specify the behavior we want for shared target names and build successfully. 


zmrow:
This force-push fixes `tuftool` (which we forgot to update), and renames the `PathExists::DoNothing` variant to `PathExists::Skip` which we thought was clearer (and shorter!).

tuftool/src/create.rs
```diff
@@ -51,6 +52,12 @@ pub(crate) struct CreateArgs {
     #[structopt(short = "t", long = "add-targets")]
     targets_indir: PathBuf,
 
+    /// Behavior when a target exists with the same name and hash in the desired repository
+    /// directory, for example from another repository when you're sharing target directories.
+    /// Options are "replace", "fail", and "skip"
```
webern: I find it a little bit hard to understand. If "desired repository directory" means `outdir` then I think `outdir` might be easier to understand (but I'm not sure that's what it means).

Maybe it's not necessary to give an example of when one would use this. It might just add cognitive load to understanding the argument. If it's clear what the argument does then the user can just ignore it if they don't think they need it.

```rust
    /// Behavior when a target exists with the same name and hash in the desired repository
    /// directory, for example from another repository when you're sharing target directories.
    /// Options are "replace", "fail", and "skip"
```

webern (approved):
LGTM, maybe the description of the argument could be improved.

tuftool/src/create.rs
tjkirch: Yeah, this could be clarified.  I thought it was important to give an example because it's not immediately apparent to everyone how you could have matching target filenames/hashes, and maybe not everyone is sharing a targets directory across repos.  (That's the only case we could think of.)  How about:

```
     /// Behavior when a target exists with the same name and hash in the targets
     /// directory, for example from another repository when they share a targets directory.
    /// Options are "replace", "fail", and "skip".
```

tjkirch:
^ This push addresses @webern's comment by trying to clarify a docstring.

etungsten (approved):

tough/src/editor/signed.rs
```diff
@@ -262,22 +297,20 @@ impl SignedRepository {
...
+        // target we just created. If they are the same, this must
+        // be the same file, symlink it.
+        ensure!(
+            target_from_path.hashes.sha256 == repo_target.hashes.sha256,
+            error::HashMismatch {
+                context: "target",
+                calculated: hex::encode(target_from_path.hashes.sha256),
+                expected: hex::encode(&repo_target.hashes.sha256),
+            }
+        );
```
bcressey: nit: it's a bit odd to find "symlink it" in a comment here when the symlink happens much later

bcressey (approved):

tjkirch:
^ This push fixes a comment that we forgot to update; thanks for the catch @bcressey.

PR #189, opened 2020-07-20
Allow multiple signatures in editor

ecpullen:
*Issue #, if available:*
N/A

*Description of changes:*
Creating a `SignedRole` from a `RepositoryEditor` doesn't allow for multiple keys to sign the role, so if a role had a threshold of 2 signatures, it could never be properly signed.
Collects all valid signing keys instead of the first one and signs the role using all of them

### DIFF
--- a/tough/src/editor/signed.rs
+++ b/tough/src/editor/signed.rs
@@ -54,12 +54,9 @@ where
         // Ensure the keys we have available to us will allow us
         // to sign this role. The role's key ids must match up with one of
         // the keys provided.
-        let (signing_key_id, signing_key) = root_keys
+        let valid_keys = root_keys
             .iter()
-            .find(|(keyid, _signing_key)| role_keys.keyids.contains(&keyid))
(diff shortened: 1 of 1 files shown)


webern (approved):
Looks right to me.

tough/src/editor/signed.rs
```diff
@@ -54,12 +54,9 @@ where
         // Ensure the keys we have available to us will allow us
         // to sign this role. The role's key ids must match up with one of
         // the keys provided.
-        let (signing_key_id, signing_key) = root_keys
+        let valid_keys = root_keys
```
zmrow, line 57: If we don't find any keys, I guess we would catch that with the `role.signatures.len()` check below?

Also - if we don't use this error variant anymore, let's make sure it gets deleted.

ecpullen: That's correct, and the error is used later as well.

iliana (approved):

zmrow (approved):
🌃

PR #198, opened 2020-07-27
 Implemented tough-kms to include AWS KMS as a new keysource

srgothi92:
*Issue #, if available:*
N/A

*Description of changes:*
-tough-kms implements the "KeySource" trait found in tough
-Unit test cases are added for tough-kms utilizing rusoto_mock
-Integration Tests are added in tuftool, to test 3 keysource: tough-kms, tough-ssm, local key.
-To run integration tests use feature flag "integ" while running test

Testing Done:
Ran Unit tests
Ran Integration Tests
Manually created root.json and used kms key for signing.


tough-kms/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+name = "tough-kms"
+version = "0.1.0"
+description = "Implements AWS KMS as a key source for TUF signing keys"
+authors = ["Shailesh Gothi <gothisg@amazon.com>"]
+license = "MIT OR Apache-2.0"
+repository = "https://github.com/awslabs/tough"
+keywords = ["TUF", "KMS"]
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
zmrow: We can get rid of this line. :)

tough-kms/src/error.rs
```diff
@@ -0,0 +1,121 @@
...
+    },
+
+    #[snafu(display("Unable to determine region from profile '{}': {}", profile, source))]
+    RusotoRegionFromProfile {
+        profile: String,
+        source: rusoto_credential::CredentialsError,
+        backtrace: Backtrace,
+    },
+
+    #[snafu(display("Unknown AWS region \"{}\": {}", region, source))]
```
zmrow: nit: you can use single quotes around the region (sort of the like the error variant above this).

`"Unknown AWS region '{}': {}"`

tough-kms/src/error.rs
```diff
@@ -0,0 +1,121 @@
...
+    key_id,
+    profile.as_deref().unwrap_or(""),
+    key_id
+    ))]
+    KmsAliasCheck {
+        profile: Option<String>,
+        key_id: String,
+    },
+
+    #[snafu(display("Public key is none"))]
```
zmrow: Could we make this error message a bit friendlier? Are we saying the public key doesn't exist? Or it exists but is empty? Or...?

tough-kms/src/error.rs
```diff
@@ -0,0 +1,121 @@
...
+    ))]
+    KmsGetPublicKey {
+        profile: Option<String>,
+        key_id: String,
+        source: rusoto_core::RusotoError<rusoto_kms::GetPublicKeyError>,
+        backtrace: Backtrace,
+    },
+
+    #[snafu(display(
+    "Alias {} exist for aws-kms:://{}/{}",
```
zmrow: Is this supposed to say "Alias {} **exists**"?

tough-kms/src/error.rs
```diff
@@ -0,0 +1,121 @@
...
+    KmsAliasCheck {
+        profile: Option<String>,
+        key_id: String,
+    },
+
+    #[snafu(display("Public key is none"))]
+    PublicKeyNone,
+
+    #[snafu(display(
+    "Write in Not defined for aws-kms://{}/{}",
```
zmrow: I'm not sure what this means; could we clarify? :)

tough-kms/src/error.rs
```diff
@@ -0,0 +1,121 @@
...
+    profile.as_deref().unwrap_or(""),
+    key_id
+    ))]
+    WriteNotDefined {
+        profile: Option<String>,
+        key_id: String,
+    },
+
+    #[snafu(display(
+    "Key metatada is None while creating key for aws-kms://{}/{}",
```
zmrow: Rather than referring to Rust concepts of `Some() / None` in these error messages, let's describe the problem in prose. Are we trying to inform the user that the metadata is empty? doesn't exist?

tough-ssm/src/lib.rs
```diff
@@ -55,12 +55,11 @@ impl KeySource for SsmKeySource {
     fn write(
         &self,
         value: &str,
-        key_id_hex: &str,
     ) -> std::result::Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
         let ssm_client = client::build_client(self.profile.as_deref())?;
         let fut = ssm_client.put_parameter(rusoto_ssm::PutParameterRequest {
             name: self.parameter_name.to_owned(),
-            description: Some(key_id_hex.to_owned()),
+            description: Some(String::from("Tough Key")),
```
zmrow: What's the reason for this change?

tough/src/key_source.rs
```diff
@@ -17,11 +17,38 @@ pub trait KeySource: Debug + Send + Sync {
     fn as_sign(&self) -> Result<Box<dyn Sign>, Box<dyn std::error::Error + Send + Sync + 'static>>;
 
     /// Writes a key back to the `KeySource`
-    fn write(
+    fn write(&self, value: &str) -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>>;
+    /// Creates a key for key source, default implementation creates rsa key pair which can
+    /// be override by individual `KeySource` implementation
+    fn create(
```
zmrow: Hmmm, I'm a little worried about a "default" implementation for this method that assumes `openssl` is installed.  

I'm also wondering why we need a `create` method?  Is it a convenience just for `tuftool`?

tough/src/key_source.rs
```diff
@@ -17,11 +17,38 @@ pub trait KeySource: Debug + Send + Sync {
     fn as_sign(&self) -> Result<Box<dyn Sign>, Box<dyn std::error::Error + Send + Sync + 'static>>;
 
     /// Writes a key back to the `KeySource`
-    fn write(
+    fn write(&self, value: &str) -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>>;
```
zmrow: I think the only reason we have a `write` method on this trait is to account for `tuftool`'s usage of it for SSM, right?  Perhaps it makes sense to evaluate what it means to eliminate this method from the trait altogether.  

@iliana : thoughts?

tuftool/src/root.rs
```diff
@@ -242,34 +241,11 @@ impl Command {
...
-                status: output.status
-            }
-        );
-        let stdout =
-            String::from_utf8(output.stdout).context(error::CommandUtf8 { command_str })?;
-
-        let key_pair = parse_keypair(stdout.as_bytes()).context(error::KeyPairParse)?;
-        let key_id = hex::encode(add_key(&mut root.signed, roles, key_pair.tuf_key())?);
         key_source
-            .write(&stdout, &key_id)
```
zmrow: I think this may be the only place we do a `key_source.write()` correct?  If so, let's look harder at eliminating this method from `key_source` as mentioned.

tuftool/src/source.rs
```diff
@@ -55,7 +56,8 @@ pub(crate) fn parse_key_source(input: &str) -> Result<Box<dyn KeySource>> {
         .base_url(Some(&pwd_url))
         .parse(input)
         .context(error::UrlParse { url: input })?;
-
+    println!("{}", url.scheme());
```
zmrow: Are these `println!()`'s  here on purpose? Or were they leftover from debugging?  :)

tuftool/tests/create_repository_integration.rs
```diff
@@ -0,0 +1,248 @@
...
+mod test_utils;
+use assert_cmd::Command;
+use chrono::{Duration, Utc};
+use std::fs::File;
+use tempfile::TempDir;
+use tough::{ExpirationEnforcement, Limits, Repository, Settings};
+
+// This file include integration tests for KeySources: tough-ssm, tough-kms and local file key.
+// Since the tests are run using the actual "AWS Ssm and AWS Kms", you would have to configure
+// AWS account profile "thar-signing-root-pdx" with root permission.
```
zmrow: Could we please remove `pdx` and `thar` from the profile name? :)

tuftool/tests/create_repository_integration.rs
```diff
@@ -0,0 +1,248 @@
...
+// Refer https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-profiles.html to configure named profile.
+// To run test include feature flag 'integration-tests' like : "cargo test --features=integration-tests"
```
zmrow: Thanks for putting this under a feature flag! 🎉

tuftool/tests/create_repository_integration.rs
```diff
@@ -0,0 +1,248 @@
...
+
+fn initialie_root_json(root_json: &str) {
```
zmrow: I think you meant `initialize...` :)

tough/src/error.rs
```diff
@@ -478,6 +478,46 @@ pub enum Error {
 
     #[snafu(display("Role missing from snapshot meta: {}", name))]
     RoleNotInMeta { name: String },
+
+    #[snafu(display("Not able to Sign message for key-id {} : {}", key_id, source))]
```
zmrow: No need to capitalize `Sign`.

Nit: single quotes around the key-id: `...for key-id '{}':...`

tough/src/error.rs
```diff
@@ -478,6 +478,46 @@ pub enum Error {
...
+        backtrace: Backtrace,
+    },
+
+    #[snafu(display("Signature is None"))]
+    SignatureNotFound,
+
+    #[snafu(display("Key Id: {} is not sign and verify key", key_id))]
+    NotSignVerifyKeyType { key_id: String },
+
+    #[snafu(display("Failed to run {}: {}", command_str, source))]
```
zmrow: Could we single quote the command string to make it easier to see it's a command?

`Failed to run '{}':...`

tough-kms/src/lib.rs
```diff
@@ -0,0 +1,139 @@
...
+    }
+}
+
+/// Implements the KeySource trait.
+impl KeySource for KmsKeySource {
+    fn as_sign(
+        &self,
+    ) -> std::result::Result<Box<dyn Sign>, Box<dyn std::error::Error + Send + Sync + 'static>>
+    {
+        let kms_client = match self.client.clone() {
```
zmrow: Why do we have to `.clone()` here? I think we should just be able to match on `self.client`, no?

zmrow (commented):
Nice work!

I didn't get through `tough-kms/tests/all_test.rs`, but here's what I have so far.

tough/src/key_source.rs
srgothi92: I'm also wondering why we need a create method? Is it a convenience just for tuftool?
Indirectly Yes, create is used by gen-rsa-key which is for convenience.
But, it helps us with end to end test case and programmatic key generation wherever required.

srgothi92 (commented):
Added comments for Design related questions.

tough-ssm/src/lib.rs
srgothi92: KeyId added in the root.json file is no longer available with write method because key is created by 'create' method which call the write method. After all this succeeds gen-rsa-key adds the key to root.json.

tuftool/src/root.rs
srgothi92: Yes, this is the only place its used.

tough/src/key_source.rs
srgothi92: write is used by SSM to store the rsa key in a parameter.
write is used by KMS to assign alias to the generated Key.
write is used by localsource to write generated key in a file.


Previously, gen-rsa-key command creates the rsa-key pair and calls write method.
Now, gen-rsa-key calls the create method of keysource, which in turn calls the write method. After this code change, write method is not longer called directly. We can get rid of it or make it private.  

gen-rsa-key command is for convenience, but it helps in end to end testing and programmatically create key.

I was also wondering does it make sense to call 'add-key' command from 'gen-rsa-key' ? We have separate command 'add-key', which user can use if they wan't to add the key. Or we can all command 'gen-add-rsa-key'. :P

srgothi92 (commented):
I added few other comments in last conversation. But, I could not see them any more. Adding them back.

Cargo.toml
```diff
@@ -3,5 +3,6 @@ members = [
     "olpc-cjson",
     "tough",
     "tough-ssm",
+    "tough-kms",
```
webern, line 6: Also adda line to this file so it will be built during continuous integration. https://github.com/awslabs/tough/blob/498fb581bac6af4e3f93583e572939ca9bbbcb03/.github/workflows/rust.yml#L15

tough-kms/src/client.rs
```diff
@@ -0,0 +1,33 @@
...
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+use crate::error::{self, Result};
+use rusoto_core::{HttpClient, Region};
+use rusoto_credential::ProfileProvider;
+use rusoto_kms::KmsClient;
+use snafu::ResultExt;
+use std::str::FromStr;
+
+/// Builds an KMS client for a given profile name.
```
webern:
```suggestion
/// Builds a KMS client for a given profile name.
```

tough-kms/src/error.rs
webern: > nit: you can use single quotes around the region (sort of the like the error variant above this).

I agree. I noticed, because I'm comparing with tough-ssm, that it should be changed in that crate also.

tough-kms/src/lib.rs
```diff
@@ -0,0 +1,139 @@
+// Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: MIT OR Apache-2.0
```
webern, line 2: We should add these (or some subset), which I copied from tough/lib.rs. These will enforce certain lints. The most important ones are `#![deny(rust_2018_idioms)]` and `#![deny(missing_docs)]`. Once you add these you will have compilation errors that you have to fix.

```suggestion
// SPDX-License-Identifier: MIT OR Apache-2.0

#![forbid(missing_debug_implementations, missing_copy_implementations)]
#![deny(rust_2018_idioms)]
// missing_docs is on its own line to make it easy to comment out when making changes.
#![deny(missing_docs)]
#![warn(clippy::pedantic)]
#![allow(
    clippy::module_name_repetitions,
    clippy::must_use_candidate,
    clippy::missing_errors_doc
)]
```

tough-kms/src/lib.rs
webern: Maybe this, unless ownership is actually required.

```suggestion
        let kms_client = &match self.client {
```

tough-kms/src/lib.rs
```diff
@@ -0,0 +1,139 @@
...
+            public_key: pb_key,
+            signing_algorithm: KmsSigningAlgorithms::Rsa(String::from("RSASSA_PSS_SHA_256")),
+        }))
+    }
+
+    fn write(
+        &self,
+        value: &str,
+    ) -> std::result::Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
+        let kms_client = match self.client.clone() {
```
webern: See if this works (and self.pofile probably won't need as_deref() anymore?)

```suggestion
        let kms_client = match &self.client {
```

tough-kms/src/lib.rs
```diff
@@ -0,0 +1,139 @@
...
+            Some(value) => value,
+            None => client::build_client_kms(self.profile.as_deref())?,
+        };
+        // Assign an alias to the Key
+        let fut = kms_client.create_alias(rusoto_kms::CreateAliasRequest {
+            alias_name: self.key_id.clone(),
+            target_key_id: value.to_string(),
+        });
+        let _response = tokio::runtime::Runtime::new()
```
webern: Double check to make sure there are no hidden, error-like, conditions that would require checking something in the response.

```text
        let _response = tokio::runtime::Runtime::new()
```

tough-kms/src/lib.rs
```diff
@@ -0,0 +1,139 @@
...
+        _bits: u16,
+        _exponent: u32,
+    ) -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
+        let kms_client = match self.client.clone() {
+            Some(value) => value,
+            None => client::build_client_kms(self.profile.as_deref())?,
+        };
+        // Kms currently does not have an API to validate if the alias exist. As a Hack,
+        // or temporary resolution, we are using the get_public_key API, if it succeeds than
+        // Alias exist and alarm the user.
```
webern:
```suggestion
        // or temporary resolution, we are using the get_public_key API, if it succeeds then
        // an alias exists and we return an error.
```

tough-kms/src/lib.rs
```diff
@@ -0,0 +1,139 @@
...
+            .fail()?;
+        }
+        // Create a new Customer managed Key in KMS
+        let fut = kms_client.create_key(rusoto_kms::CreateKeyRequest {
+            customer_master_key_spec: Some(String::from("RSA_2048")),
+            description: Some(String::from("Tuf repository signing key")),
+            key_usage: Some(String::from("SIGN_VERIFY")),
+            ..rusoto_kms::CreateKeyRequest::default()
+        });
+        let key_id = tokio::runtime::Runtime::new()
```
webern: I think you could do this whole function in a single tokio runtime.

```
        let key_id = tokio::runtime::Runtime::new()
```

tough-ssm/src/lib.rs
webern: > KeyId added in the root.json file is no longer available with write method because key is created by 'create' method which call the write method. After all this succeeds gen-rsa-key adds the key to root.json.

This seems to be the only change in the tough-ssm crate. How did changes elsewhere affect the behaviour in this crate?

edit: later I saw the changes in `tough`...

tough/Cargo.toml
```diff
@@ -17,10 +18,13 @@ olpc-cjson = { version = "0.1.0", path = "../olpc-cjson" }
 pem = "0.8.1"
 reqwest = { version = "0.10.4", optional = true, default-features = false, features = ["blocking"] }
 ring = { version = "0.16.13", features = ["std"] }
+rusoto_core = { version = "0.44", optional = true, default-features = false }
+rusoto_kms = { version = "0.44", optional = true, default-features = false }
```
webern: The purpose of placing the KMS feature in its own crate is to avoid introducing these dependencies to `tough`.

```suggestion
```

tough/Cargo.toml
```diff
@@ -17,10 +18,13 @@ olpc-cjson = { version = "0.1.0", path = "../olpc-cjson" }
...
 serde = { version = "1.0.110", features = ["derive"] }
 serde_json = "1.0.53"
 serde_plain = "0.3.0"
 snafu = "0.6.8"
+tokio = "0.2.13"
```
webern: This should also be a transient dependency, not needed directly by `tough`. (i.e. needed by tough-ssm, but not in tough's cargo.toml)

```suggestion
```

tough/src/key_source.rs
```diff
@@ -17,11 +17,38 @@ pub trait KeySource: Debug + Send + Sync {
...
+        bits: u16,
+        exponent: u32,
+    ) -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
+        // ring doesn't support RSA key generation yet
+        // https://github.com/briansmith/ring/issues/219
+        let mut command = std::process::Command::new("openssl");
+        command.args(&["genpkey", "-algorithm", "RSA", "-pkeyopt"]);
+        command.arg(format!("rsa_keygen_bits:{}", bits));
+        command.arg("-pkeyopt");
+        command.arg(format!("rsa_keygen_pubexp:{}", exponent));
```
webern: This we definitely should not do.
```
        let mut command = std::process::Command::new("openssl");
        command.args(&["genpkey", "-algorithm", "RSA", "-pkeyopt"]);
        command.arg(format!("rsa_keygen_bits:{}", bits));
        command.arg("-pkeyopt");
        command.arg(format!("rsa_keygen_pubexp:{}", exponent));
```

tough/src/key_source.rs
webern: > gen-rsa-key command is for convenience, but it helps in end to end testing and programmatically create key.

I think we should take a closer look at this design, there may be another way. Generally speaking we shouldn't be adding too many dependencies or assumptions to the core library (`tough`).

webern (commented):
I didn't get all the way through this. It seems like something is wrong with our design if adding a new key source causes a lot of churn in `tough`. The idea behind the key source trait was to ensure that `tough` (to the extent possible) doesn't know anything about what's under the hood in a key source.

Leaving the 'heavy lifting' of creating a key outside the realm of `tough` would be better if it prevents us from adding dependencies to `tough`.

I know you're working with Zack on this. I'd be happy to join a call to join any discussion. Thanks!

Nice start!

iliana:
First, run `cargo check --all` to regenerate the Cargo.lock, and commit that. CI is requiring that your Cargo.lock is accurate and maps to the dependencies you are using across the project.

No Rusoto dependencies should exist in the main tough crate. I moved all of the added structs to `tough-kms` locally and realized the reason you may have done this: the `Sign` trait's `sign` method returns a `Result<T, tough::Error>` instead of allowing the implementation to return a more generic error. Because we have to implement `Sign` differently for KMS, this is impossible without adding the code directly to `tough`.

If you're comfortable with it, modify the `Sign::sign` trait signature to return a generic error, similar to the methods in `KeySource` (they return a `Result<T, E>` where E is `Box<dyn std::error::Error + Send + Sync + 'static>>`). Alternately, if you like, I can make this change in a separate PR, and you can rebase on top of it and carry on.

srgothi92:
Response to @iliana comment:
1:
Thanks.

2: 
Not just because of `Sign` trait's `sign` method returns a `Result<T, tough::Error>`. In Sign object I  would need KmsClient to sign the messages. For `LocalKeySource` and `tough-ssm`, sign object has private key available to sign the messages. Which is not the case with Kms. The solution I currently implemented adds `rusoto_kms` as dependency and makes a call to KMS directly using `Kmsclient` available with `Sign` trait. Which definitely is not the right way. 
Solution Suggestion 1:  I am thinking, if we can expose a module in `tough_kms` that is used by `Sign` trait to sign the messages, avoiding direct dependency on `KmsClient`, not sure if it would result in cyclic dependency. tough_kms provides sign object and sign object needs tough_kms. 
Solution Suggestion 2: We can get rid of `Sign` trait and let `KeySource` expose `sign` and `tuf_key` method. Its upto keySource implementer to cache the private key and use the cached key for multiple `sign` request.   

3: 
I can take care of it, if required.

iliana:
> In Sign object I would need KmsClient to sign the messages.

`KeySource::as_sign` returns a `Box<dyn Sign>`, and `KmsRsaKey` (which has your `KmsClient`) implements `Sign`. This separation allows you to move `KmsRsaKey` to tough-kms.

zmrow:
We just realized that we probably shouldn't create a tokio runtime inside these libraries, so we probably shouldn't ship this until https://github.com/awslabs/tough/issues/213 is fixed.  :(

webern (approved):
Nice! As I understand it, because tough-kms needs to implement sign, the sign trait was changed to return a generic error to break dependency cycle. Looks good, looks clean.

iliana:
Merging this into develop fails to compile — there are two new `Sign` implementations that need to have their signatures fixed. You'll probably want to rebase and resolve those issues.

I'm fixing that locally and continuing to review.

tuftool/tests/create_repository_integration.rs
```diff
@@ -0,0 +1,268 @@
...
+    assert_eq!(
+        repo.timestamp().signed.meta["snapshot.json"].version.get(),
+        snapshot_version
+    );
+    assert_eq!(repo.snapshot().signatures.len(), 1);
+    root_json_dir.close().unwrap();
+}
+
+#[test]
+#[cfg_attr(not(feature = "integration-tests"), ignore)]
```
iliana: We use `integ` as the feature name in tough, so we should do the same here.

tuftool/tests/create_repository_integration.rs
```diff
@@ -0,0 +1,268 @@
...
+    let root_key_path = root_key_dir.path().join("local_key.pem");
+    let root_key = &format!("file://{}", root_key_path.to_str().unwrap());
+    create_repository(root_key, true);
+}
+
+#[test]
+#[cfg_attr(not(feature = "integration-tests"), ignore)]
+// Ensure we can use ssm key to create and sign a repo created by the `tuftool` binary using the `tough` library
+fn create_repository_ssm_key() {
+    let root_key = &format!("aws-ssm://{}/thar/root-key-new", PROFILE);
```
iliana: These should be key sources a tough developer can set up once and continue to use, so we should have names that indicate they're for testing.

```suggestion
    let root_key = &format!("aws-ssm://{}/tough-integ/key-a", PROFILE);
```

tuftool/tests/create_repository_integration.rs
```diff
@@ -0,0 +1,268 @@
...
+    create_repository(root_key, true);
+}
+
+#[test]
+#[cfg_attr(not(feature = "integration-tests"), ignore)]
+// Ensure we can use kms key to create and sign a repo created by the `tuftool` binary using the `tough` library
+fn create_repository_kms_key() {
+    let root_key = &format!("aws-kms://{}/alias/test_key", PROFILE);
```
iliana:
```suggestion
    let root_key = &format!("aws-kms://{}/alias/tough-integ/key-a", PROFILE);
```

tuftool/tests/create_repository_integration.rs
```diff
@@ -0,0 +1,268 @@
...
+    assert_eq!(repo.snapshot().signatures.len(), 1);
+    root_json_dir.close().unwrap();
+}
+
+#[test]
+#[cfg_attr(not(feature = "integration-tests"), ignore)]
+// Ensure we can use local rsa key to create and sign a repo created by the `tuftool` binary using the `tough` library
+fn create_repository_local_key() {
+    let root_key_dir = TempDir::new().unwrap();
+    let root_key_path = root_key_dir.path().join("local_key.pem");
```
iliana, line 224: This could instead use `snakeoil.pem` from tough's test data, and then this wouldn't need to be an integration test.

tough-kms/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+
+[dependencies]
+tough = { version = "0.8.0", path = "../tough", features = ["http"] }
+ring = { version = "0.16.13", features = ["std"] }
+rusoto_core = { version = "0.44", optional = true, default-features = false }
+rusoto_credential = { version = "0.44", optional = true }
+rusoto_kms = { version = "0.44", optional = true, default-features = false }
+bytes = "0.5.6"
+serde = "1.0.105"
+serde_json = "1.0.50"
```
iliana: These two dependencies only seem to be used in tests, so they should be `[dev-dependencies]`.

tuftool/tests/create_repository_integration.rs
```diff
@@ -0,0 +1,268 @@
...
+
+fn initialize_root_json(root_json: &str) {
+    Command::cargo_bin("tuftool")
+        .unwrap()
+        .args(&["root", "init", root_json.clone()])
+        .assert()
+        .success();
+    Command::cargo_bin("tuftool")
+        .unwrap()
+        .args(&["root", "expire", root_json.clone(), "2020-09-22T00:00:00Z"])
```
iliana: Seems like this will make tests start failing in a month... :)

```suggestion
        .args(&["root", "expire", root_json.clone(), "3030-09-22T00:00:00Z"])
```

tuftool/tests/create_repository_integration.rs
```diff
@@ -0,0 +1,268 @@
...
+// Refer https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-profiles.html to configure named profile.
+// Additionally, tough-kms key generation is not supported (issue  #211), so you would have to manually create kms CMK key.
+// To run test include feature flag 'integration-tests' like : "cargo test --features=integration-tests"
+
+const PROFILE: &str = "test-profile";
+
+fn initialize_root_json(root_json: &str) {
+    Command::cargo_bin("tuftool")
+        .unwrap()
+        .args(&["root", "init", root_json.clone()])
```
iliana: Is cloning `root_json` necessary here? (`root_json` is already a reference so I think clone does nothing...)

srgothi92:
> Merging this into develop fails to compile — there are two new Sign implementations that need to have their signatures fixed. You'll probably want to rebase and resolve those issues.

> I'm fixing that locally and continuing to review.

Done

tough-kms/src/lib.rs
```diff
@@ -0,0 +1,171 @@
...
+use tough::schema::decoded::{Decoded, RsaPem};
+use tough::schema::key::{Key, RsaKey, RsaScheme};
+use tough::sign::Sign;
+
+/// Represents a Signing Algorithms for AWS KMS.
+#[derive(Debug, Clone, PartialEq)]
+pub enum KmsSigningAlgorithms {
+    /// The key type
+    Rsa(String),
+}
```
iliana: I'm confused about why we need this type. What advantage does this have over `KmsRsaKey.signing_algorithm` just being a `String`?

tough-kms/src/lib.rs
```diff
@@ -0,0 +1,171 @@
...
+    pub key_id: String,
+    /// Aws account profile
+    pub profile: Option<String>,
+    /// KmsClient Object to query AWS KMS
+    pub client: Option<KmsClient>,
+    /// Public Key corresponding to Customer Managed Key
+    pub public_key: Decoded<RsaPem>,
+    /// Signing Algorithm to be used for the Customer Managed Key
+    pub signing_algorithm: KmsSigningAlgorithms,
+}
```
iliana: I don't think these fields should be `pub` — anything going on here is purely an implementation detail.

tough-kms/src/lib.rs
```diff
@@ -0,0 +1,171 @@
...
+        _rng: &dyn SecureRandom,
+    ) -> Result<Vec<u8>, Box<dyn std::error::Error + Send + Sync + 'static>> {
+        let kms_client = match self.client.to_owned() {
+            Some(value) => value,
+            None => client::build_client_kms(self.profile.as_deref())?,
+        };
+        let sign_fut = kms_client.sign(SignRequest {
+            key_id: self.key_id.clone(),
+            message: bytes::Bytes::from(msg.to_owned()),
+            message_type: Some(String::from("RAW")),
```
iliana: Apparently [messages can only be 4096 bytes long](https://docs.aws.amazon.com/kms/latest/APIReference/API_Sign.html#KMS-Sign-request-Message), so we will want to use digest mode here.

Along with a `use ring::digest::{digest, SHA256};` up top, I've replaced this with:

```suggestion
            message: digest(&SHA256, msg).as_ref().to_vec().into(),
            message_type: Some(String::from("DIGEST")),
```

(This doesn't fix the mysterious invalid signature issue, which I'm working through separately.)

iliana (commented):
A first pass of thoughts.

I'm unable to verify signatures produced by this code. I'm investigating why, because the answer is not obvious to me.

tuftool/tests/create_repository_integration.rs
srgothi92: I din't use `integ` because it says `Integration tests require docker` in [docs](https://docs.rs/tough/0.8.0/tough/), but my itegration test needs a different setup. Let me know if u still need me to change it to `integ`.

tuftool/tests/create_repository_integration.rs
srgothi92: Done :)

tuftool/tests/create_repository_integration.rs
srgothi92: Done :)

tuftool/tests/create_repository_integration.rs
srgothi92: I am creating a new one because I wanted to test end to end. I created a separate method for whole repository creation process, So that we can just use different types of key and ensure they all adhere to the same workflow of consumption. Unfortunately, `tough-kms` had to add an extra flag for creation, but it will be same in future and this test cases ensures that.

tough-kms/Cargo.toml
srgothi92: Yes, :). Awesome that u were able to see that from code review.

tuftool/tests/create_repository_integration.rs
srgothi92: Oops. Done.

tuftool/tests/create_repository_integration.rs
srgothi92: Nice, Done. Even changed for `key.clone()`.

tough-kms/src/lib.rs
srgothi92: Yes, It is confusing, I would like to have your feedback on this.  We had a discussion on question as described below
>  KMS supports a lot of signing methods

> > Iliana Reply: What I'd recommend is a match statement, even though the enum has one field. That way if we add new ones because TUF allows new ones, it's more clear that we need to change downstream code.

It was added in context of this, to keep it generic and make decision of signing algorithm in  `as_sign` while returning the `sign` object:
```
Ok(Box::new(KmsRsaKey {
            profile: self.profile.clone(),
            client: Some(kms_client.clone()),
            key_id: self.key_id.clone(),
            public_key: pb_key,
            signing_algorithm: KmsSigningAlgorithms::Rsa(String::from("RSASSA_PSS_SHA_256")),
        }))
```

However, It does not solve the problem in a way I wanted to.  I want something like object of algorithms to select from.
Ex:-
```
algorithms = {
   algo1 : "algo1",
   algo2: "algo2"
} 
```
 And we can directly use as `algorithms.algo1` or `algorithms.algo2` based on some conditions in `as_sign` in future when we support more signing algorithms.

tough-kms/src/lib.rs
srgothi92: Done.

tough-kms/src/lib.rs
srgothi92: Done. I wanted to use "DIGEST" but couldn't figure out `digest(&SHA256, msg).as_ref().to_vec().into()` part

srgothi92:
Fixed some of the code review comments, with multiple stupid force pushes. you might have to look at two of them to see the complete changes.

tuftool/tests/create_repository_integration.rs
iliana: Please do, yeah. We can fix the docs.

tuftool/tests/create_repository_integration.rs
iliana: I misread the test, I understand now.

iliana:
ae2f466 fixes the verification issue! The commit message explains, but when we initially discussed how to go about creating the `Decoded<RsaPem>` I forgot that the byte form of it needs to be an `RSAPublicKey` document, not a `SubjectPublicKeyInfo` document. This confusion caused us to double-wrap the public key in the SPKI wrapper, resulting in this extremely confusing situation:

```
$ jq -r .signed.keys.a6c9b56dd7ce0a87fab1fa1c39d37afefcfb0692c05b2ddaa207919f4d1173d1.keyval.public ~/tmp/root.json \
> | openssl asn1parse
    0:d=0  hl=4 l= 442 cons: SEQUENCE          
    4:d=1  hl=2 l=  13 cons: SEQUENCE          
    6:d=2  hl=2 l=   9 prim: OBJECT            :rsaEncryption
   17:d=2  hl=2 l=   0 prim: NULL              
   19:d=1  hl=4 l= 423 prim: BIT STRING        
$ jq -r .signed.keys.a6c9b56dd7ce0a87fab1fa1c39d37afefcfb0692c05b2ddaa207919f4d1173d1.keyval.public ~/tmp/root.json \
> | openssl asn1parse -strparse 19
    0:d=0  hl=4 l= 418 cons: SEQUENCE          
    4:d=1  hl=2 l=  13 cons: SEQUENCE          
    6:d=2  hl=2 l=   9 prim: OBJECT            :rsaEncryption
   17:d=2  hl=2 l=   0 prim: NULL              
   19:d=1  hl=4 l= 399 prim: BIT STRING        
```

tough-kms/tests/data/response_public_key.json
```diff
@@ -0,0 +1,3 @@
+{
+  "PublicKey" : "Ii0tLS0tQkVHSU4gUFVCTElDIEtFWS0tLS0tXG5NSUlCb2pBTkJna3Foa2lHOXcwQkFRRUZBQU9DQVk4QU1JSUJpZ0tDQVlFQW5MNnU2UTlRNnBnMUc1MDIwYTgzXG5HbEgvYUZVTzBQUTVsZUlwd1dMOGtXZ3BhV3VVRzdvUmxPVUcyLzRjd041RkN2SkpHWHFVNUF0U0txMmZaNDJKXG41WFI5UU1pcDRQZzBRNm1FOFhDdkFYQW9NbmtXU2NoZHpnVDJHb0VudGFPZVJSVENVR2IvRHNWb3hzVlhqVjZtXG5GYVJNeDduaDhnZ3NoTVdnVFlnVFVESytDU0lCQ2NCV2FwQ0ZxMUJyTTYwWFptR1RxZUF1SFNIYVVVdUY5RzNiXG5nT2ZsSDVMOUlwUWthSFdiSnRHdnlLTHI1M21oV08ycjhCUFIzK0N0TlpvakFua3dtdTRsQTk0azhDN1RMTWRjXG51dHpVNE96T0RlOVVQRVJjMzNsUnY4REJnc0gzRjA3N1pRd3YvaWtaWFdTbEFDVERXWndlbm5jQ0V3cWRlRGQ0XG4rcTJBSHlxeFJON2JVQWg1N21VTitrRmQzU1MvNFQ0NHNmQnJKdzZONEpWL21FKy9ZZlJMV3RwSUtJc1huQkNiXG5yQytkdDk2VnF6Nmc2ZVZWdnFQd2hPQ1NLY1lzbXAvaVM2cXdWbjBEcTJTQ3JHRzFGVG1CamVBOVprY2paaFVHXG5RRU15TU5ob1MrVTJOeDVvSUVJcTJrUkVwdXUrS3NCU1RVYU9nUjA3V05VeEFnTUJBQUU9XG4tLS0tLUVORCBQVUJMSUMgS0VZLS0tLS1cbiI="
```
iliana: This doesn't seem accurate based on the response I see from the KMS API. This is a base64'd PEM-formatted key, and the API seems to return a base64'd DER-formatted key.

tuftool/tests/create_repository_integration.rs
iliana: Well, more specifically, that's the `tough` docs, and `tuftool`'s definition of integration tests can mean "needs to have working AWS credentials". :)

tuftool/tests/create_repository_integration.rs
```diff
@@ -0,0 +1,240 @@
...
+    let root_key = &format!("file://{}", root_key_path.to_str().unwrap());
+    create_repository(root_key, true);
+}
+
+#[test]
+#[cfg_attr(not(feature = "integration-tests"), ignore)]
+// Ensure we can use ssm key to create and sign a repo created by the `tuftool` binary using the `tough` library
+fn create_repository_ssm_key() {
+    let root_key = &format!("aws-ssm://{}/tough-integ/key-a", PROFILE);
+    create_repository(root_key, true);
```
iliana: Now that I understand what's going on here: it might be good to confirm that you can run this twice in a row. I assume tough-ssm will happily override a given parameter name.

tuftool/tests/create_repository_integration.rs
```diff
@@ -0,0 +1,240 @@
...
+use assert_cmd::Command;
+use chrono::{Duration, Utc};
+use std::fs::File;
+use tempfile::TempDir;
+use tough::{ExpirationEnforcement, Limits, Repository, Settings};
+
+// This file include integration tests for KeySources: tough-ssm, tough-kms and local file key.
+// Since the tests are run using the actual "AWS SSM and AWS KMS", you would have to configure
+// AWS account profile "test-profile" with root permission.
+// Refer https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-profiles.html to configure named profile.
```
iliana: I would prefer to use the default profile here, actually. People can run with `AWS_PROFILE=test-profile cargo test --features=integ`.

The default credential searcher should be preferred because we don't know how people develop this, and if developers want to use EC2 IMDS credentials or environment credentials, we shouldn't get in their way of testing.

tuftool/tests/create_repository_integration.rs
```diff
@@ -0,0 +1,240 @@
...
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+mod test_utils;
+use assert_cmd::Command;
+use chrono::{Duration, Utc};
+use std::fs::File;
+use tempfile::TempDir;
+use tough::{ExpirationEnforcement, Limits, Repository, Settings};
+
+// This file include integration tests for KeySources: tough-ssm, tough-kms and local file key.
```
iliana, line 12: These integration tests should also attempt to load the repositories.

tough-kms/src/lib.rs
iliana: OK, I think I see what's necessary here. We need to support a way of saying "use this particular signing algorithm with KMS" in case the TUF spec adds additional signing algorithms for the same kind of key.

If that's the case, it needs to be a member of `KmsKeySource` so that the user can specify a different algorithm. And I would avoid using strings at all in this circumstance.

Here's what I'd propose:

```rust
#[non_exhaustive]
pub enum KmsSigningAlgorithm {
    RsassaPssSha256,
}

pub struct KmsKeySource {
    ...
    pub signing_algorithm: KmsSigningAlgorithm,
}
```

Then, verify the GetPublicKey response contains a value that matches `signing_algorithm`, returning an error if there's a mismatch (perhaps the key in KMS is ECDSA).

Because there's only one option implemented for now, I don't think we need to worry about wiring this up all the way through tuftool yet.

tuftool/tests/create_repository_integration.rs
srgothi92: Yes you can run twice in a row and `tough-ssm` will override the given parameter name.

tuftool/tests/create_repository_integration.rs
srgothi92: It does, that's how I got to know verify is failing.

iliana: Oh! :)

tuftool/tests/create_repository_integration.rs
iliana: (Clarifying -- default credential provider, which will go through profiles, IMDS, environment variables, etc)

srgothi92:
PR comments fixes.

tuftool/tests/create_repository_integration.rs
srgothi92: Done

tough-kms/tests/data/response_public_key.json
srgothi92: Done

tough-kms/src/lib.rs
srgothi92: Done.

tuftool/tests/create_repository_integration.rs
srgothi92: Done

webern: Yeah, it might be easiest (maybe per crate), to have a single `integ` feature and list all the things you need to run integration tests.

Edit: aaaactually, using an AWS Account is significantly more arduous than having docker available. But the current setup is fine since the two crates have different requirements for the `integ` feature, and these are documented...

webern (approved):

iliana:
> We just realized that we probably shouldn't create a tokio runtime inside these libraries, so we probably shouldn't ship this until #213 is fixed. :(

I'd rather merge this first, then do the async party. :)

tough-kms/src/lib.rs
```diff
@@ -0,0 +1,203 @@
...
+pub enum KmsSigningAlgorithm {
+    /// Signing Algorithm `RSASSA_PSS_SHA_256`
+    RsassaPssSha256,
+}
+
+impl Default for KmsSigningAlgorithm {
+    fn default() -> Self {
+        KmsSigningAlgorithm::RsassaPssSha256
+    }
+}
```
iliana: We shouldn't do this; we want the user of the key source to always be aware of what key signing algorithm is in use, and a default implementation makes less sense when we eventually add ECDSA support to tough-kms.

For now, tuftool can hardcode the `RsassaPssSha256` variant; we'll figure out a solution there for ECDSA support.

```suggestion
```

tough-kms/src/lib.rs
```diff
@@ -0,0 +1,203 @@
...
+                line_ending: pem::LineEnding::LF,
+            },
+        );
+        if !response
+            .signing_algorithms
+            .context(error::MissingSignAlgorithm)?
+            .contains(&self.signing_algorithm.value())
+        {
+            error::ValidSignAlgorithm.fail()?;
+        }
```
iliana: Prefer snafu's [`ensure!`](https://docs.rs/snafu/0.6.8/snafu/macro.ensure.html) macro.

```suggestion
        ensure!(
            !response
                .signing_algorithms
                .context(error::MissingSignAlgorithm)?
                .contains(&self.signing_algorithm.value()),
            error::ValidSignAlgorithm,
        );
```

tough-kms/src/lib.rs
```diff
@@ -0,0 +1,203 @@
...
+
+impl KmsSigningAlgorithm {
+    fn value(self) -> String {
+        // Currently we are supporting only single algorithm, but code stub is added to support
+        // multiple algorithms in future.
+        match self {
+            #![allow(clippy::wildcard_in_or_patterns)]
+            #[allow(unreachable_patterns)]
+            KmsSigningAlgorithm::RsassaPssSha256 | _ => String::from("RSASSA_PSS_SHA_256"),
+        }
```
iliana:
```suggestion
        String::from(match self {
            KmsSigningAlgorithm::RsassaPssSha256 => "RSASSA_PSS_SHA_256",
        })
```

tough-kms/src/lib.rs
```diff
@@ -0,0 +1,203 @@
...
+    }
+}
+
+/// Implements the `KeySource` trait for keys that live in AWS KMS
+#[derive(Default)]
```
iliana:
```suggestion
```

tuftool/tests/create_repository_integration.rs
```diff
@@ -0,0 +1,246 @@
...
+// AWS credentials with root permission.
+// Refer https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-profiles.html to configure named profile.
+// Additionally, tough-kms key generation is not supported (issue  #211), so you would have to manually create kms CMK key.
+// To run test include feature flag 'integ' like : "cargo test --features=integ"
+
+fn get_profile() -> String {
+    match env::var("AWS_PROFILE") {
+        Ok(value) => value,
+        Err(_e) => String::from("default"),
+    }
```
iliana: If we provide an empty string for the profile, it will use the default credentials provider, which permits usage of environment/IMDS credentials and the like. (And the `Default` impl for String is an empty string, so...)

```suggestion
    env::var("AWS_PROFILE").unwrap_or_default()
```

iliana (commented):
Final stretch!

tough-kms/src/lib.rs
srgothi92: Done.

tough-kms/src/lib.rs
srgothi92: Done.

tough-kms/src/lib.rs
srgothi92: Done.

tough-kms/src/lib.rs
srgothi92: Done.

tuftool/tests/create_repository_integration.rs
srgothi92: That was way simplified 👍 . Done.

srgothi92:
Fixed PR comments. Check 1st `force-pushed` for all relevant changes, 2nd was for missed `cargo fmt -- --check`

zmrow (approved):
🌃 

lgtm

iliana (approved):

Issue #126, opened 2020-05-28
Unify file naming logic


zmrow:
Currently whenever a user must refer to a file for a role or target, they must build the filename using the version or, in the case of targets, the sha256.  We should unify this naming in a single place.

https://github.com/awslabs/tough/blob/abb33f5b1c9d3ec358c09bf604ed8f64394c3e79/tough/src/lib.rs#L780-L784

We have this naming logic in 4 places now, maybe we should do it in one place?  (`tuftool/tests/refresh_command`, `Repository::*target_digest_and_filename`, `lib::load_targets`, and now here)

_Originally posted by @tjkirch in https://github.com/awslabs/tough/pull/125_

zmrow:
My initial idea for at least the role naming is a trait they could all implement. Using the supplied root (just to see if the `consistent_snapshot` is true), it could spit out the proper file name using its version.

```
trait Filename {
    fn filename(&self, root: &Root) -> String
}
```

PR #200, opened 2020-07-31
Adds targets.rs a TargetsEditor

ecpullen:
**This is the first part of a set of 4 PRs to finish implementing delegated targets.**
*Issue #, if available:*
Related to #5
*Description of changes:*
Add a `TargetsEditor` that can be used to edit a `Targets` metadata file. Upcoming PRs will integrate `TargetsEditor` into `RepositoryEditor` and `tuftool`

`TargetsEditor` contains the various bits of data needed to construct
 or edit a `Targets` role. A new Targets may be created using the `new()` method. An existing `Targets` may be loaded and edited using the `from_targets()` method. When a targets is loaded in this way, versions and expirations are discarded. It is good practice to update these whenever a repo is changed. A  `Targets` from an existing repository can be loaded using the `from_repo()` method. `Targets` loaded this way will have the versions and expirations removed, but the proper keyholder to sign the targets and the `Transport` used to load the repo will be saved. Targets, versions, and expirations may be added to their respective roles via the provided "setter" methods. The final step in the process is the `sign()` method, which takes a given set of signing keys, builds each of the roles using the data provided, and signs the roles. This results in a `SignedDelegatedTargets` which can be used to write the updated metadata to disk.

Adds `KeyHolder` to use `Root` or `Delegations` to sign roles.

Adds `DelegatedTargets` to wrap `Targets` with its name.

Adds method to `Role` to unify file naming logic #126 

*Testing*
Testing will be included once all PRs have landed.


tough/src/schema/mod.rs
```diff
@@ -701,47 +1028,14 @@ impl Delegations {
...
-            valid >= u64::from(role_keys.threshold),
-            error::SignatureThreshold {
-                role: RoleType::Targets,
-                threshold: role_keys.threshold,
-                valid,
-            }
-        );
-        Ok(())
-    }
-
```
ecpullen, line 738: Moved to verify.rs

tough/src/editor/keys.rs
```diff
@@ -1,21 +1,66 @@
...
-use snafu::{ensure, ResultExt};
+use snafu::{ensure, OptionExt, ResultExt};
 use std::collections::HashMap;
 
-/// A map of key ID (from root.json) to its corresponding signing key
-pub(crate) type RootKeys = HashMap<Decoded<Hex>, Box<dyn Sign>>;
+/// A map of key ID (from root.json or the Delegations field of any Targets) to its corresponding signing key
+pub(crate) type KeyList = HashMap<Decoded<Hex>, Box<dyn Sign>>;
+
+impl KeyHolder {
```
zmrow, line 15: It feels a little odd to have `KeyHolder` defined somewhere else and the `impl` here. Does it make sense to move it here? Or move the `impl` there?

Maybe I'm just missing something. :)

tough/src/editor/keys.rs
```diff
@@ -1,21 +1,66 @@
 // Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: MIT OR Apache-2.0
+
+//! Provides a `KeyHolder` object for managing signing
```
zmrow: This module doesn't provide the `KeyHolder` right?

tough/src/editor/keys.rs
```diff
@@ -1,21 +1,66 @@
...
+
 use crate::error::{self, Result};
 use crate::key_source::KeySource;
 use crate::schema::decoded::{Decoded, Hex};
-use crate::schema::Root;
+use crate::schema::{Delegations, KeyHolder, Root};
+use crate::schema::{RoleId, RoleKeys};
```
zmrow: These imports can all be combined.

tough/src/schema/mod.rs
```diff
@@ -47,11 +47,22 @@ pub enum RoleType {
     /// The timestamp role is used to prevent an adversary from replaying an out-of-date signed
     /// metadata file whose signature has not yet expired.
     Timestamp,
+    /// A targets role that is not the top level targets
+    DelegatedTargets,
```
zmrow: We could probably just say: `A delegated targets role`. Having a different type makes it clear that it's not *the* `Targets` role. :)

tough/src/schema/mod.rs
```diff
@@ -93,6 +112,16 @@ pub struct Signature {
     pub sig: Decoded<Hex>,
 }
 
+/// A `KeyHolder` is metadata that is responsible for verifying the signatures of a role.
+/// `KeyHolder` contains
```
zmrow: Contains what? This thought never got finished... :)

tough/src/editor/keys.rs
zmrow: @iliana may have opinions on this.

tough/src/schema/mod.rs
```diff
@@ -47,11 +47,22 @@ pub enum RoleType {
...
 }
 
 forward_display_to_serde!(RoleType);
 forward_from_str_to_serde!(RoleType);
 
+/// A role identifier
+#[derive(Debug, Clone)]
+pub enum RoleId {
+    /// Top level roles are identified by a RoleType
+    TopLevel(RoleType),
```
zmrow: I understand what "top level" means because I've been talking with you and digging around in here... I'm wondering if it's clear to others. 

Perhaps it should be called `StandardRole()` since these type of Roles will be present in all TUF repos?

I may also be overthinking it. :)

tough/src/schema/mod.rs
```diff
@@ -64,6 +75,14 @@ pub trait Role: Serialize {
     /// number less than the one currently trusted.
     fn version(&self) -> NonZeroU64;
 
+    /// The filename that the role metadata should be written to
+    fn filename(&self, consistent_snapshot: bool) -> String;
```
zmrow, line 79: I like this idea since it moves all file naming concerns to a single place.

tough/src/schema/mod.rs
```diff
@@ -514,7 +555,27 @@ impl Targets {
         targets
     }
 
-    /// Returns a vec of all rolenames
+    ///Returns a hashmap of all targets and all delegated targets recursively with consistent snapshot names
```
zmrow: `...targets recursively. Targets are named per consistent snapshot rules`  

or something?

Also, there should be a space between the slashes and the docstring.

tough/src/schema/mod.rs
```diff
@@ -514,7 +555,27 @@ impl Targets {
...
+            );
+        }
+        if let Some(delegations) = &self.delegations {
+            targets.extend(delegations.targets_map_consistent());
+        }
+
+        targets
+    }
+
+    ///Returns a vec of all rolenames
```
zmrow: Missing space between slash and docstring here as well. :)

tough/src/schema/mod.rs
```diff
@@ -528,6 +589,170 @@ impl Targets {
 
         roles
     }
+
+    ///recursively clears all targets
```
zmrow: Here also (also please capitalize)

tough/src/schema/mod.rs
```diff
@@ -528,6 +589,170 @@ impl Targets {
...
+            for delegated_role in &mut delegations.roles {
+                if let Some(targets) = &mut delegated_role.targets {
+                    targets.signed.clear_targets();
+                }
+            }
+        }
+    }
+
+    /// Finds a targets by its `name`
+    pub fn targets_by_name(&mut self, name: &str) -> Result<&mut Self> {
```
zmrow: Since this method specifically *only* looks for delegated roles, let's mention that in the method name and docstring

Perhaps something like `get_delegated_targets()` or `delegated_targets_with_name()` or something like that?

tough/src/schema/mod.rs
```diff
@@ -528,6 +589,170 @@ impl Targets {
...
+                }
+            }
+        }
+        Err(Error::RoleNotFound {
+            name: name.to_string(),
+        })
+    }
+
+    /// Finds a targets role and verifies that it has access to `path`
+    pub fn targets_by_name_verify_path(&mut self, name: &str, path: &str) -> Result<&mut Self> {
```
zmrow: Same here - it seems like we're only looking for delegated targets. Let's mention that in the method name and docstring.  I don't love that the method name will be long, but it's best to be clear.

tough/src/schema/mod.rs
```diff
@@ -528,6 +589,170 @@ impl Targets {
...
+                }
+            }
+        }
+        Err(Error::RoleNotFound {
+            name: name.to_string(),
+        })
+    }
+
+    /// Returns a result with the `Signed<Targets>` called `name`
+    pub fn signed_targets_by_name(&self, name: &str) -> Result<&Signed<Self>> {
```
zmrow: Same here.

tough/src/schema/mod.rs
```diff
@@ -528,6 +589,170 @@ impl Targets {
...
+        self.targets.insert(name.to_string(), target);
+    }
+
+    /// Remove a target from targets
+    pub fn remove_target(&mut self, name: &str) -> Option<Target> {
+        self.targets.remove(name)
+    }
+
+    /// Returns a vec of all rolenames
+    pub fn get_roles_str(&self) -> Vec<&String> {
```
zmrow: Let's eliminate the "str" from this method name.  Also since this will only return delegated roles' names, perhaps we name it `delegated_role_names()` or something?

tough/src/schema/mod.rs
```diff
@@ -528,6 +589,170 @@ impl Targets {
...
+                    roles.append(&mut targets.signed.get_roles_str())
+                }
+            }
+        }
+
+        roles
+    }
+
+    /// Returns a result with the `DelegatedRole` called `name`
+    pub fn get_delegated_role_by_name(&mut self, name: &str) -> Result<&mut DelegatedRole> {
```
zmrow: Could this just be called `get_delegated_role()` or even just `delegated_role()`? 

Also, how does this overlap with `targets_by_name()`? I'm not sure when I would use this method instead. It might be best to only have a few methods for finding and returning delegated targets.

tough/src/schema/mod.rs
```diff
@@ -528,6 +589,170 @@ impl Targets {
...
+                    }
+                }
+            }
+        }
+        Err(error::Error::RoleNotFound {
+            name: name.to_string(),
+        })
+    }
+
+    /// returns a vec of all targets roles delegated by this role
```
zmrow: Nit: capitalize the beginning of the docstring

tough/src/schema/mod.rs
```diff
@@ -528,6 +589,170 @@ impl Targets {
...
+                if let Some(targets) = &role.targets {
+                    delegated_targets.push(targets.clone().delegated_targets(&role.name));
+                    delegated_targets.extend(targets.signed.signed_delegated_targets());
+                }
+            }
+        }
+        delegated_targets
+    }
+
+    /// link all current targets to `new_targets` metadata, returns a list of `new_targets` not included in the original targets
```
zmrow: Nit: capitalize here too.

tough/src/schema/mod.rs
```diff
@@ -528,6 +589,170 @@ impl Targets {
...
+    pub fn update_targets(&self, new_targets: &mut Signed<Targets>) -> Vec<String> {
```
zmrow, line 732: Could we add some additional documentation on why and when I would use this method?

tough/src/schema/mod.rs
```diff
@@ -528,6 +589,170 @@ impl Targets {
...
+        let mut needed_roles = Vec::new();
+        // Copy existing targets into proper places of new_targets
+        if let Some(delegations) = &mut new_targets.signed.delegations {
```
zmrow, line 735: Let's add some additional comments here. I'm having a hard time understanding what's happening.

tough/src/schema/mod.rs
```diff
@@ -793,6 +1087,69 @@ impl Delegations {
...
+                return Err(Error::UnmatchedPath {
+                    child: path.to_string(),
+                });
+            }
+        }
+        Ok(())
+    }
+
+    /// Creates a `Signed<DelegatedTargets>` for the `DelegatedRole`
+    pub fn signed_delegated_targets(self) -> Option<Signed<DelegatedTargets>> {
```
zmrow: Since this method doesn't sign anything, its name isn't quite right. It seems like more of a convenience for creating a `Signed<>` object?

tough/src/schema/verify.rs
```diff
@@ -49,6 +49,44 @@ impl Root {
     }
 }
 
+impl Delegations {
+    /// verifies that roles matches contain valid keys
```
zmrow: Nit: capitalize start of docstring

tough/src/schema/error.rs
```diff
@@ -113,6 +113,10 @@ pub enum Error {
 
     #[snafu(display("Targets doesn't contain delegations field"))]
     NoDelegations,
+
+    /// No role with name
```
zmrow: No comment needed here

tough/src/error.rs
```diff
@@ -342,6 +343,14 @@ pub enum Error {
         backtrace: Backtrace,
     },
 
+    /// A role metadata file could not be verified.
```
zmrow: No comments needed here - the error's message should be enough.

tough/src/error.rs
```diff
@@ -458,7 +467,16 @@ pub enum Error {
     TargetNotFound { target_url: String },
 
     #[snafu(display("Delegated role not found: {}", name))]
-    DelegateNotFound {
+    DelegateNotFound { name: String },
```
zmrow, line 469: This has the same message as `DelegateMissing`. Do both variants need to exist?

tough/src/error.rs
```diff
@@ -478,6 +496,62 @@ pub enum Error {
...
     RoleNotInMeta { name: String },
+
+    #[snafu(display("The key for {} was not included", role))]
+    KeyNotFound {
+        role: String,
+        source: schema::Error,
+        backtrace: Backtrace,
+    },
+
+    #[snafu(display("No keys were found for {}", role))]
```
zmrow: Nit: `...found for role '{}'`

tough/src/error.rs
```diff
@@ -478,6 +496,62 @@ pub enum Error {
...
+    /// Invalid path permissions
+    #[snafu(display("Invalid path permission of {} : {:?}", name, paths))]
+    InvalidPathPermission {
+        name: String,
+        paths: Vec<String>,
+        source: schema::Error,
+    },
+
+    /// Overflow error with checked add
+    #[snafu(display("Overflow of checked add"))]
```
zmrow: I'm not sure what this error message means. Could we clarify?

tough/src/error.rs
```diff
@@ -478,6 +496,62 @@ pub enum Error {
...
+    Overflow,
+
+    /// SignedDelegatedTargets has more than 1 signed targets
+    #[snafu(display("Not exactly 1 role was created: {}", count))]
```
zmrow: Let's clarify this message as well. Are we trying to say that more than 1 delegated targets roles was created?

tough/src/editor/targets.rs
```diff
@@ -0,0 +1,536 @@
...
+use snafu::{OptionExt, ResultExt};
+use std::borrow::Cow;
+use std::collections::HashMap;
+use std::num::NonZeroU64;
+use std::path::Path;
+use url::Url;
+
+const SPEC_VERSION: &str = "1.0.0";
+
+/// `TargetsEditor` contains the various bits of data needed to construct
```
zmrow, line 34: I think a bit more documentation is needed here.  I believe the only time a user would need to use this directly is if they were dealing with delegations or they are a delegatee. Correct? Otherwise most users can ignore this.

You've got a great start by documenting what to do, let's add some words explaining why they would use this and most importantly making it clear when *not* to use it. :)

zmrow: We should probably just preface the whole thing with something like "If you're not working with delegations, use the RepositoryEditor" 😃

tough/src/editor/targets.rs
```diff
@@ -0,0 +1,536 @@
...
+                }
+            }
+        } else {
+            delegations.keys.remove(keyid);
+        }
+        Ok(self)
+    }
+
+    /// Delegates `paths` to `targets` adding a `DelegatedRole` to `new_roles`
+    pub fn delegate_role(
```
zmrow, line 311: The comment here is a little confusing. Could we just say `Adds a DelegatedRole `?

tough/src/editor/targets.rs
```diff
@@ -0,0 +1,536 @@
...
+                    signed: targets.signed.targets,
+                    signatures: targets.signatures,
+                }),
+            });
+        Ok(self)
+    }
+
+    /// Removes a role from delegations if `recursive` is `false`
+    /// requires the role to be an immediate delegated role
+    /// if `true` removes whichever role eventually delegates 'role'
```
zmrow: There's a bit of run-on sentence going on here. Could we clarify the `recursive` bit? If recursive is "false", it means that we will only remove roles that this `targets` has delegated? We will not remove roles further down the chain right?

tough/src/editor/targets.rs
```diff
@@ -0,0 +1,536 @@
...
+                } else {
+                    true
+                }
+            });
+        }
+        Ok(self)
+    }
+
+    /// Adds a role to `new_roles`
+    pub fn add_role(
```
zmrow, line 360: Could we clarify the difference between this method and `delegate_role()` to make it clear which method a user should use?

tough/src/editor/targets.rs
```diff
@@ -0,0 +1,536 @@
...
+                expires,
+                targets,
+                _extra,
+                delegations,
+            },
+        })
+    }
+
+    /// Creates a `Signed<DelegatedTargets>` for this role using the provided keys
+    pub fn create_signed(&self, keys: &[Box<dyn KeySource>]) -> Result<Signed<DelegatedTargets>> {
```
zmrow, line 460: Let's clarify the difference between this method and `sign()` below. When would I use them?

tough/src/editor/signed.rs
```diff
@@ -86,7 +87,13 @@ where
             sig: sig.into(),
         });
 
-        // Serialize the newly signed role, and calculate its length and
+        SignedRole::from_signed(role)
+    }
+
+    /// Creates a `SignedRole<Role>` from a `Signed<Role>`.
+    /// This is used to create signed roles for any signed metadata
+    pub fn from_signed(role: Signed<T>) -> Result<SignedRole<T>> {
```
zmrow: Should this be public?  It's a bit confusing unless you knwo the difference between `SignedRole<>` and `Signed<>`

tough/src/editor/signed.rs
```diff
@@ -132,31 +139,14 @@ where
...
-                if consistent_snapshot {
-                    format!("{}.snapshot.json", self.signed.signed.version())
-                } else {
-                    "snapshot.json".to_string()
-                }
-            }
-            RoleType::Timestamp => "timestamp.json".to_string(),
-            RoleType::Root => format!("{}.root.json", self.signed.signed.version()),
-        };
+        let filename = self.signed.signed.filename(consistent_snapshot);
```
zmrow, line 153: Nice!

tough/src/editor/signed.rs
```diff
@@ -494,3 +504,187 @@ impl SignedRepository {
...
+    fn consistent_snapshot(&self) -> bool {
+        self.consistent_snapshot
+    }
+}
+
+/// `PathExists` allows the user of our copy/link functions to specify what happens when the target
+/// is being written to a shared targets directory and the file already exists from another repo.
+#[derive(Debug, Deserialize, Clone, Copy)]
+#[serde(rename_all = "kebab-case")]
+pub enum PathExists {
```
zmrow: Could we move all of this back to where it was? It pre-empts all of this target moving/copying code very nicely.

tough/src/editor/signed.rs
```diff
@@ -494,3 +504,187 @@ impl SignedRepository {
...
+enum TargetPath {
+    /// No existing file found, we can create a new one at this path.
+    New { path: PathBuf },
+    /// Existing regular file found at this path.
+    File { path: PathBuf },
+    /// Existing symlink found at this path.
+    Symlink { path: PathBuf },
+}
+
+trait TargetsWalker {
```
zmrow, line 551: It looks like this unifies the logic for copying/linking targets for `SignedRepository` and `SignedDelegatedTargets`. Is that correct?  The logic is mostly the same?

Could we add a few lines of docstring to the trait and allude to the "default" implementations of `walk_targets()` and `target_path()`?

zmrow (commented):
Nice work! This will be a great addition for folks dealing with delegated targets. 

I mention this in the review, but let's make sure it's *very* clear that `TargetsEditor` is primarily meant for those dealing with delegations.  We should also make it very clear that `RepositoryEditor`'s interface is the same and is the primary method for updating/changing a repo when delegations aren't involved.

tough/src/editor/targets.rs
tjkirch: Can we call it `DelegationEditor` instead, to avoid that whole problem?

tough/src/schema/mod.rs
ecpullen: I agree. Standard role is much clearer.

tough/src/editor/targets.rs
zmrow: I believe Repository Editor is going to use this under the hood to handle `Targets`, so it `DelegationsEditor` would be confusing in that case. :)

tough/src/error.rs
ecpullen: This error is thrown whenever a checked add fails. Maybe `Checked add failed due to overflow`?

tough/src/error.rs
ecpullen: We are trying to say that 0 or more than 1 role was created. Maybe `Exactly 1 role was required, but {} were created`?

tough/src/error.rs
zmrow: I'm not sure what all those words mean in concert: Checked add to what? Overflow of what?

tough/src/error.rs
zmrow: That sounds better!

tough/src/editor/keys.rs
ecpullen: My original thoughts were that it is possible that a user will want to create a `KeyHolder`, but there is no reason for a user to use the methods from the impl.

tough/src/error.rs
ecpullen: This is an error for a different PR. It has been removed.

tough/src/error.rs
ecpullen: They have different sources for their errors. I'm not sure if 1 error can have more than 1 source.

tough/src/editor/signed.rs
```diff
@@ -3,20 +3,23 @@
 
 //! Provides the `SignedRepository` object which represents the output of `RepositoryEditor` after
 //! signing, ready to be written to disk.
+//! //! Provides the `SignedDelegatedTargets` object which represents the output of `TargetsEditor` after
```
webern, line 6: Small typo
```suggestion
//! Provides the `SignedDelegatedTargets` object which represents the output of `TargetsEditor` after
```

tough/src/editor/keys.rs
zmrow: Let's remove this line since we don't provide the `KeyHolder` object here.

tough/src/editor/targets.rs
zmrow: Based on the comment, is the only difference between the two methods that this one fetches the metadata from a URL? If so, let's make the methods allude to that fact.

Also - let's add to the docstring a few words as to why and when I would use either of these

tough/src/editor/targets.rs
zmrow: When would I want to use `Signed<DelegatedTargets>` vs. `SignedDelegatedTargets`?

zmrow (commented):
I think I resolved my comments that were fixed and also had a few more below.

Overall I think the docstrings on the `TargetsEditor` and `signed.rs` could use a peek. Lets make sure a user it's clear when a user should use the method.

zmrow (approved):
🎂 

Nice work!  This looks good to me!

From what I understand, some of the methods in `editor/signed.rs` are being simplified a bit in the PR for the `RepositoryEditor`. If it makes more sense in that context, we'll keep them over there.

Onward!

tough/src/schema/mod.rs
```diff
@@ -528,6 +590,163 @@ impl Targets {
...
+
+    /// Returns a result with the `Signed<Targets>` called `name`
+    pub fn signed_delegated_targets_with_name(&self, name: &str) -> Result<&Signed<Self>> {
+        if let Some(delegations) = &self.delegations {
+            for role in &delegations.roles {
+                if let Some(targets) = &role.targets {
+                    if role.name == name {
+                        return Ok(&targets);
+                    } else if let Ok(role) = targets.signed.signed_delegated_targets_with_name(name)
+                    {
```
webern, line 658: did cargo fmt produce this? looks weird

tough/src/editor/signed.rs
```diff
@@ -494,3 +523,166 @@ impl SignedRepository {
...
+
+    /// Walks a given directory and calls the provided function with every file found.
+    /// The function is given the file path, the output directory where the user expects
+    /// it to go, and optionally a desired filename.
+    fn walk_targets<F>(
+        &self,
+        indir: &Path,
+        outdir: &Path,
+        f: F,
+        replace_behavior: PathExists,
```
webern, line 565: Maybe a little strange to accept replace_behavior only to have it passed to the function that was also passed. The repace_behavior could be captured or curried into `f`.

Not a blocker, just an observation.

webern (approved):
🦜

tough/src/schema/mod.rs
webern: It must have since it passed CI!

PR #202, opened 2020-08-03
Repoeditor

ecpullen:
**This is the second part of a set of 4 PRs to finish implementing delegated targets.**
Follows #200 

*Issue #, if available:*
Related to #5

*Description of changes:*
* Integrates `TargetsEditor` with `RepositoryEditor`
* `RepositoryEditor` uses a modal design to edit `Targets`. `TargetsEditor` is used to perform all actions on a specified `Targets`. To change the `Targets` being used call `change_delegated_targets()` to create a new `TargetsEditor` for the specified role. To sign a `Targets` role from the `TargetsEditor` use `sign_targets_editor()`. This will clear out the targets editor and insert the newly signed targets in `signed_targets`.
* To update an existing targets from a metadata file use `update_delegated_targets()`.
* To add a new role from metadata to the `Targets` in `TargetsEditor` use `add_role()`.
* Adds 2 new types of accepted keys ED25519 and ECDSA to sign.rs

*Testing:*
* Creates a repo and adds delegated targets using `RepositoryEditor`
* Creates a repo and creates targets using `TargetsEditor`
* Adds `Targets` using `TargetsEditor`
* Adds `Targets` using `RepositoryEditor`
* Updates `Targets` using `RepositoryEditor`


tough/src/editor/targets.rs
```diff
@@ -452,22 +452,62 @@ impl<'a, T: Transport> TargetsEditor<'a, T> {
         })
     }
 
+    /// Creates a `KeyHolder` to sign `name` with the signing keys provided
```
zmrow: It might make more sense to say "Creates a `KeyHolder` to sign this `Targets` role..."

tough/src/editor/targets.rs
```diff
@@ -452,22 +452,62 @@ impl<'a, T: Transport> TargetsEditor<'a, T> {
...
+    fn create_key_holder(&self, keys: &[Box<dyn KeySource>]) -> Result<KeyHolder> {
+        // There isn't a KeyHolder, so create one based on the provided keys
+        let mut temp_delegations = Delegations::new();
```
zmrow: This isn't really temporary right? It ends up being held in the `KeyHolder::Delegations`

tough/src/editor/mod.rs
```diff
@@ -44,28 +51,40 @@ const SPEC_VERSION: &str = "1.0.0";
 /// `sign()` method, which takes a given set of signing keys, builds each of
 /// the roles using the data provided, and signs the roles. This results in a
 /// `SignedRepository` which can be used to write the repo to disk.
+///
+/// `RepositoryEditor` uses a modal design to edit `Targets`. `TargetsEditor`
+/// is used to perform all actions on a specified `Targets`. To change the
```
zmrow, line 57: Let's clarify this a little bit.  Let's first make it clear that unless you are dealing with delegations, you don't need to worry about which `Targets` object you are working on (since it will just be the `Targets` role)

It would be good to add that if you are dealing with delegations, you can perform actions on a specified delegated role by calling `change_delegated_targets()`.

tough/src/editor/mod.rs
```diff
@@ -44,28 +51,40 @@ const SPEC_VERSION: &str = "1.0.0";
...
+/// `Targets` being used call `change_delegated_targets()` to create a new `TargetsEditor`
+/// for the specified role. To sign a `Targets` role from the `TargetsEditor` use `sign_targets_editor()`.
+/// This will clear out the targets editor and insert the newly signed targets in `signed_targets`.
+///
+/// To update an existing targets from a metadata file use `update_delegated_targets()`.
+///
+/// To add a new role from metadata to the `Targets` in `TargetsEditor` use `add_role()`.
 #[derive(Debug)]
-pub struct RepositoryEditor {
+pub struct RepositoryEditor<'a, T: Transport> {
```
zmrow, line 66: Hmmmm, I'm wondering if we can eliminate the need to specify a transport if I'm just using the Editor to start a new repo. It's a bit unwieldy and confusing, especially since `targets_editor` is an `Option<>`.

zmrow: For example, if I'm calling `RepositoryEditor::new()`, specifying a transport is a bit confusing, especially since I may need to not know about it at all.

@iliana may have some valuable guidance here.

tough/src/editor/mod.rs
```diff
@@ -44,28 +51,40 @@ const SPEC_VERSION: &str = "1.0.0";
...
     snapshot_extra: Option<HashMap<String, Value>>,
 
     timestamp_version: Option<NonZeroU64>,
     timestamp_expires: Option<DateTime<Utc>>,
     timestamp_extra: Option<HashMap<String, Value>>,
+
+    targets_editor: Option<TargetsEditor<'a, T>>,
+
+    /// The signed top level targets, will be None if no top level targets have been signed
+    signed_targets: Option<Signed<Targets>>,
```
zmrow, line 80: When would `signed_targets` be `Some()`?

tough/src/editor/mod.rs
```diff
@@ -125,15 +146,30 @@ impl RepositoryEditor {
...
+        let signed_delegated_targets = if delegated_targets.is_empty() {
+            None
+        } else {
+            let mut roles = Vec::new();
+            for role in delegated_targets {
+                roles.push(SignedRole::from_signed(role)?)
+            }
+            Some(SignedDelegatedTargets {
+                roles,
+                consistent_snapshot: self.signed_root.signed.signed.consistent_snapshot,
```
zmrow: Let's add some comments here describing why the delegated targets process is slightly different.

tough/src/editor/mod.rs
```diff
@@ -193,12 +232,24 @@ impl RepositoryEditor {
         Ok(self)
     }
 
+    /// Returns a mutable reference to the targets editor if it exists, or an error if it doesn't
```
zmrow: No need to mention the error - we can infer that from the `Result<>` :)

tough/src/editor/mod.rs
```diff
@@ -193,12 +232,24 @@ impl RepositoryEditor {
...
+    fn targets_editor_mut(&mut self) -> Result<&mut TargetsEditor<'a, T>> {
+        self.targets_editor
+            .as_mut()
```
zmrow, line 256: Is this necessary? Don't we own `targets_editor`?

tough/src/editor/mod.rs
```diff
@@ -208,6 +259,31 @@ impl RepositoryEditor {
     /// of targets to add, and require advanced performance, you may want to
     /// construct `Target`s directly in parallel and use `add_target()`.
     pub fn add_target_path<P>(&mut self, target_path: P) -> Result<&mut Self>
+    where
+        P: AsRef<Path>,
+    {
+        let (target_name, target) = RepositoryEditor::<T>::build_target(target_path)?;
```
zmrow: I think I prefer this function as it was before. I'm not sure the extra `build_target` function adds much. Or maybe I'm missing something?  Do additional places use this that I missed?

zmrow: Also - this is a good example of why we should get rid of the `<T>` ... it's no fun to always need to specify that every time we call methods

tough/src/editor/mod.rs
```diff
@@ -225,30 +301,56 @@ impl RepositoryEditor {
...
 
-    /// Remove all targets from this repo
-    pub fn clear_targets(&mut self) -> &mut Self {
-        self.existing_targets
-            .get_or_insert_with(HashMap::new)
-            .clear();
-        self.new_targets.get_or_insert_with(HashMap::new).clear();
-        self
+    #[allow(clippy::too_many_arguments)]
+    /// Delegate target with name.
```
zmrow: Let's expand this docstring. In what delegated targets situation should I use this method?

tough/src/editor/mod.rs
```diff
@@ -287,42 +389,217 @@ impl RepositoryEditor {
         self
     }
 
-    // =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
+    /// Takes the current Targets from `targets_editor` and inserts the role to its proper place in `signed_targets`
+    /// Sets `targets_editor` to None
+    /// Must be called before `change_delegated_targets()`
```
zmrow, line 403: Is there any security we can provide in `change_delegated_targets()`?  Relying on docstrings to deter this behavior seems a little scary!

tough/src/editor/mod.rs
```diff
@@ -287,42 +389,217 @@ impl RepositoryEditor {
...
-    /// Build the `Targets` struct
-    fn build_targets(&self) -> Result<Targets> {
-        let version = self.targets_version.context(error::Missing {
-            field: "targets version",
-        })?;
-        let expires = self.targets_expires.context(error::Missing {
-            field: "targets expiration",
-        })?;
+    /// Changes the targets refered to in `targets_editor` to role
+    /// Throws error if the `targets_editor` was not cleared using `sign_targets_editor()`
```
zmrow, line 425: Oh good - glad this is the behavior we chose.

tough/src/editor/mod.rs
```diff
@@ -287,42 +389,217 @@ impl RepositoryEditor {
...
 
-    /// Build the `Targets` struct
-    fn build_targets(&self) -> Result<Targets> {
-        let version = self.targets_version.context(error::Missing {
-            field: "targets version",
-        })?;
-        let expires = self.targets_expires.context(error::Missing {
-            field: "targets expiration",
-        })?;
+    /// Changes the targets refered to in `targets_editor` to role
```
zmrow, line 423: Let's add a little bit more to this docstring in case a user didn't read the docstring on the RepositoryEditor.  Let's further explain what's going to happen here and in what situations would a user want to use this.

tough/src/editor/mod.rs
```diff
@@ -287,42 +389,217 @@ impl RepositoryEditor {
...
-        // the most common use case, it's possible this is what a user wants.
-        // If it's important to have a non-empty targets, the object can be
-        // inspected by the calling code.
-        let mut targets: HashMap<String, Target> = HashMap::new();
-        if let Some(ref existing_targets) = self.existing_targets {
-            targets.extend(existing_targets.clone());
+    #[allow(clippy::too_many_lines)]
+    /// Updates the metadata for `name`
+    /// Clears the current `targets_editor`
+    pub fn update_delegated_targets(
```
zmrow, line 467: Could we add a few more lines here? When would a user want this?

tough/src/editor/mod.rs
```diff
@@ -287,42 +389,217 @@ impl RepositoryEditor {
...
-        let _extra = self.targets_extra.clone().unwrap_or_else(HashMap::new);
-        Ok(Targets {
-            spec_version: SPEC_VERSION.to_string(),
-            version,
-            expires,
-            targets,
-            _extra,
-            delegations: None,
-        })
+    /// Adds a role to the targets currently in `targets_editor`
```
zmrow, line 588: Same here. When would I use this method vs. `delegate_role`?

zmrow (commented):
Publishing what I have for now - more to come!

tough/src/editor/mod.rs
ecpullen: It is only `None` if the repo was created using `Repository::new()`.

tough/src/editor/mod.rs
ecpullen: The doc strings that you mention in other comments are warnings about this error.

tough/src/editor/mod.rs
ecpullen: I think we need that to convert `&mut Option<TargetsEditor>` to `Option<&mut TargetsEditor>` so that we can unwrap without moving.

tough/src/editor/mod.rs
tjkirch: Can we own the transport instead of using a reference so we don't need the lifetime?  Can we Box it instead of the generic so we don't have to add the generic here either?  I agree that it's confusing...

ecpullen: I'm not sure what we can do. I know that Repository is set up this way.

tough/src/editor/mod.rs
ecpullen: This doc string is warning that calling `change_delegated_targets()` without calling `signed_targets_editor()` will throw an error.

tough/src/editor/mod.rs
webern: Repository has a type parameter for Transport.

I think epullen's PR's should not address this, and should continue the unwieldy pattern of `'a, T`. Once the code churn slows, I think someone (I'd be happy to work on it) should come through and change the definitions so that Repository completely hides its Transport type (e.g. it holds/takes a `Box<dyn Transport>`).

tough/src/editor/mod.rs
webern: Could say something like "Will error if called before change_delegated_targets()"

tough/src/editor/signed.rs
```diff
@@ -3,20 +3,23 @@
 
 //! Provides the `SignedRepository` object which represents the output of `RepositoryEditor` after
 //! signing, ready to be written to disk.
+//! //! Provides the `SignedDelegatedTargets` object which represents the output of `TargetsEditor` after
```
webern: small typo
```suggestion
//! Provides the `SignedDelegatedTargets` object which represents the output of `TargetsEditor` after
```

tough/src/editor/targets.rs
```diff
@@ -0,0 +1,556 @@
...
+            .insert(name.to_string(), target);
+        self
+    }
+
+    /// Add a target to the repository using its path
+    ///
+    /// Note: This function builds a `Target` synchronously;
+    /// no multithreading or parallelism is used. If you have a large number
+    /// of targets to add, and require advanced performance, you may want to
+    /// construct `Target`s directly in parallel and use `add_target()`.
```
webern: Seems like I could also use this function to roll my own parallelization. Seems like this warning is only relevant for a function that operates on a batch, or directory walk.

tough/src/editor/targets.rs
```diff
@@ -0,0 +1,556 @@
...
+            .context(error::PathUtf8 { path: target_path })?
+            .to_owned();
+
+        self.add_target(&target_name, target);
+        Ok(self)
+    }
+
+    /// Add a list of target paths to the targets
+    ///
+    /// See the note on `add_target_path()` regarding performance.
```
webern: Seems like the performance warning should only go on this function.

tough/src/editor/targets.rs
```diff
@@ -0,0 +1,556 @@
+// Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: MIT OR Apache-2.0
+#![allow(clippy::used_underscore_binding)] // #20
+
+//! Provides a `TargetsEditor` object for building and editing targets roles.
```
webern: How hard would it be to put a usage example in this docstring? Rust will run even run the examples. I don't know if there's way too much setup code required though. https://doc.rust-lang.org/stable/rust-by-example/meta/doc.html

webern (approved):

zmrow:
I resolved my comments.

I think the only outstanding issue I have is the [need to specify a transport](https://github.com/awslabs/tough/pull/202/files#r465141828) whenever you use the `RepositoryEditor`. I was hoping that could be solved before we ship this PR. Perhaps @iliana has input there?

iliana:
> I think the only outstanding issue I have is the [need to specify a transport](https://github.com/awslabs/tough/pull/202/files#r465141828) whenever you use the `RepositoryEditor`. I was hoping that could be solved before we ship this PR.

I don't think it needs to be solved before merging — it's not great ergonomically but we can fix it in a later release.

zmrow:
> > I think the only outstanding issue I have is the [need to specify a transport](https://github.com/awslabs/tough/pull/202/files#r465141828) whenever you use the `RepositoryEditor`. I was hoping that could be solved before we ship this PR.
> 
> I don't think it needs to be solved before merging — it's not great ergonomically but we can fix it in a later release.

Sounds good - I'll approve now.

zmrow (approved):
💸

webern:
> > I think the only outstanding issue I have is the [need to specify a transport](https://github.com/awslabs/tough/pull/202/files#r465141828) whenever you use the `RepositoryEditor`. I was hoping that could be solved before we ship this PR.
> 
> I don't think it needs to be solved before merging — it's not great ergonomically but we can fix it in a later release.

I agree that we should change this in a later release. I think it is best handled in a PR/commit that is dedicated to this purpose.

iliana (approved):

ecpullen:
Force push was a rebase onto develop.

Issue #209, opened 2020-08-07
tuftool: Add support to sign root.json with multiple keys in single command


sanu11:
root.json can only be signed with one key using the `tuftool root sign root.json <keysource>` command. 
Add support to include multiple keys in the command eg: `tuftool root sign root.json <keysource> <keysource>` or `tuftool root sign root.json -k <keysource> -k <keysource>`

PR #210, opened 2020-08-07
tuftool: Added support for signing root.json with multiple keys and  cross-signing support for root.json's root sign command

sanu11:
*Issue #, if available:*
#209 

*Description of changes:*
Sign command takes a vector of key sources now, instead of single key source.
The root sign command can be executed as : `tuftool root sign root.json <keysource> <keysource> `


tuftool/src/root.rs
```diff
@@ -94,7 +94,7 @@ pub(crate) enum Command {
         /// Path to root.json
         path: PathBuf,
         #[structopt(parse(try_from_str = parse_key_source))]
-        key_source: Box<dyn KeySource>,
+        key_source: Vec<Box<dyn KeySource>>,
```
zmrow: If we're going to accept multiple `KeySource`s, it probably makes sense to add a short flag to this argument. (Generally good CLI practice as well)

So when you call it it would look like:
`tuftool .... -k keysource1 -k keysource2`

I believe you need to add `short = "k"` to the `structopt` annotations.

sanu11: Yup. Added that. Earlier I was confused if I should add it or no, since there are no other flags in this command.  But yeah, if there are multiple keys, it makes sense to add the flag. And also since now I have added another flag --cross-flag so its better to have --key to not confuse with this one : )

tuftool/src/root.rs
```diff
@@ -275,16 +283,35 @@ impl Command {
...
+            None => SignedRole::new(
+                root.signed.clone(),
+                &root.signed,
+                key_source,
+                &SystemRandom::new(),
+            )
+            .context(error::SignRoot { path })?,
+            Some(..) => {
+                // This will get the keys from cross-sign root.json for validation
+                let cross_sign_root = cross_sign.as_ref().unwrap();
```
ecpullen: I think a more rust way of doing this would be 
`Some(cross_sign_root) => {`
It also prevents using `unwrap()`.

sanu11: Yup. changed it. Thanks

tuftool/src/root.rs
```diff
@@ -93,8 +93,12 @@ pub(crate) enum Command {
     Sign {
         /// Path to root.json
         path: PathBuf,
-        #[structopt(parse(try_from_str = parse_key_source))]
-        key_source: Box<dyn KeySource>,
+        ///Key source(s) to sign the file with
+        #[structopt(short = "k", long = "key",parse(try_from_str = parse_key_source))]
+        key_source: Vec<Box<dyn KeySource>>,
+        ///Optional - Path of older root.json that contains the key-id
```
webern, line 99: Should this be plural? (key-ids)

tuftool/src/root.rs
```diff
@@ -275,16 +283,34 @@ impl Command {
...
+                let old_root: Signed<Root> = load_file(&cross_sign_root)?;
+                SignedRole::new(
+                    root.signed,
+                    &KeyHolder::Root(old_root.signed),
+                    key_source,
+                    &SystemRandom::new(),
+                )
+                .context(error::SignRoot { path })?
+            }
+        };
```
webern: Something like this (pseudo-code) might be a little DRY-er.

```suggestion
        let loaded_root = match cross_sign {
            None => root,
            Some(cross_sign_root) => load_file(&cross_sign_root)?
        };
        let signed_root = SignedRole::new(
            root.signed,
            &KeyHolder::Root(loaded_root),
            key_source,
            &SystemRandom::new(),
        )
        .context(error::SignRoot { path })?
```

webern (commented):
Does cross-signing *add* signatures or *replace* signatures? Don't we need to have the union of signatures from keys found in root1 and root2 in root2?

sanu11 (commented):
> Does cross-signing _add_ signatures or _replace_ signatures? Don't we need to have the union of signatures from keys found in root1 and root2 in root2?

Yeah, so for that there is this issue that I am working on. https://github.com/awslabs/tough/issues/214 
So, you can run the command twice. Once you run the tufool command to sign with keys in root1 and then run it again with --cross-sign to sign it with keys in root2

tuftool/src/root.rs
sanu11: oh yes, it should be plural.

tuftool/src/root.rs
sanu11: Yup changed it in new commit.

tough/src/editor/signed.rs
```diff
@@ -85,12 +85,13 @@ where
                 sig: sig.into(),
             });
         }
-        if role_keys.threshold.get() > role.signatures.len() as u64 {
+
+        // since for root the check depends on cross-sign
+        if T::TYPE != RoleType::Root && role_keys.threshold.get() > role.signatures.len() as u64 {
```
sanu11, line 90: I need suggestion here. So, the reason I added this RoleType check is:
1) The statement inside the if condition validates the threshold with role_keys, which would be validating number of signatures in current root.json with the threshold value in cross-sign root.json ( when --cross-sign flag is not None) Ideally, it should validate with itself. 
2) There is already a check for number of signatures in tuftool's sign command. And this function is only called from tuftool's sign command for ROOT role. I am not sure, if  this is the correct way though, because if in future, this function is being called for ROOT role from somewhere else, there would be no check for threshold number of signatures.
3) This may be rare usecase. But say, if we set threshold to 1 and add a signature to root.json and then we change the threshold to 2 and run the sign command to append the second signature, it would give error without this condition applied. Reason being, the old signature is appended later after the threshold check and it would see 1 signature and threshold of 2. (That's again because of the approach I took for appending signatures. )

tough/src/editor/signed.rs
```diff
@@ -155,6 +155,21 @@ where
         let path = outdir.join(filename);
         std::fs::write(&path, &self.buffer).context(error::FileWrite { path })
     }
+
+    /// Append the old signatures for root role
+    pub fn add_old_signatures(mut self, old_signatures: Vec<Signature>) -> Result<Self>{
```
sanu11: Would this be right way to append signatures ? So, this is called after the `SignedRole::new` function  which signs the file with current keys.  I haven't modified the existing `SignedRole::new` function to accomodate signature as the parameter because: 
The `SignedRole::new` function is used for signing other files as well and since they are generated later they would not have existing signature in them. 

Another approach can be to add one more function, say `sign` which will take `Signed<T>` as the parameter. It will be called by `SignedRole::new` with `role = Signed {
            signed: role,
            signatures: Vec::new(),
        };` as one of the parameter for other metadata except root. And for root, it will directly call this`sign`function instead of `SignedRole::new` with `Signed<Root>` as parameter which will have existing signatures in it

tough/src/editor/signed.rs
```diff
@@ -155,6 +156,26 @@ where
         let path = outdir.join(filename);
         std::fs::write(&path, &self.buffer).context(error::FileWrite { path })
     }
+
+    /// Append the old signatures for root role
+    pub fn add_old_signatures(mut self, old_signatures: Vec<Signature>) -> Result<Self> {
```
zmrow, line 161: Ideally we'd like to restrict this to the root role right?

sanu11: Yes, as far as I know. Because the rest of the metadata is not manually signed by us. 
For root role, we need this because we would run the sign command (using tuftool) twice, once to sign with the keys from 1.root.json and once to sign with the keys from 2.root.json (This is the usecase for key rotation)

zmrow (approved):

tough/src/editor/signed.rs
iliana: This modification seems right but it's still hard to wrap my brain around it 😅

iliana (approved):

Issue #226, opened 2020-09-17
tuftool: panic when parsing malformed key source


samuelkarp:
I tried to run `tuftool root add-key` with a malformed key source and saw it panic.  This is with a tuftool built from the current `HEAD` of the `develop` branch (https://github.com/awslabs/tough/commit/969bf59be72112a1e355797d01f246dfba9a753e).

```
$ tuftool root add-key root.json aws-kms://tuf-junk --role root
thread 'main' panicked at 'byte index 1 is out of bounds of ``', /home/samuelkarp/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/libcore/str/mod.rs:2052:47
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

<details>
<summary>With RUST_BACKTRACE=1</summary>

```
$ RUST_BACKTRACE=1 tuftool root add-key root.json aws-kms://tuf-junk --role root
thread 'main' panicked at 'byte index 1 is out of bounds of ``', /home/samuelkarp/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/libcore/str/mod.rs:2052:47
stack backtrace:
   0: backtrace::backtrace::libunwind::trace
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.46/src/backtrace/libunwind.rs:86
   1: backtrace::backtrace::trace_unsynchronized
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.46/src/backtrace/mod.rs:66
   2: std::sys_common::backtrace::_print_fmt
             at src/libstd/sys_common/backtrace.rs:78
   3: <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt
             at src/libstd/sys_common/backtrace.rs:59
   4: core::fmt::write
             at src/libcore/fmt/mod.rs:1076
   5: std::io::Write::write_fmt
             at src/libstd/io/mod.rs:1537
   6: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:62
   7: std::sys_common::backtrace::print
             at src/libstd/sys_common/backtrace.rs:49
   8: std::panicking::default_hook::{{closure}}
             at src/libstd/panicking.rs:198
   9: std::panicking::default_hook
             at src/libstd/panicking.rs:218
[...]

PR #227, opened 2020-09-17
Fixed empty string handling while parting AWS KMS key

srgothi92:
*Issue #, if available:*
#226

*Description of changes:*
Handled empty URL path before extracting first '/' from the path.

Testing Done:
Modified tutool integration test to use `aws-kms://tuf-junk` and see it fail with error `Unable to parse keypair: Failed to get public key for aws-kms://tuf-junk/ : profile not found` instead of panic.


iliana (approved):

zmrow (approved):
:ghost:

samuelkarp (approved):

PR #234, opened 2020-10-08
Update dependencies

zmrow:
*Issue #, if available:*
N/A

*Description of changes:*
* ran `cargo update`

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #235, opened 2020-10-08
Update code for clippy's new 'option-if-let-else' check

zmrow:
*Issue #, if available:*
N/A
Related to #234 

*Description of changes:*
Clippy appears to have a new check... it is very pedantic about using `Option::map_or()`

I made the changes in the spots where it made sense.  I opted to ignore the suggestion in 2 places where there were long `if/else` chains, as it made the code very hard to follow there.

*Testing done:*
Basically ran all the same checks as the GH actions runs.


tjkirch (approved):
They're getting too aggressive with these lints.

webern (approved):

PR #253, opened 2020-10-22
tough: remove lifetime annotation from settings

### LINKED ISSUE #249: tough: Change definition of Datastore to eliminate lifetime
Datastore is defined as:

```rust
#[derive(Debug, Clone)]
pub(crate) struct Datastore<'a>(Arc<RwLock<&'a Path>>);
```

This propagates a lifetime up to `Repository` where users have to deal with it. I assert that users would rather just give up ownership of that `Path` than have lifetime annotations on usages of `Repository`.

The proposal is to change the definition to

```rust
#[derive(Debug, Clone)]
pub(crate) struct Datastore(Arc<RwLock<Path>>);
```

The new function could be generic on `Into<Path>`, e.g.

```rust
    pub(crate) fn new<P: Into<Path>>(path: P) -> Self {
        Self(Arc::new(RwLock::new(path.into())))
    }
```

Together with #248, this would eliminate all type annotations from the `Repository` struct object.

webern:
*Issue #, if available:*

Closes #249 

*Description of changes:*

```
Datastore: change the inner type from &'a Path to either PathBuf or
TempDir. This supports a common use case in which the user wants a
TempDir and does not want to own the lifetime of the Path.

Settings: change the datastore field to be an Option<PathBuf> instead of
&'a Path. This allows the user to specify None if they want a TempDir
to be created for them automatically, but retains the ability for the
user to specify the location of an existing directory. In the latter
case, it is often more convenient for the user to move ownership of the
Path-like object than to provide a reference.

Settings: change the two URL fields to owned Strings instead of &'a str.
This gives the user the flexibility to move temporary strings into the
Settings struct's ownership instead of creating local variables that the
user might not otherwise need.
```
*Testing*

For this one I am trusting the compiler and unit tests.

In the test case changes, we can see the improvement to the interface. Fewer lines of code are needed in the test cases for declaring variables that were only required for the sake of passing a ref. I believe most/all use cases in Bottlerocket will similarly benefit from the reduction of local variables that aren't otherwise needed.

### DIFF
--- a/tuftool/tests/delegation_commands.rs
+++ b/tuftool/tests/delegation_commands.rs
@@ -8,6 +8,7 @@ use chrono::{Duration, Utc};
 use std::fs::File;
 use std::path::Path;
 use tempfile::TempDir;
+use test_utils::dir_url;
 use tough::{ExpirationEnforcement, Limits, Repository, Settings};
 
 fn create_repo<P: AsRef<Path>>(repo_dir: P) {
@@ -76,7 +77,7 @@ fn create_add_role_command() {
 
     // Set new expiration date for the new role
     let expiration = Utc::now().checked_add_signed(Duration::days(4)).unwrap();
-    let metadata_base_url = &test_utils::dir_url(repo_dir.path().join("metadata"));
+    let metadata_base_url = &dir_url(repo_dir.path().join("metadata"));
     let meta_out = TempDir::new().unwrap();
 
     // create role A
@@ -111,7 +112,7 @@ fn create_add_role_command() {
             "-o",
             new_repo_dir.path().to_str().unwrap(),
             "-i",
-            test_utils::dir_url(&meta_out.path().join("metadata")).as_str(),
+            dir_url(&meta_out.path().join("metadata")).as_str(),
             "-k",
             root_key.to_str().unwrap(),
             "--root",
@@ -140,16 +141,15 @@ fn create_add_role_command() {
         .success();
 
     // Load the updated repo
-    let temp_datastore = TempDir::new().unwrap();
-    let updated_metadata_base_url = &test_utils::dir_url(new_repo_dir.path().join("metadata"));
-    let updated_targets_base_url = &test_utils::dir_url(new_repo_dir.path().join("targets"));
+    let updated_metadata_base_url = &dir_url(new_repo_dir.path().join("metadata"));
+    let updated_targets_base_url = &dir_url(new_repo_dir.path().join("targets"));
     let repo = Repository::load(
         &tough::FilesystemTransport,
         Settings {
             root: File::open(&root_json).unwrap(),
-            datastore: temp_datastore.as_ref(),
-            metadata_base_url: updated_metadata_base_url,
-            targets_base_url: updated_targets_base_url,
+            datastore: None,
+            metadata_base_url: updated_metadata_base_url.clone(),
+            targets_base_url: updated_targets_base_url.clone(),
             limits: Limits::default(),
             expiration_enforcement: ExpirationEnforcement::Safe,
         },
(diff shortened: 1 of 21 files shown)


srgothi92 (approved):
LGTM.

webern:
Actually this misses the point, I'm still working on this because the Settings struct takes a &Path.

webern:
> webern force-pushed the webern:datastore-lifetime-wrk branch from 4ceff1d to 3be7cb9

OK, so in version 2 of this PR, the user can still specify a path for the datastore if they really want to. But if it's just going to be a TempDir then they can pass None and the Datastore creates the TempDir for you.

The next logical step would be to delete the `datastore` field from the `Settings` struct entirely, which would force the use of TempDir. Every use case in the Bottlerocket project that I know of is using a TempDir anyway. I checked reverse dependencies on crates.io and nobody is publicly depending on tough yet, so maybe that use case doesn't exist anywhere.

zmrow (approved):
:t-rex: 

This looks good to me - but it's probably good to get @iliana 's approval as well. :)

webern:
Before invoking iliana I'm going to do one more change that I think is essential to friendlier usability of the `Settings`. The URLs should be accepted as `String` instead of `&'a str`. If it were really important to accept `&'a str` I could use a `Cow`, but I think the cloning of two strings when constructing the `Repository` is trivial, so I'm going to go with `String`s. Version 3 of this PR coming up.

webern:
> webern force-pushed the webern:datastore-lifetime-wrk branch from 3be7cb9 to e75ec03

Version 3 of this PR: I changed `&'a str` for the URLS in repository settings to `String`s.

webern:
> webern force-pushed the webern:datastore-lifetime-wrk branch from e75ec03 to 5bd02fa

* cleaned up test cases that were using the module-qualified `test_utils::dir_url` instead of `use test_utils::dir_url`
* It was not necessary to derive Clone for the DatastorePath enum because, which means that TempDir did not need to be wrapped with `Arc`.

webern:
> webern force-pushed the webern:datastore-lifetime-wrk branch from 5bd02fa to e640965 

Me: 3, Clippy: 1

webern:
> webern force-pushed the webern:datastore-lifetime-wrk branch from e640965 to d00e956

rebase with small conflict in `download.rs`

zmrow (approved):
Still onboard!

:motor_boat:

tuftool/tests/delegation_commands.rs
```diff
@@ -1280,16 +1267,14 @@ fn remove_role_recursive_command() {
...
     let repo = Repository::load(
         &tough::FilesystemTransport,
         Settings {
             root: File::open(&root_json).unwrap(),
-            datastore: temp_datastore.as_ref(),
-            metadata_base_url: updated_metadata_base_url,
-            targets_base_url: updated_targets_base_url,
+            datastore: None,
+            metadata_base_url: updated_metadata_base_url.clone(),
+            targets_base_url: dir_url(new_repo_dir.path().join("targets")),
```
tjkirch, line 1277: nit: why do metadata URL above and targets URL here?

tjkirch (approved):

tuftool/tests/delegation_commands.rs
webern: It's used later in a `tuftool update` call on line 1360.

webern:
> webern force-pushed the webern:datastore-lifetime-wrk branch from d00e956 to c75b239

rebase

tuftool/tests/delegation_commands.rs
tjkirch: I know, I guess I didn't word my question well.  I meant, why not do targets URL the same way, so this doesn't look different?  I was searching around for reasons, where if it looked consistent I think the purpose would be more obvious.  Still, just a nit.

Issue #222, opened 2020-09-01
tuftool: download should support file://


iliana:
This is incredibly useful for testing, so let's just make it work :)

nealmcb:
Note that tuftool update already supports file:// as seen in the example in the README.md.

iliana:
Yep, I think it's a mismatch in our code between use of `Repository` for `download` and `RepositoryEditor` for `update`.

srgothi92:
Fixed in #252

PR #256, opened 2020-10-28
tough: simplify repository load

### LINKED ISSUE #248: tough: code reuse is difficult for repository structs due to the type parameter T: Transport
I thought there was already an issue for this, but I'm not seeing it.

Repository is generic on `T: Transport` which leads to a lot of unwieldy code for users. We often want to write something like this:

```rust
let repo = Repository::load(if self.metadata_base_url.scheme() == "file" {
         FilesystemTransport
     } else {
        HttpTransport::new()
    }, 
    settings)?;
```

We can't do this because of the type parameter T on Repository

```rust
pub struct Repository<'a, T: Transport> {
    transport: &'a T,
    consistent_snapshot: bool,
    datastore: Datastore<'a>,
    // ... etc
```

This means that all code that uses a repository needs to be generic on `T`, which leads to unnatural constructs. The proposal is to change the definition of Repository to:

```rust
pub struct Repository<'a> {
    transport: Box<dyn Transport>,
    consistent_snapshot: bool,
    datastore: Datastore<'a>,
    // ... etc
```

This will touch a lot of files and very much be a breaking change. We should do this before `1.0.0`.

Also, we could potentially do away with the lifetime as well if we alter the definition of `Datastore` so that it doesn't hold a ref. #249

### LINKED ISSUE #249: tough: Change definition of Datastore to eliminate lifetime

### LINKED ISSUE #250: tough: transport trait returns a generic error type
There are cases where it would be good to know more about what type of error was thrown during the `Transport` trait's `fetch` function. For example, when checking for a higher-versioned `root.json` file, it would be good to know if the error was `NotFound` vs. say, some network failure or other io failure.

The trait is currently defined as:

```rust
/// A trait to abstract over the method/protocol by which files are obtained.
pub trait Transport {
    /// The type of `Read` object that the `fetch` function will return.
    type Stream: Read;

    /// The type of error that the `fetch` function will return.
    type Error: std::error::Error + Send + Sync + 'static;

    /// Opens a `Read` object for the file specified by `url`.
    fn fetch(&self, url: Url) -> Result<Self::Stream, Self::Error>;
}
```

Because the error type is a trait-bound generic, and because the trait bounds do not provide information about the underlying error (e.g. FileNotFound vs some sort of IoFailure), the user of the `fetch` function cannot retrieve this information if their code is generic on `T: Transport`.

we have thought of two ways to fix this, both are 'breaking' and should be considered before v1.0.0.

## Option 1

Return a specific error type such that implementers of `Transport` need to return our error type. For example:

(Note these could/should be made richer to cary, e.g. the file url and maybe an `Option<Box<dyn std::error::Error + Send + Sync + 'static>>` if that's a thing.

```rust
/// A trait to abstract over the method/protocol by which files are obtained.
pub trait Transport {
    /// The type of `Read` object that the `fetch` function will return.
    type Stream: Read;

    /// Opens a `Read` object for the file specified by `url`.
    fn fetch(&self, url: Url) -> Result<Self::Stream, TransportError>;
}

/// Provides a `Transport` for local files.
#[derive(Debug, Clone, Copy)]
pub struct FilesystemTransport;

/// The Error type returned by `Transport` objects.
enum TransportError {
    FileNotFound,
    Failure,
}
```

## Option 2

Add a custom trait to the error type that users must implement. e.g.

```rust
/// A trait to abstract over the method/protocol by which files are obtained.
pub trait Transport {
    /// The type of `Read` object that the `fetch` function will return.
    type Stream: Read;

    /// The type of error that the `fetch` function will return.
    type Error: std::error::Error + Send + Sync + FetchError + 'static;

    /// Opens a `Read` object for the file specified by `url`.
    fn fetch(&self, url: Url) -> Result<Self::Stream, Self::Error>;
}

enum FetchErrorKind {
    FileNotFound,
    Failure,
}

trait FetchError {
    fn kind(&self) -> FetchErrorKind;
}

impl FetchError for std::io::Error {
    fn kind(&self) -> FetchErrorKind {
        match self.kind() {
            ErrorKind::NotFound => FetchErrorKind::FileNotFound,
            _ => FetchErrorKind::Failure,
        }
    }
}
```

webern:
## Issues

Closes #248 
Closes #249 
Closes #250 
Kinda Reimplements the Solution to #222

A companion PR is open in Bottlerocket that shows the changes there:
https://github.com/bottlerocket-os/bottlerocket/pull/1190

## Description of changes

This PR eliminates the type and lifetime parameters on the Repository struct. Additionally, the Repository load function is re-organized to allow `tough` users to default all of the parameters that we most often see defaulted in known use cases.

This is a large PR, but the bulk of the files touched are tests, repo-editor and tuftool code where the `Repository::load` function was in use.

Here is a guide to the files that are driving all of the changes:

#### `tough/src/transport.rs`

In order to eliminate the type parameter from `Repository`, it was necessary to remove the 'associated types' from `Transport`. That is, `Transport` was generic on the returned `Read` type and on the `Error` type. To solve this, I went with returning a `Box<dyn Read>` and a concrete `TransportError`, which is easily constructed to wrap ones underlying errors.

One other possibility would have been to return a `Box<dyn Error + Send + Sync + TransportProblem>` as the error type, but this would be weird because it would not implement `Error` itself. (BTW adding `TransportProblem` in that pretend signature is another way of solving #250, which is also solved by returning a `TransportError`.)

The benefit of the changes to the `Transport` trait is that we can now hold a `Box<dyn Transport>` in the `Repository` object without caring about its underlying types. The downside is a slight additional burden on those that would implement their own custom `Transport` types, in that they need to wrap their `Error`s correctly into `TransportError`s.

~The addition of the `box_clone` function to the trait is a necessity of the type system if we want transports to be cloneable. (Nope, adding + Clone to the trait will not work, I tried to explain in the comments). You could be tricky and provide a default implementation of the function by introducing another trait, as seen [here](https://stackoverflow.com/a/30353928/2779792), but I think my solution is easier to understand for a would-be transport implementor.~ This has been changed to use `DynClone` which is much better.

##### `DefaultTransport`

Why not implement a transport that supports both `file` and `http`? In fact, both of these objects are trivially constructible and tiny, so the cost of having both instead of choosing one is negligible. The only weirdness is that HTTP won't be there when the library is compiled without the `http` feature. To help users who may encounter a surprise, an error message is conditionally compiled in the absence of the `http` feature that specifically mentions that the feature is missing. Tests are conditionally compiled that exercise both of these 'static branches'.

#### `tough/src/lib.rs`

Having eliminated the Transport type parameter, and having provided a default Datastore using TempDir in #253, it became clear to me that there only three things needed to load a repo:
 * root.json
 * metadata url
 * targets url

Everything else can be defaulted, and *is* defaulted (or would be if it could) in the majority of our existing use cases. So I created a new struct, called `RepositoryLoader` which is constructed with the three required arguments. After construction it can be used like a builder to set optional parameters such as Datastore and Transport. Finally you call `load` (i.e. usually called `build` in a builder) to load a repository.

#### Everything Else

Everything else (test files, repo editor, `tuftool`) is changed to accommodate the changes to `tough/src/transport.rs` and `tough/src/lib.rs`. In many cases you can see the benefit is simpler code (e.g. `tuftool/src/key_role.rs`).

## Commits

From oldest to newest:

```
    tough: remove type parameter for transport

    Removes both the lifetime and type parameters for the Repository struct.
    By holding the Transport as a trait object, using Box<dyn Transport>,
    the Repository struct no longer needs a type parameter.

    In order to accomplish this, it was necessary to remove associated types
    from the Transport trait. This leads us to return a Box<dyn Read>
    and to introduce a concrete TransportError type to be used by
    implementors of the Transport trait.

    The upside of this change is that, for the most common known use cases
    of Repository use (file and http), it is no longer necessary to make the
    code generic on T: Transport. The trade-off is a slight bit of
    additional work for those that would implement their own Transport
    trait.

    Some refactoring in the HTTP transport was called for. Though not
    strictly necessary, the retry loop was so difficult to understand that
    I found it necessary to factor out the inspection of the Result we get
    from reqwest into its own set of functions.
```

```
    tough: add a default transport for http and file

    Adds a default transport object which supports both file transport and,
    if the http feature is enable, HTTP transport. Gives an error message
    in the absence of the http feature to help the user understand why http
    transport did not work.
```

```
    tough: simplify repository load

    Introduces a builder object, RepositoryLoader, that takes root,
    metadata_base_url, and targets_base_url in its new function. After that
    the user can add optional settings such as transport, limits, etc.
    Finally, the user calls the load function to load a repository.

    The load function on the Repository struct is now private so that the
    user does not have to navigate multiple methods of loading a repository.
```

## Testing

* The following tuftool commands are working
  * `tuftool create`
  * `tuftool download file://`
  * `tuftool download https://`
  * tuftool root commands e.g. add-key, bump-version, etc
  * tuftool root sign
* Bottlerocket: extensive testing done, e.g., see https://github.com/bottlerocket-os/bottlerocket/pull/1190 for details.

## Disclaimer


tough/src/editor/test.rs
```diff
@@ -6,7 +6,7 @@ mod tests {
     use crate::editor::RepositoryEditor;
     use crate::key_source::LocalKeySource;
     use crate::schema::{Signed, Snapshot, Target, Targets, Timestamp};
-    use crate::transport::FilesystemTransport;
+    // use crate::transport::FilesystemTransport;
```
webern:
```suggestion
```

webern:
pubsys relies on the fact that the Transport implementations it was using both are `Send`, so I added it to the requirements of the Transport trait's Read return type. Unfortunately this reduces the flexibility for Transport implementers.

Edit: uhhh, that didn't fix my pubsys issue, so maybe that commit will come back out.

webern:
> webern force-pushed the webern:transport-develop branch from 880a530 to a3aed91

* Propagate the `+ Send` requirement to various function return signatures as necessary to allow us to async move a returned Read, as we do in pubsys check repo.
* Remove now unused errors and make the Transport error hold a concrete type so users can find out what went wrong during transport (which we can use in pubsys!)

tough/src/editor/mod.rs
```diff
@@ -608,10 +608,15 @@ impl<'a, T: Transport> RepositoryEditor<'a, T> {
...
     ) -> Result<&mut Self> {
         let limits = self.limits.context(error::MissingLimits)?;
-        let transport = self.transport.context(error::MissingTransport)?;
-
+        let transport = self
+            .transport
+            .as_ref()
+            .context(error::MissingTransport)?
+            .boxed_clone();
+        // self.check_transport()?;
```
webern:
```suggestion
```

tough/src/http.rs
```diff
@@ -180,81 +192,217 @@ impl RetryState {
...
+        }
+    }
+}
+
+/// Checks the HTTP response code and converts a non-successful response code to an error.
+fn parse_response_code(response: reqwest::blocking::Response) -> HttpResult {
+    match response.error_for_status() {
+        Ok(ok) => {
+            trace!("response is success");
+            // http status is ok. return early from this function with happiness
```
webern:
```suggestion
            // http status code indicates success
```

tough/src/transport.rs
```diff
@@ -1,36 +1,215 @@
...
 use url::Url;
 
 /// A trait to abstract over the method/protocol by which files are obtained.
-pub trait Transport {
-    /// The type of `Read` object that the `fetch` function will return.
-    type Stream: Read;
+///
+/// The trait hides the underlying types involved by returning the `Read` object as a
+/// `Box<dyn Read>` and by requiring concrete type [`TransportError`] as the error type.
+///
```
webern:
```suggestion
/// The trait hides the underlying types involved by returning the `Read` object as a
/// `Box<dyn Read + Send>` and by requiring concrete type [`TransportError`] as the error type.
///
```

tough/src/transport.rs
```diff
@@ -1,36 +1,215 @@
...
-    /// The type of error that the `fetch` function will return.
-    type Error: std::error::Error + Send + Sync + 'static;
+    /// Returns a clone of `self` as a `Box<dyn Transport>`.
+    ///
+    /// # Why
+    ///
+    /// The [`Repository`] object holds a `Box<dyn Transport>`, and because we want the `Repository`
+    /// object to implement `Clone`, we need a way of cloning the boxed `Transport` without knowing
+    /// its underlying type. We cannot require require the `Clone` trait bound on `Transport`
+    /// because, if we do, the trait can no longer serve as a 'trait object'.
```
webern:
```suggestion
    /// its underlying type. We cannot require the `Clone` trait bound on `Transport`
    /// because, if we do, the trait can no longer serve as a 'trait object'.
```

webern:
Testing completed on `4cdaf14`, now described in PR.

webern:
On no. Well I used `DynClone` which is much better than having a weird function on the `Transport` trait, and I did a git fixup. But something has gone horribly wrong and now I seem to be including two commits that aren't mine. Maybe I just need to rebase.

webern:
> webern force-pushed the webern:transport-develop branch from 89cac1a to 76cfbc2

OK, rebased over develop with some conflict resolution.

webern:
> webern force-pushed the webern:transport-develop branch from 76cfbc2 to 3e858e5

Rebase after merging #253. This removes the datastore lifetime commit (c75b239) from the current PR.

tough/src/http.rs
```diff
@@ -180,81 +188,217 @@ impl RetryState {
...
             .request(Method::GET, url.as_str())
             .header(header::RANGE, header_value)
             .build()
-            .context(error::HttpRequestBuild { url: url.clone() })?;
+            .context(http_error::RequestBuild)?;
         Ok(request)
     }
 }
+
+mod http_error {
```
tjkirch: It looks like this naming isn't because we have another error module in scope, or anything like that - can we stick with "error" to fit convention?  (See also my suggestions below, which would take away the need to reference the name of this module to clarify the meaning of `to_transport`.)

tough/src/http.rs
```diff
@@ -180,81 +188,217 @@ impl RetryState {
...
             }
-            Err(err) => {
-                // if it's a status code error other than 5XX, return the error
-                if let Some(status) = err.status() {
-                    if !status.is_success() && !status.is_server_error() {
-                        return Err(err).context(error::HttpFetch { url: url.clone() });
-                    }
+            HttpResult::Fatal(err) => {
+                trace!("{:?} - returning fatal error from fetch: {}", r, err);
+                return Err(http_error::HttpError::FetchFatal { source: err });
```
tjkirch: I could be missing the purpose, but it's unusual to reference the error enum variants directly rather than using snafu's generated structs; could you do something like `return error::FetchFatal { source: err }.fail();`?  Or just `Err(err).context(error::FetchFatal)?;`?

(Also, it's unrelated since you need a Result here, but I also learned of the `.build()` method, which is like `.fail()` but just gives the inside error without wrapping it in a Result.)

tough/src/transport.rs
```diff
@@ -1,36 +1,120 @@
...
+///
+#[derive(Debug, Copy, Clone)]
+#[non_exhaustive]
+pub enum TransportErrorKind {
+    /// The trait does not handle the URL scheme named in `String`. e.g. `file://` or `http://`.
+    UnsupportedUrlScheme,
+    /// The file cannot be found.
+    FileNotFound,
+    /// The transport failed for any other reason, e.g. IO error, HTTP broken pipe, etc.
+    Failure,
```
tjkirch: What do you think of matching the `io` convention and calling it `Other`?

tough/src/http.rs
```diff
@@ -180,81 +188,217 @@ impl RetryState {
...
+        InvalidHeader {
+            header_value: String,
+            source: reqwest::header::InvalidHeaderValue,
+        },
+
+        #[snafu(display("Unable to create HTTP request: {}", source))]
+        RequestBuild { source: reqwest::Error },
+    }
+
+    impl Into<std::io::Error> for HttpError {
```
tjkirch: It's [recommended](https://doc.rust-lang.org/std/convert/trait.From.html) to implement From instead of Into - and very strongly recommended, since 1.41 - is there a reason I'm missing to stick with Into here?

(Same above with `Into<HttpResult>`)

tough/src/http.rs
```diff
@@ -180,81 +188,217 @@ impl RetryState {
...
+            match self {
+                HttpError::FetchFileNotFound { .. } => {
+                    std::io::Error::new(ErrorKind::NotFound, self)
+                }
+                _ => std::io::Error::new(ErrorKind::Other, self),
+            }
+        }
+    }
+
+    pub(super) fn to_transport(url: String, e: HttpError) -> TransportError {
```
tjkirch: Similar to the last comment, I think this could be clearer as something like `TransportError::from((url, e))` because it expresses the type the user wants and uses a standard function rather than us having to come up with a name.  (I was having a hard time understanding the types and linkage here.)

tough/src/http.rs
```diff
@@ -105,19 +104,12 @@ impl Read for RetryRead {
...
-                    "an error occurred and we cannot retry because the server \
-                    does not support range requests '{}': {:?}",
-                    self.url, retry_err
-                );
-                return Err(retry_err);
-            }
             let new_retry_read =
-                fetch_with_retries(&mut self.retry_state, &self.settings, &self.url)?;
+                fetch_with_retries(&mut self.retry_state, &self.settings, &self.url)
+                    .map_err(|e| std::io::Error::new(std::io::ErrorKind::Other, e))?;
```
tjkirch, line 113: If I'm understanding things correctly, we map to an io::Error here, and then the From/Into implementation below maps it back to an HttpError, and we do that because the From/Into implementation checks whether it was a NotFound or Other type of error and constructs the right variant of HttpError.  I think that's a little confusing, or at least not obvious.  I'm still not sure I have a full picture, because fetch_with_retries is already bucketing NotFound into the right HttpError.  Why do we do this mapping?

If possible, I'd love to see some kind of simplification in the error/result types so that io::Errors are categorized immediately as NotFound/Other so they don't have to escape an io call site, and maybe not tying TransportError and HttpError together (maybe using TransportError more directly?) so that confusion could be reduced...  if not, I do think we could use clearer function signatures / documentation about the relationships and conversions.

tough/src/http.rs
```diff
@@ -180,81 +188,217 @@ impl RetryState {
...
+                trace!("{:?} - returning fatal error from fetch: {}", r, err);
+                return Err(http_error::HttpError::FetchFatal { source: err });
+            }
+            HttpResult::FileNotFound(err) => {
+                trace!("{:?} - returning file not found from fetch: {}", r, err);
+                return Err(http_error::HttpError::FetchFileNotFound { source: err });
+            }
+            HttpResult::Retryable(err) => {
+                trace!("{:?} - retryable error: {}", r, err);
+                if r.current_try >= cs.tries - 1 {
```
tjkirch, line 233: nit about existing code: `>= x - 1` is an off-by-one waiting to happen; we should probably start `current_try` at 1 instead of 0...

tough/src/http.rs
```diff
@@ -180,81 +188,217 @@ impl RetryState {
...
+        }
+        // http status is an error
+        Err(err) => match err.status() {
+            None => {
+                // this shouldn't happen, we received this err from the err_for_status function,
+                // so the error should have a status. we cannot consider this a retryable error.
+                trace!("error is fatal (no status): {}", err);
+                HttpResult::Fatal(err)
+            }
+            Some(status) => parse_status_err(err, status),
```
tjkirch: I had some trouble following the naming down here; would it be clearer as a single match in this function, using conditional guards?  Something like:
```rust
Some(status) if status.is_server_error() => HttpResult::Retryable ...
Some(status) if matches!(status.as_u16(), 403 | 404) => HttpResult::FileNotFound ...
Some(status => HttpResult::Fatal ...
```

tough/src/http.rs
```diff
@@ -180,81 +188,217 @@ impl RetryState {
...
+        match self {
+            Ok(response) => {
+                trace!("response received");
+                // checks the status code of the response for errors
+                parse_response_code(response)
+            }
+            Err(err) => {
+                // an error occurred before the HTTP header could be read
+                trace!("retryable error during fetch: {}", err);
+                HttpResult::Retryable(err)
```
tjkirch: I'm confused - how do we know that all (non-server) errors are retryable?

tough/src/http.rs
```diff
@@ -180,81 +188,217 @@ impl RetryState {
...
             .header(header::RANGE, header_value)
             .build()
-            .context(error::HttpRequestBuild { url: url.clone() })?;
+            .context(http_error::RequestBuild)?;
         Ok(request)
     }
 }
+
+mod http_error {
+    #![allow(clippy::default_trait_access)]
```
tjkirch: Is this still needed?  I'm not seeing a call below that would violate it...

tough/src/http.rs
```diff
@@ -180,81 +188,217 @@ impl RetryState {
...
+    #[allow(missing_docs)]
+    pub enum HttpError {
+        #[snafu(display("The HTTP client could not be built: {}", source))]
+        Client { source: reqwest::Error },
+
+        #[snafu(display("A non-retryable error occurred: {}", source))]
+        FetchFatal { source: reqwest::Error },
+
+        #[snafu(display("File not found: {}", source))]
+        FetchFileNotFound { source: reqwest::Error },
```
tjkirch: Is it going to be clear *what* wasn't found, from context introduced elsewhere?

I guess this question applies to most of the error variants; is the user going to know what was being attempted when a request went sideways?  Does reqwest always say?

tough/src/lib.rs
```diff
@@ -176,10 +176,11 @@ impl Default for Limits {
 
 /// A TUF repository.
 ///
-/// You can create a `Repository` using the `load` method.
+/// You can create a `Repository` using the `load` method, or the `load_default`
+/// method (to use default options).
```
tjkirch: `load_default` wasn't introduced until a later commit.

tough/src/transport.rs
```diff
@@ -1,36 +1,120 @@
...
+
+/// The error type that [`Transport`] `fetch` returns.
+#[derive(Debug)]
+pub struct TransportError {
+    /// The kind of error that occurred.
+    pub kind: TransportErrorKind,
+    /// The URL that the transport was trying to fetch.
+    pub url: String,
+    /// The underlying error that occurred.
+    pub source: Box<dyn std::error::Error + Send + Sync>,
```
tjkirch, line 57: Out of curiosity, why require Sync?  (And perhaps even... why Send?)

tough/src/transport.rs
```diff
@@ -1,36 +1,120 @@
...
+impl TransportError {
+    /// Creates a new [`TransportError`].
+    pub fn new<S, E>(kind: TransportErrorKind, url: S, source_error: E) -> Self
+    where
+        E: Into<Box<dyn std::error::Error + Send + Sync>>,
+        S: AsRef<str>,
+    {
+        Self {
+            kind,
+            url: url.as_ref().to_owned(),
```
tjkirch: Would `Into<String>` be clearer if our goal is actually to get a `String` for the url?

tough/src/transport.rs
```diff
@@ -1,36 +1,120 @@
...
+            source: source_error.into(),
+        }
+    }
+
+    /// Creates a [`TransportError`] for reporting an unhandled URL type.
+    pub fn unsupported_url<S: AsRef<str>>(url: S) -> Self {
```
tjkirch: `unsupported_scheme` perhaps, given the kind/message it creates?

tough/src/transport.rs
```diff
@@ -1,36 +1,120 @@
...
+        TransportError::new(
+            TransportErrorKind::UnsupportedUrlScheme,
+            url,
+            "Transport cannot handle the given URL scheme.".to_string(),
+        )
+    }
+}
+
+/// [`TransportError`] implements the standard error interface.
```
tjkirch: snafu 0.6.9 supports deriving Snafu on structs (not just for making an error opaque) and it might get you what you want here, without manually deriving Error and Display.

tough/src/lib.rs
```diff
@@ -215,15 +216,15 @@ impl<'a, T: Transport> Repository<'a, T> {
     ///
     /// `metadata_base_url` and `targets_base_url` are the HTTP(S) base URLs for where the client
     /// can find metadata (such as root.json) and targets (as listed in targets.json).
-    pub fn load<R: Read>(transport: &'a T, settings: Settings<R>) -> Result<Self> {
+    pub fn load<R: Read>(transport: Box<dyn Transport>, settings: Settings<R>) -> Result<Self> {
```
tjkirch: I think you may be able to simplify the interface by taking `Into<Box<dyn Transport>>` (or `impl Into<>` to avoid generics?) because Box has a blanket impl for any T (that just boxes it) so you can call `.into()` inside this function to box the input if it's not already.  Then this would accept `Transport` or `Box<Transport>`.

(I'm thinking this would cut out half of the changed files from this commit, which are just adding `Box::new` to load calls...)

tjkirch (commented):
Comments on the first commit.

tough/src/transport.rs
```diff
@@ -1,36 +1,191 @@
-use std::io::Read;
+#[cfg(feature = "http")]
+use crate::ClientSettings;
+#[cfg(feature = "http")]
+use crate::HttpTransport;
```
tjkirch:
```suggestion
#[cfg(feature = "http")]
use crate::{ClientSettings, HttpTransport};
```

tjkirch (commented):
Just started reviewing the "simplify repository load" commit, but what about using something like https://github.com/idanarye/rust-typed-builder to make a convenient builder for the arguments instead of splitting the struct?  If the crate doesn't work, the basic approach (from my message elsewhere) is something like:

> I wonder if a SettingsBuilder would be a clearer way to do the Repository creation.  You could even do it without a Result, all at compile time, since there are only 3 mandatory settings.  e.g. the first Builder is empty and has 3 methods, one for each mandatory arg; each of those methods results in a new type that has two methods, the ones not yet given, etc.  Only the struct with all 3 has the methods for all the optional settings, and a build() method.

I assume that's basically what typed-builder does.

[edit] This could simplify the calls in tuftool, perhaps removing the need for the common module, and the special case in tuftool/src/update.rs that can't use the common function just because it needs to set one more option.

tough/src/lib.rs
```diff
@@ -198,72 +231,76 @@ pub struct Repository {
 }
 
 impl Repository {
+    /// Load and verify TUF repository metadata with default [`Options`]. See
+    /// [`Repository::load`] for more into.
+    pub fn load_default<R, S1, S2>(settings: Settings<R, S1, S2>) -> Result<Self>
```
tjkirch: I expect it to be pretty rare that people want to customize the options.  Should this be `load` and `load` be `load_custom` or whatever?  (If we change or eliminate the settings / options split, this wouldn't apply, of course.)

tuftool/src/common.rs
```diff
@@ -0,0 +1,28 @@
+/// This module if for code that is re-used by different `tuftool` subcommands.
```
tjkirch:
```suggestion
/// This module is for code that is re-used by different `tuftool` subcommands.
```

tuftool/src/common.rs
```diff
@@ -0,0 +1,28 @@
...
+use std::path::Path;
+use tough::{Repository, Settings};
+
+/// Some commands only deal with metadata and never use a targets directory.
+/// When loading a repo that does not need a targets directory, we pass this as
+/// the targets URL.
+pub(crate) const UNUSED_URL: &str = "file:///unused/url";
+
+/// Load a repo for metadata processing only. Such a repo will never use the
+/// targets directory, so a dummy path is passed.
```
tjkirch, line 15: Might also be worth noting that it's only for local file paths -- maybe even renaming the function? -- it seems too easy to pass a URL as a string, which is a valid `AsRef<Path>`.

tuftool/src/remove_key_role.rs
```diff
@@ -55,44 +53,16 @@ pub(crate) struct RemoveKeyArgs {
...
+            role,
+            TargetsEditor::from_repo(repository, role)
+                .context(error::EditorFromRepo { path: &self.root })?,
+        )
     }
 
-    /// Removes keys from adelegated role using targets Editor
-    fn with_targets_editor(&self, role: &str, mut editor: TargetsEditor) -> Result<()> {
+    /// Removes keys from a delegated role using targets Editor
+    fn remove_key_with_targets_editor(&self, role: &str, mut editor: TargetsEditor) -> Result<()> {
```
tjkirch: nit: rename similar methods in other modules to fit a pattern?  For example, I see a similar change in remove_role.rs, but not add_key_role.rs.

tjkirch (commented):
So many beautiful red lines!

tough/src/http.rs
zmrow: Agreed there - I'm a bit confused as well.

tough/src/lib.rs
```diff
@@ -135,12 +152,30 @@ pub struct Settings<R: Read> {
     pub expiration_enforcement: ExpirationEnforcement,
 }
 
+impl Default for Options {
```
zmrow: :money_with_wings:

tough/src/lib.rs
zmrow: Yeah - I like changing this to `load`.  So far we haven't needed to change options in any of our code.

zmrow (commented):
Sorry for the late response.  

I'm in agreement with @tjkirch in re: to the error comments.  Everything else looks fairly straightforward!

Nice job!

tough/src/http.rs
webern: The implementation is confusing by nature, it might actually require a meeting to try and go through it. Basically what happens is:
- `Read` is a trait, it returns a `std::io::Error/Result`
- `fetch` returns a transport error.

Let's walk through a case where a broken-pipe occurs while reading the request body. It goes like this:

- `fetch` returns Ok(Read)
- User calls Read
- Error occurs in the inner Read object, this will be an io::Error
- We do retries, which means we are using fetch again, this returns a TransportError/Result.
- If we try fetch and fail, run out of retries etc, now we are holding a ~TransportError...~ HttpError but:
  - We are inside of a Read object's `read` function so we need to wrap it up in std::io::Error/Result
  - Furthermore, since we are inside of a `Read` trait call, it would be illogical at this point to say the file isn't found or anything else, here we don't bother looking at the `TransportError` because no matter what it is, the true and correct answer to the caller is `std::io::ErrorKind::Other`

tuftool/src/remove_key_role.rs
webern: I apologize, but I'm absolutely going to need a method of marking these things 'staged' prior to pushing. So the 'staged' issues will get marked like this (this is the message of a locally held commit):

fixup: simplify repository load - tuftool with_targets_editor naming convention

tuftool/src/common.rs
webern: fixup: simplify repository load - tuftool load_metadata documentation

tuftool/src/common.rs
webern: fixup: simplify repository load - tuftool common.rs typo fix

tough/src/transport.rs
webern: fixup: remove type parameter for transport - tough transport.rs use snafu

tough/src/transport.rs
webern: fixup: remove type parameter for transport - tough transport.rs unsupported_scheme

tough/src/transport.rs
webern: I tried this and it seemed less convenient for the caller. At a call site like this:
`TransportError::unsupported_scheme(url)`
The compiler was unhappy with:
```
the trait `std::convert::From<url::Url>` is not implemented for `std::string::String`
```
whereas, with `AsRef<str>` the compiler is happy to accept a `Url`.

tough/src/transport.rs
webern: I'm following what appears to be a pattern without a deep understanding of why. `std::io`, `Snafu`, and some of @iliana's code all use this convention. My guess is that we are adopting an idiom that assumes that everyone wants errors to be sendable and queryable in multithreaded applications.

Check out the `new` function here: https://doc.rust-lang.org/std/io/struct.Error.html

tough/src/transport.rs
webern: fixup: remove type parameter for transport - tough transport.rs unsupported_scheme

tough/src/transport.rs
webern: fixup: remove type parameter for transport - tough transport.rs cleanup http imports

tough/src/lib.rs
webern: My IDE made this possible in 2 seconds, yay. I agree this is better.
fixup: simplify repository load - load_with_options

tough/src/lib.rs
webern: FIXME: I'll need to come back and figure this out.

tough/src/http.rs
webern: Yes, all transport errors will include the URL, it is a member of the TransportError struct, so we add it in the final wrap before returning from fetch.

tough/src/http.rs
webern: This impl was unused which added some confusion, I've removed it.

tough/src/http.rs
webern: P.S. The implementation of `From<HttpError> for std::io::Error` was completely unused and probably added to the confusion, I have removed it.

tough/src/http.rs
webern: fixup: remove type parameter for transport - tough http.rs rename to_transport_error to from and remove http_error module

tough/src/http.rs
webern: fixup: remove type parameter for transport - tough http.rs rename to_transport_error to from and remove http_error module

tough/src/http.rs
webern: P.S. my IDE fails me on this one. It turns these red, I think because the macro expansion happens after the usage site. It compiles just fine, but I abhor red things in my code editor which is why I had it there. I've removed it since it's definitely not needed.

tough/src/http.rs
webern: Yeah, that's advanced syntax not at my fingertips. Looks better.

save fixup: remove type parameter for transport - tough http.rs better match syntax

tough/src/http.rs
webern: FIXME: These are not server errors, these are any errors that occur before an HTTP status is comprehended by reqwest. For example
- Broken pipe before the status is reached
- connection timeout
- reqwest exploded

I'll come back to this and see how messy it would be to make a decision based on the error here.

Your confusion is understandable. The comprehension of these errors is one of the most confusing parts of implementing this module.

tough/src/http.rs
webern: FIXME: I'm afraid to touch it right now.

tough/src/http.rs
webern: Err(err).context(TheStruct) works. Fail didn't, maybe it needs to be brought into scope or something.

fixup: remove type parameter for transport - tough http.rs use snafu fancy stuff

webern (commented):
I really needed a system of knowing which fixes I've staged locally, which is why nearly every issue has a response with `fixup: blah`.

A few issues I decided to defer to another pass, those are marked with a `FIXME: something`.

I will work on getting things rebased to the correct commits next, once I have done so and force pushed, I will mark all of the 'done' issues with the sha/link of that push.

webern:
> webern force-pushed the webern:transport-develop branch from 3e858e5 to 59899dd

Addresses a fair amount of the issues raised thus far. Each issue will be marked with `59899dd` to indicate that it was addressed in this push.

tuftool/src/common.rs
webern: Addressed in [59899dd](https://github.com/awslabs/tough/compare/3e858e5b5d629576a0c13f3f7b627cd8c49af9a4..59899ddf735ceb7e4edc883a9b1c6a195506f820),

tough/src/http.rs
webern: Addressed in [59899dd](https://github.com/awslabs/tough/compare/3e858e5b5d629576a0c13f3f7b627cd8c49af9a4..59899ddf735ceb7e4edc883a9b1c6a195506f820),

tough/src/http.rs
webern: Addressed in [59899dd](https://github.com/awslabs/tough/compare/3e858e5b5d629576a0c13f3f7b627cd8c49af9a4..59899ddf735ceb7e4edc883a9b1c6a195506f820),

tough/src/transport.rs
webern: Addressed in [59899dd](https://github.com/awslabs/tough/compare/3e858e5b5d629576a0c13f3f7b627cd8c49af9a4..59899ddf735ceb7e4edc883a9b1c6a195506f820),

tough/src/http.rs
webern: Addressed in [59899dd](https://github.com/awslabs/tough/compare/3e858e5b5d629576a0c13f3f7b627cd8c49af9a4..59899ddf735ceb7e4edc883a9b1c6a195506f820),

tough/src/http.rs
webern: Addressed in [59899dd](https://github.com/awslabs/tough/compare/3e858e5b5d629576a0c13f3f7b627cd8c49af9a4..59899ddf735ceb7e4edc883a9b1c6a195506f820),

tough/src/http.rs
webern: Addressed in [59899dd](https://github.com/awslabs/tough/compare/3e858e5b5d629576a0c13f3f7b627cd8c49af9a4..59899ddf735ceb7e4edc883a9b1c6a195506f820),

tough/src/http.rs
webern: Addressed in [59899dd](https://github.com/awslabs/tough/compare/3e858e5b5d629576a0c13f3f7b627cd8c49af9a4..59899ddf735ceb7e4edc883a9b1c6a195506f820),

tough/src/transport.rs
webern: Addressed in [59899dd](https://github.com/awslabs/tough/compare/3e858e5b5d629576a0c13f3f7b627cd8c49af9a4..59899ddf735ceb7e4edc883a9b1c6a195506f820),

tough/src/transport.rs
webern: Addressed in [59899dd](https://github.com/awslabs/tough/compare/3e858e5b5d629576a0c13f3f7b627cd8c49af9a4..59899ddf735ceb7e4edc883a9b1c6a195506f820),

tough/src/lib.rs
webern: I'm still deciding whether it's worth it to make this change to the early commit named `remove type parameter for transport` since that is a transient commit anyway.

But where I should, and definitely will, implement this change is in the commit named `simplify repository load` where the `Box<dyn Transport>` is a member of `Options`.

tough/src/transport.rs
webern: Addressed in [59899dd](https://github.com/awslabs/tough/compare/3e858e5b5d629576a0c13f3f7b627cd8c49af9a4..59899ddf735ceb7e4edc883a9b1c6a195506f820),

tough/src/lib.rs
webern: Addressed in [59899dd](https://github.com/awslabs/tough/compare/3e858e5b5d629576a0c13f3f7b627cd8c49af9a4..59899ddf735ceb7e4edc883a9b1c6a195506f820),

tuftool/src/common.rs
webern: Addressed in [59899dd](https://github.com/awslabs/tough/compare/3e858e5b5d629576a0c13f3f7b627cd8c49af9a4..59899ddf735ceb7e4edc883a9b1c6a195506f820),

tuftool/src/remove_key_role.rs
webern: Addressed in [59899dd](https://github.com/awslabs/tough/compare/3e858e5b5d629576a0c13f3f7b627cd8c49af9a4..59899ddf735ceb7e4edc883a9b1c6a195506f820),

webern:
> webern force-pushed the webern:transport-develop branch from 59899dd to 55ff028

rebase develop

webern:
> Just started reviewing the "simplify repository load" commit, but what about using something like https://github.com/idanarye/rust-typed-builder to make a convenient builder for the arguments instead of splitting the struct? If the crate doesn't work, the basic approach (from my message elsewhere) is something like:

I like the idea of a builder, however I don't like the idea of rust-typed-builder for the required parameters. I think 'keep it simple' goes a long way here. We have three required params and it's highly unlikely that any additional _required_ params will be added in the future. So I would suggest something more like this:

```rust
RepositoryLoader::new(Settings{
    root: myroot,
    metadata_url: foo,
    targets_url: bar,
})
.expiration_enforcement(ExpirationEnforcement::Unsafe)
.datastore(mypath)
.load();
```

or

```rust
RepositoryLoader::new(root, metadata_url, targets_url)
.expiration_enforcement(ExpirationEnforcement::Unsafe)
.datastore(mypath)
.load();
```

In both of these cases the `Options` struct goes away in favor of a builder.

I think either of these options is going to be much easier for the user to understand than a compilation error because a required argument is missing from a 'clever' macro-defined set of types and functions. I haven't tried that lib to see what the compiler actually says, but I think the two usage examples above are dead-simple and easy to understand for a user.

So, yeah, I think it's a great callout that we could have a builder here, but I think we can do it without macros! :)

tjkirch:
Sure!  I like the latter option with parameters instead of a struct.

tough/src/http.rs
tjkirch: Do we still need `impl Into<HttpResult>`?

tough/src/http.rs
```diff
@@ -180,81 +189,182 @@ impl RetryState {
...
+        header_value: String,
+        source: reqwest::header::InvalidHeaderValue,
+    },
+
+    #[snafu(display("Unable to create HTTP request: {}", source))]
+    RequestBuild { source: reqwest::Error },
+}
+
+impl TransportError {
+    fn from(url: String, e: HttpError) -> Self {
```
tjkirch: Can we implement the `From` trait instead?

tough/src/transport.rs
tjkirch: Ahh, I guess because `Url` implements `AsRef<str>` but not `From<Url> for String`.  No biggie.

tough/src/http.rs
webern: From what? A tuple?

tjkirch: Yep, `impl From<(String, HttpError)> for TransportError`.  Manual `from` is fine if you hate it.  ;p

tough/src/http.rs
webern: It is being used.

tough/src/http.rs
webern: I added an issue for this, I'd like to get some test assertions in place before making this change. https://github.com/awslabs/tough/issues/272

tough/src/http.rs
tjkirch: I mean as opposed to `From`, which is strongly recommended in the docs.

webern:
> webern force-pushed the webern:transport-develop branch from 55ff028 to f5bfcbb

* Use From<(tuple)> somewhere
* Improve an error checking in http.rs

> webern force-pushed the webern:transport-develop branch from f5bfcbb to 33f4823

Rebase

tough/src/http.rs
webern: This is done in [f5bfcbb](https://github.com/awslabs/tough/compare/55ff0282802e8c7829762f923d6e92ced3a49b93..f5bfcbb4be7dcafa0a25b4ff8fb02649a2c886f1)

tough/src/http.rs
webern: This is improved [f5bfcbb](https://github.com/awslabs/tough/compare/55ff0282802e8c7829762f923d6e92ced3a49b93..f5bfcbb4be7dcafa0a25b4ff8fb02649a2c886f1)

webern:
> webern force-pushed the webern:transport-develop branch from 33f4823 to 3def6ac

Fix a silly typo.

webern:
> webern force-pushed the webern:transport-develop branch from 3def6ac to 9bbdd3e

* fix a new clippy lint
* use From instead of Into in http.rs

tough/src/http.rs
webern: Now done: [9bbdd3e](https://github.com/awslabs/tough/compare/3def6acea63efe7f322cf60fb534273e8294c5a5..9bbdd3ebc741d487744bced70d0f0d037cdc649c)

webern:
> webern force-pushed the webern:transport-develop branch from 9bbdd3e to df22150

Fix an incorrect comment

tough/src/lib.rs
webern: This is fixed [df22150](https://github.com/awslabs/tough/compare/9bbdd3ebc741d487744bced70d0f0d037cdc649c..df221502ffbbca28945411b88b2e35c777e017e7)

webern:
> webern force-pushed the webern:transport-develop branch from df22150 to 14439d9

Eliminate the `tough: add send trait to transport return` commit by squashing it (via fixup) into the `tough: remove type parameter for transport`. This makes more sense and also makes `simplify repository load` the last commit, which will make git a little easier for me as I work on the `RepositoryLoader`.

webern:
> webern force-pushed the webern:transport-develop branch from 14439d9 to 36d812d

Load repositories with a builder-like struct `RepositoryLoader`.

webern:
>  webern force-pushed the webern:transport-develop branch from 14439d9 to 36d812d

Fix a comment.

tough/src/lib.rs
```diff
@@ -94,45 +95,127 @@ impl From<ExpirationEnforcement> for bool {
...
+    pub fn limits(mut self, limits: Limits) -> Self {
+        self.limits = Some(limits);
+        self
+    }
 
-    /// Metadata expiration enforcement.
-    ///
-    /// When this parameter is set to `Unsafe`, a `Repository` will succeed in loading even if it
-    /// has expired metadata files.
+    /// Optionally set a `datastore` directory path. `datastore` is a directory on a persistent
```
tjkirch:
```suggestion
    /// Set a `datastore` directory path. `datastore` is a directory on a persistent
```
If you call this, you're setting it :)

tough/src/lib.rs
```diff
@@ -94,45 +95,127 @@ impl From<ExpirationEnforcement> for bool {
...
+    /// filesystem. This directory's contents store the most recently fetched timestamp, snapshot,
+    /// and targets metadata files to detect version rollback attacks.
     ///
-    /// CAUTION: TUF metadata expiration dates, particularly `timestamp.json`, are designed to
```
tjkirch, line 132: We seem to be losing some of the documentation on the settings?

tough/src/lib.rs
```diff
@@ -194,73 +284,60 @@ pub struct Repository<'a, T: Transport> {
...
-    pub fn load<R: Read>(transport: &'a T, settings: Settings<R>) -> Result<Self> {
-        let metadata_base_url = parse_url(settings.metadata_base_url)?;
-        let targets_base_url = parse_url(settings.targets_base_url)?;
-
-        let datastore = Datastore::new(settings.datastore)?;
+impl Repository {
+    /// Load and verify TUF repository metadata using a [`RepositoryLoader`] for the settings.
+    fn load<R: Read>(loader: RepositoryLoader<R>) -> Result<Self> {
+        // let metadata_base_url = parse_url(settings.metadata_base_url)?;
+        // let targets_base_url = parse_url(settings.targets_base_url)?;
```
tjkirch:
```suggestion
```

tough/src/lib.rs
```diff
@@ -400,18 +477,23 @@ fn check_expired<T: Role>(datastore: &Datastore, role: &T) -> Result<()> {
...
-    if !url.ends_with('/') {
-        url.to_mut().push('/');
+/// Checks to see if the `Url` has a trailing slash and adds one if not. Without a trailing slash,
+/// the last component of a `Url` is considered to be a file. `metadata_url` and `targets_url`
+/// must refer to a base (i.e. directory), so we need them to end with a slash.
+fn parse_url(url: Url) -> Result<Url> {
+    let mut s = url.to_string();
+    if s.as_str().ends_with('/') {
+        Ok(url)
+    } else {
```
tjkirch:
```suggestion
    if url.as_str().ends_with('/') {
        Ok(url)
    } else {
        let mut s = url.to_string();
```
No need to build the string if we're not going to append.

tuftool/src/common.rs
```diff
@@ -0,0 +1,35 @@
...
+///
+/// - `root` must be a path to a file that can be opened with `File::open`.
+/// - `metadata_url` can be local or remote.
+///
+pub(crate) fn load_metadata_repo<P>(root: P, metdata_url: Url) -> Result<Repository>
```
tjkirch:
```suggestion
pub(crate) fn load_metadata_repo<P>(root: P, metadata_url: Url) -> Result<Repository>
```

tjkirch (approved):
I don't want to lose the documentation on the settings, but otherwise this looks good to me, assuming testing is up to date.

zmrow (approved):
:national_park:

webern:
> force-pushed the webern:transport-develop branch from 231ceaa to 6ce9dc8

Fix cleanup issues. Next push I will rebase away the conflict.

tough/src/lib.rs
webern: Fixed in [6ce9dc8](https://github.com/awslabs/tough/compare/231ceaa8956d1fadd85c58e7f8202daff603bd5b..6ce9dc89c1658d2a3a11c9b6ef847cfe10f8a3fd)

tough/src/lib.rs
webern: I tried to preserve it all, I think this one was the only one that I lost. Fixed in [6ce9dc8](https://github.com/awslabs/tough/compare/231ceaa8956d1fadd85c58e7f8202daff603bd5b..6ce9dc89c1658d2a3a11c9b6ef847cfe10f8a3fd)

tough/src/lib.rs
webern: Fixed in [6ce9dc8](https://github.com/awslabs/tough/compare/231ceaa8956d1fadd85c58e7f8202daff603bd5b..6ce9dc89c1658d2a3a11c9b6ef847cfe10f8a3fd)

tough/src/lib.rs
webern: Fixed in [6ce9dc8](https://github.com/awslabs/tough/compare/231ceaa8956d1fadd85c58e7f8202daff603bd5b..6ce9dc89c1658d2a3a11c9b6ef847cfe10f8a3fd)

tuftool/src/common.rs
webern: Fixed in [6ce9dc8](https://github.com/awslabs/tough/compare/231ceaa8956d1fadd85c58e7f8202daff603bd5b..6ce9dc89c1658d2a3a11c9b6ef847cfe10f8a3fd)

webern:
> webern force-pushed the webern:transport-develop branch from 6ce9dc8 to 0842d87

Rebase with no diff. Deleted my clippy fix commit in favor develop.
I will now re-run the testing scenarios using 0842d87

webern:
Testing re-verified.

tjkirch (approved):

webern:
>  webern force-pushed the webern:transport-develop branch from 0842d87 to aeead79

rebase

tough/src/transport.rs
```diff
@@ -1,36 +1,172 @@
...
+/// To distinguish this case from other [`Transport`] failures, we use
+/// `TransportErrorKind::FileNotFound`.
+///
+#[derive(Debug, Copy, Clone)]
+#[non_exhaustive]
+pub enum TransportErrorKind {
+    /// The trait does not handle the URL scheme named in `String`. e.g. `file://` or `http://`.
+    UnsupportedUrlScheme,
+    /// The file cannot be found.
+    FileNotFound,
```
iliana, line 42:
```suggestion
    /// The file is not available.
    FileNotAvailable,
```

I feel like the current language implies the HTTP transport should only return this error kind on a 404, which is not what we'll be doing, because the production Bottlerocket repositories (currently) return a 403 for https://updates.bottlerocket.aws/2020-07-07/aws-k8s-1.17/x86_64/3.root.json, because that's what S3 does when the principal making the request does not have `s3:ListObjects` access.

The [TUF specification (v1.0.16)](https://github.com/theupdateframework/specification/blob/fd85a8a12403435d84e202ea8790eb4abad4559d/tuf-spec.md#5-detailed-workflows) is a bit vague here, simply saying "If this file is not available ... then go to step 5.2.9." Replicating the language here makes it clearer what an implementer should do.

As we implement a standard HTTP transport for tough, we have to decide which types of HTTP error codes count as "definitely not available" (e.g. 404, 403, maybe 410 Gone?) vs "could be available, but something is wrong" (e.g. 401 Unauthorized, 500 Internal Server Error), and ideally document that.

tough/src/transport.rs
```diff
@@ -1,36 +1,172 @@
...
+#[derive(Debug, Snafu)]
+#[snafu(visibility = "pub")]
+#[snafu(display("{:?} error fetching '{}': {}", kind, url, source))]
+pub struct TransportError {
+    /// The kind of error that occurred.
+    pub kind: TransportErrorKind,
+    /// The URL that the transport was trying to fetch.
+    pub url: String,
+    /// The underlying error that occurred.
+    pub source: Box<dyn std::error::Error + Send + Sync>,
```
iliana, line 57: Should this be an `Option`, instead of providing the `unsupported_scheme()` shorthand? Doesn't make sense to require an Error here for something like `TransportErrorKind::UnsupportedUrlScheme`, and there might be situations where another kind of error happens but the implementer doesn't have an error object to work with.

tough/src/transport.rs
```diff
@@ -1,36 +1,172 @@
...
+    UnsupportedUrlScheme,
+    /// The file cannot be found.
+    FileNotFound,
+    /// The transport failed for any other reason, e.g. IO error, HTTP broken pipe, etc.
+    Other,
+}
+
+/// The error type that [`Transport`] `fetch` returns.
+#[derive(Debug, Snafu)]
+#[snafu(visibility = "pub")]
```
iliana, line 49: Re-exporting the generated partial error structs from snafu makes part of our API based on a not-yet-1.0 library. I think manually implementing `std::error::Error` is a better, more-standard experience for library developers implementing `Transport`.

tough/src/transport.rs
```diff
@@ -1,36 +1,172 @@
...
+    /// Creates a new `DefaultTransport`. Same as `default()`.
+    pub fn new() -> Self {
+        Self::default()
+    }
+}
+
+#[cfg(feature = "http")]
+impl DefaultTransport {
+    /// Create a new `DefaultTransport` using the given HTTP `ClientSettings`.
+    #[allow(dead_code)]
```
iliana, line 138: What's the lint that gets hidden here?

tough/src/transport.rs
```diff
@@ -1,36 +1,172 @@
...
+///
+pub trait Transport: Debug + DynClone {
+    /// Opens a `Read` object for the file specified by `url`.
+    fn fetch(&self, url: Url) -> Result<Box<dyn Read + Send>, TransportError>;
+}
 
-    /// The type of error that the `fetch` function will return.
-    type Error: std::error::Error + Send + Sync + 'static;
+// Implement `Clone` for `Transport` trait objects.
+dyn_clone::clone_trait_object!(Transport);
```
iliana, line 20: If I'm reading `dyn_clone`'s documentation correctly, any implementations of Transport just need to implement `Clone` to meet this constraint, right? Can we make that clearer in the documentation?

tough/src/io.rs
```diff
@@ -6,15 +6,15 @@ use ring::digest::{Context, SHA256};
 use std::io::{self, Read};
 use url::Url;
 
-pub(crate) struct DigestAdapter<T> {
```
iliana, line 9: Is there a particular advantage to removing the type parameters on these implementation-detail types?

iliana (approved):
OK to merge as-is with one condition. Most of my concerns are around the redesign of the Transport trait, which is particularly complex and nuanced. Ultimately, I like the approach you took.

If we merge as-is, please open an issue to track removal of snafu's macro-generated `IntoError` structs for `TransportError` from the public interface of this library. A 1.0 tough cannot rely on a non-1.x snafu's implementation details there, and as a library it's not appropriate to strongly suggest or force the use of snafu on implementations of `Transport`.

tough/src/transport.rs
tjkirch: How about an [opaque struct](https://docs.rs/snafu/0.6.10/snafu/guide/opaque/index.html) that only implements a few traits and doesn't expose internals?

iliana: Opaque structs might solve the separate issue of our error variants being particularly purpose-specific, which makes stabilization more difficult.

The main concern I had when seeing this visibility parameter was the partial `IntoError` structs being part of the tough library interface. One thing I didn't notice is that `mod transport` isn't `pub`, so the partial structs here don't end up in the interface. But just to be on the safe side...

```suggestion
#[snafu(visibility = "pub(crate)")]
```

webern:
I've created fast-follow issues for everything. I will merge this as is, then fix the easy things. Then think more about TransportError.

Issue #262, opened 2020-11-09
tough-kms: some repos cannot be loaded after signing with kms


webern:
Occasionally, the KMS sign function produces a signature with a length shorter than the RSA key modulus used in signing. For example with an `RSA_3072` key, most KMS signatures will be 384 bytes long. But approximately 1 out of 200 of signatures will be shorter than that.

Ring has a check to ensure that the length of the signature being verified exactly matches the length of the key modulus. This means that 1/200 KMS signings produces a signature that ring verify will reject.

Both KMS and `openssl pkeyutl` assert that these 'short' signatures are valid. To ensure that all of ring, KMS and openssl agree, we should prepend signatures returned from KMS that are shorter than the modulus length with zeros until the modulus length is achieved.

webern:
This has been fixed in #263

PR #263, opened 2020-11-09
tough-kms: pad short signatures

webern:
*Issue #, if available:*

#262

*Description of changes:*

```
This is a critical fix that prevents tough from creating repos that it
cannot load when using tough-kms to sign the repo.

Sometimes KMS returns a signature that is shorter than the key modulus.
Both KMS and OpenSSL will validate such signatures, but ring requires
that the signature length exactly match the modulus. If we pad such
signatures by adding zeros to the beginning until the signature length
matches the modulus length, then we can ensure ring and any other
implementations with the length requirement will be able to validate
the signature.
```

*Testing*

 - [x] I ran several thousand [loops](https://github.com/webern/tough/blob/1950bc7768350f06778e62735226c0ac38846e61/tuftool/tests/create_repository_integration.rs#L249-L286) of `create repo` and `load repo` with KMS signing. Before the fix these were failing approximately one out of every 200 iterations. After the fix I saw no failures over thousands of iterations.
 - [x] Check padded signatures. I [captured](https://github.com/webern/tough/blob/1950bc7768350f06778e62735226c0ac38846e61/tough-kms/src/lib.rs#L267) two signatures after the padding function had added one or more bytes. I verified these padded signatures with:
   - [x] openssl
   - [x] KMS
 - [x] I ran `tuftool` tests with `--features integ` which enables the regular kms tests.


tough-kms/src/lib.rs
```diff
@@ -188,7 +196,130 @@ impl Sign for KmsRsaKey {
...
+        },
+    );
+    if signature.len() == modulus_size_bytes {
+        return Ok(signature);
     }
+    // we now know that the signature is shorter than the modulus
+    let padding_size: usize = modulus_size_bytes - signature.len();
+    let mut padded: Vec<u8> = vec![0; modulus_size_bytes];
+    padded.splice(padding_size..modulus_size_bytes, signature.iter().cloned());
+    Ok(padded)
```
tjkirch:
```suggestion
    signature.splice(..0, [0].repeat(padding_size));
    Ok(signature)
```
We have ownership of `signature`, so this seems simpler..?

tough-kms/src/lib.rs
```diff
@@ -188,7 +196,130 @@ impl Sign for KmsRsaKey {
...
+    }
+}
+
+/// Parses the `CustomerMasterKeySpec` string returned by KMS, e.g. `RSA_3072` and returns the size
+/// of the modulus in bytes. For example `RSA_3072` has a modulus of 3072 bits, so the function will
+/// return 384 == (3072 / 8). If the parsed number is not divisible by 8, an error is returned.
+fn parse_modulus_length_bytes(spec: &str) -> error::Result<usize> {
+    // only RSA is currently supported
+    ensure!(
+        spec.starts_with("RSA_"),
```
tjkirch, line 223: Are we not sure whether the other types would be OK with the same padding strategy?  I guess it doesn't matter since we only support one algorithm and RSA keys?

(I was initially thinking this would make it a breaking release, but it has to be a breaking release anyway due to the change to `KmsRsaKey`...)

tjkirch (approved):

tough-kms/src/lib.rs
```diff
@@ -188,7 +196,130 @@ impl Sign for KmsRsaKey {
...
+    );
+    // prevent a panic if the string is precisely "RSA_"
+    ensure!(spec.len() > 4, error::BadCustomerMasterKeySpec { spec });
+    // extract the digits
+    let mod_len_str = &spec[4..];
+    // parse the digits
+    let mod_bits = mod_len_str
+        .parse::<usize>()
+        .context(error::BadCustomerMasterKeySpecInt { spec })?;
+    // make sure the modulus size is compatible with u8 bytes
```
srgothi92, line 234: Do you think instead of parsing `customer_master_key_spec` we can use enum that maps `customer_master_key_spec` to number of bytes. That way we won't need to parse string and can avoid unsupported customer_master_key_spec from being used.

tough-kms/src/lib.rs
```diff
@@ -188,7 +196,130 @@ impl Sign for KmsRsaKey {
...
+
+/// Parses the `CustomerMasterKeySpec` string returned by KMS, e.g. `RSA_3072` and returns the size
+/// of the modulus in bytes. For example `RSA_3072` has a modulus of 3072 bits, so the function will
+/// return 384 == (3072 / 8). If the parsed number is not divisible by 8, an error is returned.
+fn parse_modulus_length_bytes(spec: &str) -> error::Result<usize> {
+    // only RSA is currently supported
+    ensure!(
+        spec.starts_with("RSA_"),
+        error::BadCustomerMasterKeySpec { spec }
+    );
```
srgothi92, line 225: Instead of check for RSA prefix here do you think we can call this method only when signing algorithm is `KmsSigningAlgorithm.RsassaPssSha256` that way when we plan to support additional algorithm developer can handle different algorithm in `tuf_key` method.

srgothi92 (commented):
For some strange reason my review comment dint come as inline comments and were shown as PR comments. Adding them back as inline comments.

webern:
> webern force-pushed the webern:pad-kms-signatures branch from 2a3264b to b47c414

Fixes the test regression and uses @tjkirch's improved vec splice.

tough-kms/src/lib.rs
webern: The change seems specific to RSA, but I honestly don't know for sure. Elsewhere @srgothi92 recommends only running the function if the key type variant matches.

The fields of the KmsRsaKey struct are private, so the programmer interface is unchanged. It seems like it could be a patch release?

tough-kms/src/lib.rs
webern: I think I would prefer to be agnostic about the bitsize, provided it is divisible by 8, rather than maintaining a variant. That way if KMS adds support for a size, it will still 'just work', instead of falsely reporting that the bit size is unsupported.

tough-kms/src/lib.rs
webern: I could call it only when the algorithm is `KmsSigningAlgorithm.RsassaPssSha256` (looking into this now). But even so, I need to raise an error if the string isn't what I expect it to be. Even if we went with a variant as you suggest above, we'd be raising an error for unexpected strings here.

webern: It looks pretty ugly to me, especially since this is happening inside of a function that is a member of `KmsRsaKey`, so by definition it will only happen for RSA algorithms. Also the algorithm is a string.

```rust
        // sometimes KMS produces an RMS signature that is shorter than the modulus. in those cases,
        // we have observed that openssl and KMS will both validate the signature, but ring will
        // not. if we pad the beginning of the signature with zeros to make the signature exactly
        // the same length as the modulus, then ring will verify the signature.
        let signature = match self.signing_algorithm.as_str() {
            "I don't know exactly what this is right now" => {
                pad_signature(signature, self.modulus_size_bytes)?
            }
            _ => signature,
        };
```

tough-kms/src/lib.rs
tjkirch: Ah, I missed that the fields are private, no problem there then.  Do we only allow `RSA_*` because we only support `RsassaPssSha256`?  If that's true, I think there's no change in support.

etungsten (approved):

tough-kms/src/lib.rs
srgothi92: I feel it is a better check then checking for `RSA` prefix which is some kind of naming convention followed by KMS. I assume this check would be added for `parse_modulus_length_bytes` in tuf_key instead of `pad_signature`. Additionally it  helps us to throw error if user provided key type is not supported by us. For string comparison we can use `KmsSigningAlgorithm::RsassaPssSha256.value()`

webern:
> webern force-pushed the webern:pad-kms-signatures branch from b47c414 to ffd3933

restores a small comment that was lost because because git is hard

tough-kms/src/lib.rs
srgothi92: Another thought can we use signing algorithm: `RSASSA_PSS_SHA_256 | RSASSA_PSS_SHA_384 | RSASSA_PSS_SHA_512 | RSASSA_PKCS1_V1_5_SHA_256 | RSASSA_PKCS1_V1_5_SHA_384 | RSASSA_PKCS1_V1_5_SHA_512 | ECDSA_SHA_256 | ECDSA_SHA_384 | ECDSA_SHA_512`
to extract bit size because it has consistent naming for different algo, instead of `CustomerMasterKeySpec` which has `RSA_2048 | RSA_3072 | RSA_4096 | ECC_NIST_P256 | ECC_NIST_P384 | ECC_NIST_P521 | ECC_SECG_P256K1 | SYMMETRIC_DEFAULT` where bit size naming is not consistent. Either way feels bad to me :(, because we are dependent on a naming convention. I searched if there is a way to identify kms key bit size, but couldn't find anything.

Or modify `KmsSigningAlgorithm.RsassaPssSha256` enum to provide signing_algorithm string and bit size.

tough-kms/src/lib.rs
webern: > Do we only allow `RSA_*` because we only support `RsassaPssSha256`?

Correct, we only support `RsassaPssSha256`.

tough-kms/src/lib.rs
webern: The algorithm is independent of the size of the key modulus. e.g. RSASSA_PSS_SHA_256 works with 2048, 3072, etc. `CustomerMasterKeySpec` is the only place I know of that we can get the key size. I'm not sure how the ring algorithm figures it out, I guess it must be encoded in the x509_spki representation of the public key, but that seems like a more difficult place to parse it from.

tough-kms/src/lib.rs
webern: I think the `KmsRsaKey` should hold a `KmsSigningAlgorithm` instead of a string. Do you remember why you made it a string?

```rust
/// Implements the Sign trait for KMS rsa Key
pub struct KmsRsaKey {
    /// Key Id of Customer Managed Key in KMS used to sign the message
    key_id: String,
    /// Aws account profile
    profile: Option<String>,
    /// KmsClient Object to query AWS KMS
    client: Option<KmsClient>,
    /// Public Key corresponding to Customer Managed Key
    public_key: Decoded<RsaPem>,
    /// Signing Algorithm to be used for the Customer Managed Key
    signing_algorithm: KmsSigningAlgorithm, // currently String
    /// The size of the RSA key modulus in bytes.
    modulus_size_bytes: usize,
}
```

webern: Seems like a tiny, non-breaking change to hold the variant instead of a string there, which then makes checking the algorithm before using the padding function much cleaner. I will push a version that does this.

webern:
>  webern force-pushed the webern:pad-kms-signatures branch from ffd3933 to 0124b97

Change `KmsRsaKey` to hold a `KmsSigningAlgorithm` instead of its `String` representation and only run the padding function if the alg is `RsassaPssSha256`. For the moment it looks silly because there is only one algorithm in enum, but if any future algorithm is added, the compiler will force the developer to think about whether or not they want to run the padding function.

tough-kms/src/lib.rs
srgothi92: the `signing_algorithm` type is already `KmsSigningAlgorithm` https://github.com/awslabs/tough/blob/develop/tough-kms/src/lib.rs#L63. Not, sure how you saw it as a string.

webern: Take a look at [0124b97](https://github.com/awslabs/tough/compare/ffd3933d0f0b5c4ab1aeaa888eb0aa1c9374c2d5..0124b97dfc0db514c1d4fb463ca7315429933586)

srgothi92: Ignore the above comment. You are correct.

srgothi92 (approved):
LGTM. 🙌

webern:
Testing is complete, paging @iliana (figuratively)

zmrow (approved):
:nut_and_bolt:

iliana:
> Testing is complete, paging @iliana (figuratively)

I wish :pager: was a reaction emoji now

iliana (approved):

bcressey (approved):

PR #288, opened 2020-12-11
change expires < now to expires <= now

### LINKED ISSUE #287: Update expiration comparison from expires < now to expires <= now
There was a discussion in the issue tracker of TUF reference implementation (https://github.com/theupdateframework/tuf/issues/1231) about how expiration datetime's should be compared, in light of a recent rephrasing of the freeze attack checks in [TUF 1.0.16](https://github.com/theupdateframework/specification/blob/v1.0.16/tuf-spec.md):

> The expiration timestamp in the [ROLE] metadata file MUST be higher than the fixed update start time.

Having done some research into how other security providers are comparing expiration equivalents (i.e. OpenSSL x509 certificate checking code, and GnuPG expiration checks – see theupdateframework/tuf#1231 for more details ), and how other TUF implementations are performing the same check (rust-tuf, go-tuf), we came to a rough consensus that the correct way to implement expiration comparisons is:

> expires <= now

When reviewing other implementations I discovered (and @iliana [confirmed](https://github.com/theupdateframework/tuf/issues/1231#issuecomment-738668794)) that Tough currently checks that `expires < now`, therefore I volunteered to create this bug.

webern:
*Issue #, if available:*

Closes #287
Discussed at https://github.com/theupdateframework/tuf/issues/1231

*Description of changes:*

A rephrasing requires freeze attach checks to ensure that the current
time is less than or equal to the metadata expiration time.

*Testing*

I tried to create a test that checked for the edge case (sys_time == expiration), but I don't think it's possible. As such, I'm relying on the existing tests.

### DIFF
--- a/tough/src/lib.rs
+++ b/tough/src/lib.rs
@@ -392,9 +392,11 @@ fn system_time(datastore: &Datastore) -> Result<DateTime<Utc>> {
     Ok(sys_time)
 }
 
+/// TUF v1.0.16, 5.2.9, 5.3.3, 5.4.5, 5.5.4, The expiration timestamp in the `[metadata]` file MUST
+/// be higher than the fixed update start time.
 fn check_expired<T: Role>(datastore: &Datastore, role: &T) -> Result<()> {
     ensure!(
-        system_time(datastore)? < role.expires(),
+        system_time(datastore)? <= role.expires(),
         error::ExpiredMetadata { role: T::TYPE }
     );
     Ok(())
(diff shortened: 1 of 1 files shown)


tjkirch (approved):
I'd like to make a dramatic case here that the field/function name "expires" implies "expires at" and not "expires after", but in the end, we should probably do what upstream does.

nit: typo in the commit, "attach" instead of "attack"

iliana (approved):

PR #296, opened 2020-12-18
add makefile so local ci check is easy

### LINKED ISSUE #266: replicate ci locally with script or make
Whenever I work on the project, I end up doing a find and replace of the github actions file to produce a script that does the same locally so that I can run all the CI stuff locally instead of waiting for CI to tell me what clippy wants or whatever.

We could add a script or a makefile to make that easier. Downside: it would have to be maintained along with the GitHub actions yaml since it would likely *not* be used by the CI.

webern:
*Issue #, if available:*

Closes #266

*Description of changes:*

```
Add a Makefile that does all of the continuous integration builds and
checks. Then call that Makefile target in the actual ci run. This will
be helpful when developers want to check that ci will pass locally
before pushing changes to a pull request.
```

This will significantly improve my workflow experience!

*Testing*

It worked for me locally and we'll see if it works on the GitHub runner.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -8,13 +8,4 @@ jobs:
     steps:
     - uses: actions/checkout@v1
     - run: rustup update stable
-    - run: cargo fmt -- --check
-    - run: cargo clippy --locked -- -D warnings
-    - run: cargo build --locked -p olpc-cjson
-    - run: cargo build --locked -p tough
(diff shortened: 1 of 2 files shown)


srgothi92 (approved):
Nice. ❤️

tjkirch (approved):

zmrow (approved):

PR #308, opened 2020-12-29
tough: improve transport error

### LINKED ISSUE #294: Figure out what to do about 403's wrt to TransportError
From #256 are grouped here:

Regarding the use of FileNotFound for 403 errors, @iliana said:

> I feel like the current language implies the HTTP transport should only return this error kind on a 404, which is not what we'll be doing, because the production Bottlerocket repositories (currently) return a 403 for https://updates.bottlerocket.aws/2020-07-07/aws-k8s-1.17/x86_64/3.root.json, because that's what S3 does when the principal making the request does not have s3:ListObjects access.
>
> The TUF specification (v1.0.16) is a bit vague here, simply saying "If this file is not available ... then go to step 5.2.9." Replicating the language here makes it clearer what an implementer should do.
>
> As we implement a standard HTTP transport for tough, we have to decide which types of HTTP error codes count as "definitely not available" (e.g. 404, 403, maybe 410 Gone?) vs "could be available, but something is wrong" (e.g. 401 Unauthorized, 500 Internal Server Error), and ideally document that.

I'll have to check the TUF spec on this. It's unfortunate for our use case that S3 returns a 403 for non-existent objects. I think I understand why, but it's too bad for us.

### LINKED ISSUE #303: tough: Make underlying error optional in TransportError
https://github.com/awslabs/tough/pull/256#discussion_r545457745

webern:
*Issue #, if available:*

Closes #294
Closes #303

*Description of changes:*

The first commit: 

```
Make the inner error of the TransportError optional. Two public
functions are provided for constructing a TransportError. This is better
than a function that takes an Option<E> for the optional inner error
because the user would need to specify a type when trying to pass None.

Expand documentation of the TransportErrorKind::FileNotFound variant to
better explain the intent.

Hide the members of the TransportError struct to make breaking changes
less likely. Instead make them accessible by getter functions.

Changing the inner error to an Option made Snafu's macros less useful
because it gives special meaning to a struct member named 'source', and
would not compile when this became an Option. I wanted to keep the
member name because it seems to be canonical in Rust. Additionally, the
display macro became difficult to use to produce a clean error message.
It seemed easier to implement Error and Display myself.
```

The second commit

```
Provide documentation about what HTTP codes constitute FileNotFound for
the HTTP transport and add 410 to the list.
```

The third commit:

I briefly considered making the `FileNotFound` codes configurable for the `HttpTransport`, but I think YAGNI. Also, after a bit of searching, I think it is a best practice for HTTP servers to obscure the difference between 404 and 403, so I think it is correct for the transport to treat them as equivalent.

However, the idea of adding a setting got me thinking about breaking changes. If we use a builder instead of a settings struct, we set ourselves up better for preventing future breaking changes when a new setting is introduced.

```
Replaces the http ClientSettings struct with a builder,
HttpTransportBuilder. By introducing a builder we can hide settings
details and prevent future breaking changes (e.g. if we need to add a
setting).
```

*Testing*

I didn't do any testing beyond running tests locally and in CI. No algorithms were harmed during the making of this PR.

### DIFF
--- a/tough/src/http.rs
+++ b/tough/src/http.rs
@@ -12,25 +12,29 @@ use std::io::Read;
 use std::time::Duration;
 use url::Url;
 
-/// Settings for the HTTP client including retry strategy and timeouts.
+/// A builder for [`HttpTransport`] which allows settings customization.
+///
+/// # Example
+///
+/// ```
+/// # use tough::HttpTransportBuilder;
+/// let http_transport = HttpTransportBuilder::new()
+/// .tries(3)
+/// .backoff_factor(1.5)
+/// .build();
+/// ```
+///
 #[derive(Clone, Copy, Debug)]
-pub struct ClientSettings {
-    /// Set a timeout for connect, read and write operations.
-    pub timeout: Duration,
-    /// Set a timeout for only the connect phase.
-    pub connect_timeout: Duration,
-    /// The total number of times we will try to get the response.
-    pub tries: u32,
-    /// The pause between the first and second try.
-    pub initial_backoff: Duration,
-    /// The maximum length of a pause between retries.
-    pub max_backoff: Duration,
-    /// The exponential backoff factor, the factor by which the pause time will increase after each
-    /// try until reaching `max_backoff`.
-    pub backoff_factor: f32,
+pub struct HttpTransportBuilder {
+    timeout: Duration,
+    connect_timeout: Duration,
+    tries: u32,
+    initial_backoff: Duration,
+    max_backoff: Duration,
+    backoff_factor: f32,
 }
 
-impl Default for ClientSettings {
+impl Default for HttpTransportBuilder {
     fn default() -> Self {
         Self {
             timeout: std::time::Duration::from_secs(30),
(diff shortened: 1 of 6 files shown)


webern:
I got clippied because I don't have my Makefile from here #296!

tough/src/transport.rs
```diff
@@ -25,61 +26,97 @@ dyn_clone::clone_trait_object!(Transport);
...
+    /// or the machine hosting the file).
+    ///
+    /// For some transports, the distinction is obvious. For example, a local file transport should
+    /// return `FileNotFound` for `std::error::ErrorKind::NotFound` and nothing else. For other
+    /// transports it might be less obvious, but the intent of `FileNotFound` is to indicate that
+    /// the file probably doesn't exist.
     FileNotFound,
-    /// The transport failed for any other reason, e.g. IO error, HTTP broken pipe, etc.
+    /// The transport failed for any other reason, e.g. IO error, HTTP broken pipe, etc. The
+    /// implication is that the [`Transport`] the file might exist, but the [`Transport`] failed.
```
webern: FIXME
```rust
    /// implication is that the [`Transport`] the file might exist, but the [`Transport`] failed.
```

tough/src/http.rs
```diff
@@ -12,25 +12,29 @@ use std::io::Read;
...
+    connect_timeout: Duration,
+    tries: u32,
+    initial_backoff: Duration,
+    max_backoff: Duration,
+    backoff_factor: f32,
 }
 
-impl Default for ClientSettings {
+impl Default for HttpTransportBuilder {
     fn default() -> Self {
```
tjkirch, line 38: It won't let me leave a suggestion on a line that didn't change, but the comment below should probably say `225ms` instead of `220ms`, being the 1.5x backoff of 150ms.

tough/src/http.rs
```diff
@@ -44,22 +48,66 @@ impl Default for ClientSettings {
...
+    }
+
+    /// Construct an [`HttpTransport`] transport from this builder's settings.
+    pub fn build(self) -> HttpTransport {
+        HttpTransport { settings: self }
+    }
+}
+
+/// A [`Transport`] over HTTP with retry logic. Use the [`HttpTransportBuilder`] to construct an
+/// `HttpTransport`.
```
tjkirch:
```suggestion
/// A [`Transport`] over HTTP with retry logic. Use the [`HttpTransportBuilder`] to construct a
/// custom `HttpTransport`, or use `HttpTransport::default()`.
```

tough/tests/http.rs
```diff
@@ -25,7 +25,7 @@ mod http_happy {
     /// Test that `tough` works with a healthy HTTP server.
     #[test]
     fn test_http_transport_happy_case() {
-        run_http_test(HttpTransport::new());
+        run_http_test(HttpTransportBuilder::new().build());
```
tjkirch:
```suggestion
        run_http_test(HttpTransport::default());
```
nit: then we wouldn't need to import Builder either.

tough/src/transport.rs
```diff
@@ -1,8 +1,9 @@
 #[cfg(feature = "http")]
-use crate::{ClientSettings, HttpTransport};
+use crate::{HttpTransport, HttpTransportBuilder};
 use dyn_clone::DynClone;
-use snafu::Snafu;
-use std::fmt::Debug;
+use serde::export::Formatter;
+use std::error::Error;
+use std::fmt::{Debug, Display};
```
tjkirch:
```suggestion
use std::error::Error;
use std::fmt::{Debug, Display, Formatter};
```
Serde says `export` is `Not public API.` even though it's marked `pub`.  I think it's just a reexport of `std::fmt::Formatter`, though?

tough/src/transport.rs
tjkirch: I'd suggest removing this new sentence; I don't think the Transport will always know whether the file exists (IO error before the attempt, for example) and it doesn't seem worth introducing further complexity to distinguish those.

tough/src/transport.rs
```diff
@@ -25,61 +26,97 @@ dyn_clone::clone_trait_object!(Transport);
...
+impl Display for TransportError {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        if let Some(e) = self.source.as_ref() {
+            write!(
+                f,
+                "Transport {:?} error fetching '{}': {}",
+                self.kind, self.url, e
+            )
+        } else {
+            write!(f, "Transport {:?} error fetching '{}'", self.kind, self.url)
```
iliana: IMO would be better to `impl Display for TransportErrorKind` and use that friendlier version here.

iliana (approved):

tough/src/transport.rs
webern: oh my IDE failed to choose the right thing.

webern:
> webern force-pushed the webern:transport-error-optional branch from 2515aca to 93889cf now

A rebase to get me the Makefile of happiness.

webern:
> webern force-pushed the webern:transport-error-optional branch from 93889cf to 38aa258 2 minutes ago

I believe this fixes all of the PR comments.

tjkirch (approved):

Issue #13, opened 2019-11-25
tuftool's native-tls/rustls switch doesn't affect reqwest


iliana:
We have:

https://github.com/awslabs/tough/blob/9ac95c8aaf94b816393b578af247edf1af958c44/tuftool/Cargo.toml#L11-L15

but we also depend on reqwest, and we aren't switching which TLS implementation it uses.

Ideally I'd like to be able to enable/disable `rusoto_ssm` and switch between `native-tls` and `rustls` separately but I don't know if that's possible.

webern:
I opened this, which seems related: https://github.com/awslabs/tough/issues/271

webern:
Actually... does `tuftool` need to give the user this configurability? Since it is a command-line tool and will always be compiled in isolation, what is the benefit of giving a user the ability to choose `native-tls` (assuming we chose `rustls`)?

iliana:
Honestly I'd go the other direction and use native-tls in tuftool, if we're going to make the choice for the user.

webern:
Moving this out of the v0.10.0 milestone as the v0.10.0 diff is already huge and releasing it sooner rather than later moves things in the right direction.

PR #318, opened 2021-01-06
tough: add tls features

### LINKED ISSUE #271: tough: the http feature does not include tls
The http feature doesn't bring TLS into scope, leading to this kinda hacky comment https://github.com/bottlerocket-os/bottlerocket/blob/55b584b0d872282190bfcb0d2d917c6b58891cda/tools/pubsys/Cargo.toml#L20

What we could probably do is have:

 - feature `http` bring in rusoto with rustls
 - feature `http-native-tls` bring in rusoto with native-tls
 - feature `http-no-tls` bring in rusoto without a tls implementation.

webern:
*Issue #, if available:*

Closes #271 
Does not address #13, I will work on that separately.

*Description of changes:*

```
Before this commit, there was no way to enable TLS for the HTTP
transport. The only way to do so was to manually include the same
version of reqwest and enable TLS on reqwest directly.

Here we keep the behavior we already had with the http feature (no TLS),
but we add two new features for the two TLS implementations.
```

**Testing**

I created a simple program referencing tough locally. The program loaded the Bottlerocket repo. With the `http` feature it failed (since the URL is https).  Both `http-rustls-tls` and `http-native-tls` worked.

I downloaded the repo with tuftool, and that was working regardless of whether tuftools Cargo.toml specifies tough with `features = ["http"]`, or `http-rustls-tls` or `http-native-tls`, so I left tuftool's Cargo.toml unchanged.

I also ran these:
```
    cd tough && cargo test --features 'integ,http' --locked
    cd tough && cargo test --features 'integ,http-rustls-tls' --locked
    cd tough && cargo test --features 'integ,http-native-tls' --locked
```
And all passed. I did not add these to the Makefile or CI yet because I don't want to make CI slower.

### DIFF
--- a/tough/tests/http.rs
+++ b/tough/tests/http.rs
@@ -1,7 +1,10 @@
 mod test_utils;
 
-/// Instead of guarding every individual thing with `#[cfg(feature = "http")]`, use a module.
-#[cfg(feature = "http")]
+#[cfg(any(
+    feature = "http",
+    feature = "http-rustls-tls",
+    feature = "http-native-tls"
+))]
 mod http_happy {
     use crate::test_utils::{read_to_end, test_data};
     use mockito::mock;
@@ -83,7 +86,11 @@ mod http_happy {
     }
 }
 
-#[cfg(feature = "http")]
+#[cfg(any(
+    feature = "http",
+    feature = "http-rustls-tls",
+    feature = "http-native-tls"
+))]
 #[cfg(feature = "integ")]
 mod http_integ {
     use crate::test_utils::test_data;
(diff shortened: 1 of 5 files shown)


tough/tests/http.rs
```diff
@@ -1,7 +1,11 @@
 mod test_utils;
 
 /// Instead of guarding every individual thing with `#[cfg(feature = "http")]`, use a module.
```
srgothi92: I guess this comment also needs modification.

srgothi92 (approved):

tough/tests/http.rs
webern: This was never a very useful comment so I deleted it.

webern:
Needs more thought, closing.

PR #327, opened 2021-01-15
Tuftool 0.6.0

webern:
Prep releases of tough-kms, tough-ssm and tuftool.

### DIFF
--- a/tuftool/CHANGELOG.md
+++ b/tuftool/CHANGELOG.md
@@ -4,20 +4,26 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]
+## [0.6.0] - 2021-01-19
+### Breaking Changes
+- Correct spelling of `tuftool download` argument from `--target-url` to `--targets-url` [#309]
+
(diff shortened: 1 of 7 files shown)


tuftool/CHANGELOG.md
```diff
@@ -93,7 +99,8 @@ Major update: much of the logic in `tuftool` has been factored out and added to
 ### Added
 - Everything!
 
-[Unreleased]: https://github.com/awslabs/tough/compare/tuftool-v0.5.0...develop
+[Unreleased]: https://github.com/awslabs/tough/compare/tuftool-v0.6.0...develop
```
srgothi92, line 102: Why do we add [Unreleased] in `tuftool` but not for `tough-kms` and `tough-ssm` ? Is it because the may not get any new changes ?

srgothi92 (approved):

zmrow (approved):
:safety_vest:

tuftool/CHANGELOG.md
webern: I don't think there's a well-reasoned purpose for the discrepancy. This line is invisible in the markdown rendering, but is convenient if we choose to add a `[Unreleased]` section back to the top for collecting unreleased changes over time. I think your observation that we are less likely to make such a section in `tough-kms` or `tough-ssm` is as good an explanation as any.

iliana (approved):

PR #342, opened 2021-02-08
Windows fix [ Partial / 'tough' crate only ]

Cytro54:
*Issue #, if available:*
"tough" crate doesn't compile on Windows

*Description of changes:*
 - Replaces the `symlink` function with the `symlink_file` when needed
 - Patches the FileSystemTransport to patch a bug with 'url.path()' that returned an invalid path.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,16 @@
 
 **tuftool** is a Rust command-line utility for generating and signing TUF repositories.
 
+## Integration Testing
+Integration tests require `docker`.
+
+### Windows❗ Warnings
+- Tests can break on Windows if Git's `autocrlf` feature changes line endings.
+  This is due to the fact that some tests require files to have a *precise* byte size and hash signature.
+  *We have mitigated this with a `.gitattributes` file in the test data directory*.
+
+- Cygwin **must** be installed at `C:\cygwin64\` and have the `make` package installed for integration tests to work properly. 
+
 ## Documentation
 See [tough - Rust](https://docs.rs/tough/) for the latest `tough` library documentation.
 
--- a/integ/failure-server/run.sh
+++ b/integ/failure-server/run.sh
@@ -14,6 +14,14 @@ set -eo pipefail
 DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
 TUF_REFERENCE_REPO="${DIR}/../../tough/tests/data/tuf-reference-impl"
 
+# if we are under cygwin , set a windows style path
+# (docker for windows requires windows style paths to work ,it doesn't understand cygwin-style paths)
+if [[ "$OSTYPE" == "cygwin" ]]; then
+  DIR="$(cygpath --windows ${DIR})"
+  TUF_REFERENCE_REPO="$(cygpath --windows ${TUF_REFERENCE_REPO})"
+fi
+
+
 function waitforit() {
   echo "waiting $1 seconds for $2 to start"
   sleep $1
--- a/tough/src/editor/signed.rs
+++ b/tough/src/editor/signed.rs
@@ -21,7 +21,12 @@ use serde_plain::forward_from_str_to_serde;
 use snafu::{ensure, OptionExt, ResultExt};
 use std::collections::HashMap;
 use std::fs;
+
+#[cfg(not(target_os = "windows"))]
 use std::os::unix::fs::symlink;
+#[cfg(target_os = "windows")]
+use std::os::windows::fs::symlink_file as symlink;
+
 use std::path::{Path, PathBuf};
 use url::Url;
 use walkdir::WalkDir;
--- a/tough/src/transport.rs
+++ b/tough/src/transport.rs
@@ -140,6 +140,7 @@ pub struct FilesystemTransport;
 
 impl Transport for FilesystemTransport {
     fn fetch(&self, url: Url) -> Result<Box<dyn Read + Send>, TransportError> {
+        // If the scheme isn't "file://", reject
         if url.scheme() != "file" {
             return Err(TransportError::new(
                 TransportErrorKind::UnsupportedUrlScheme,
@@ -147,7 +148,17 @@ impl Transport for FilesystemTransport {
             ));
         }
 
-        let f = std::fs::File::open(url.path()).map_err(|e| {
+        // Convert the file URL into a file path
+        let file_path = &url.to_file_path().map_err(|_e| {
+            TransportError::new_with_cause(
+                TransportErrorKind::Other,
+                &url,
+                "unable to get filepath from URL".to_string(),
+            )
+        })?;
+
+        // And open the file
+        let f = std::fs::File::open(file_path).map_err(|e| {
             let kind = match e.kind() {
                 ErrorKind::NotFound => TransportErrorKind::FileNotFound,
                 _ => TransportErrorKind::Other,
--- a/tough/tests/data/.gitattributes
+++ b/tough/tests/data/.gitattributes
@@ -1,0 +1,2 @@
+# Denote all files as binary so they will not be modified by autocrlf.
+* binary
--- a/tough/tests/http.rs
+++ b/tough/tests/http.rs
@@ -96,7 +96,20 @@ mod http_integ {
     pub fn integ_dir() -> PathBuf {
         let mut p = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
         p.pop();
-        p.join("integ").canonicalize().unwrap()
+        p = p.join("integ");
+        p
+    }
+
+    /// Returns a command object that runs the provided script under BASH , wether we are under cygwin or unix.
+    pub fn bash_base() -> Command {
+        // if under cygwin, run the bash script under cygwin64 bash
+        if cfg!(target_os = "windows") {
+            let mut command = Command::new("c:\\cygwin64\\bin\\bash");
+            command.arg("-l");
+            return command;
+        } else {
+            return Command::new("bash");
+        }
     }
 
     pub fn tuf_reference_impl() -> PathBuf {
@@ -121,18 +134,25 @@ mod http_integ {
     /// triggering retries in the `fetch` loop.
     #[test]
     fn test_retries() {
+        use std::ffi::OsString;
         // run docker images to create a faulty http representation of tuf-reference-impl
-        let output = Command::new("bash")
-            .arg(
-                integ_dir()
-                    .join("failure-server")
-                    .join("run.sh")
-                    .into_os_string(),
-            )
+
+        // Get the "run.sh" path
+        let script_path = integ_dir()
+            .join("failure-server")
+            .join("run.sh")
+            .into_os_string()
+            .into_string()
+            .unwrap();
+
+        // Run it under BASH
+        let output = bash_base()
+            .arg(OsString::from(script_path))
             .stdout(Stdio::inherit())
             .stderr(Stdio::inherit())
             .output()
             .expect("failed to start server with docker containers");
+
         if !output.status.success() {
             panic!("Failed to run integration test HTTP servers, is docker running?");
         }
@@ -162,7 +182,7 @@ mod http_integ {
         }
 
         // stop and delete the docker containers, images and network
-        let output = Command::new("bash")
+        let output = bash_base()
             .arg(
                 integ_dir()
                     .join("failure-server")
--- a/integ/failure-server/.gitattributes
+++ b/integ/failure-server/.gitattributes
@@ -1,0 +1,4 @@
+# Denote bash scripts as binary so they will not be modified by autocrlf.
+# (Bash doens't seem to like CRLF-encoded scripts | Just to be sure)
+run.sh binary
+teardown.sh binary


gregdek:
Ping @webern @iliana since I can't set official reviews on this repo :)

webern:
@Cytro54 thank you for this PR. We may end up requesting some changes but we're excited to add Windows support!

tough/src/editor/signed.rs
```diff
@@ -21,7 +21,12 @@ use serde_plain::forward_from_str_to_serde;
 use snafu::{ensure, OptionExt, ResultExt};
 use std::collections::HashMap;
 use std::fs;
+
+#[cfg(not(target_os = "windows"))]
 use std::os::unix::fs::symlink;
+#[cfg(target_os = "windows")]
+use std::os::windows::fs::symlink_file;
```
webern:
```suggestion
use std::os::windows::fs::symlink_file as symlink;
```



Since these two versions of the functions have the same signature, we can rename the `symlink_file` to `symlink` which I believe will eliminate the need to have `cfg` statements elsewhere in this file.

Could you try making this change and undoing the other changes in this file to see if it works?

make.ps1
```diff
@@ -0,0 +1,11 @@
+
```
webern: I think it would be better if we stick to one entrypoint (the `Makefile`) instead of introducing a script here. Maybe you don't have make installed? I think the makefile would work as-is.

tough/src/transport.rs
```diff
@@ -156,6 +157,39 @@ impl Transport for FilesystemTransport {
...
+    fn fetch(&self, url: Url) -> Result<Box<dyn Read + Send>, TransportError> {
+        if url.scheme() != "file" {
+            return Err(TransportError::new(
+                TransportErrorKind::UnsupportedUrlScheme,
+                url,
+            ));
+        }
+
+        // If the path returned from url.path() begins w/ "/" , we remove this "/" so windows doesn't complain
+        // This is dirty, but the only way i found to do it ＞﹏＜ ; any edits are welcome.
```
webern: Can you explain what is going on here? Do you have an example of one of these URLs that have a w/ as its first path component? Although I don't understand the issue being fixed here yet, I wonder if there might be a crate for this kind of thing if it is encountered frequently.

More generally, we should try to reduce the amount of code that is inside the `cfg`  switch to only the parts that are different (rather than having duplicate code in both static branches). But I'd like to understand the issue better before suggesting specific changes.

webern (commented):
Thanks again for submitting this PR! I have some initial suggestions and questions.

tough/src/transport.rs
Cytro54: So, what happens here is that the url.path() function, when called with a path that begins with the "file://" protocol returned paths that looked like : "/c:/foo/bar" instead of the expected "c:/foo/bar". 
This broke some tests with windows complaining about invalid file paths. ; I wrote a hack that removes, if found the "/" in front of the returned address. 
It looks like the crate is having issues in this case under windows and there looks like there are some issues on their repo about these kinds of problems, but I'm not sure of when it'll be fixed, so in the meantime, I wrote this to get it working.

Cytro54: As an example, when I passed to the url.path() function an url that looked like "file:///c:/foo/bar", it returned "/c:/foo/bar", which wasn't a valid Windows path.

tough/src/editor/signed.rs
Cytro54: Applied the suggestion in the #174d5cf commit

make.ps1
Cytro54: I've tried using the `make` command on the makefile under cygwin with success, but I'm worried that GitHub actions may not support this software. I'll look into it (if you want to) so that automatic testing can also work under windows.

webern:
My recommendation would be that we remove the powershell script and remove the CI workflow changes from this PR. My reasoning is that getting the CI workflow just right can be a pain and can take a lot of pushing, tweaking and trial-and-error. I suggest that we dedicate another Issue/PR to adding the Windows platform to CI after this PR lands. I would be happy to take that up myself unless you want to work on it (I have some other CI changes in mind, and would like to try to use the `matrix` feature for multiple platforms).

Regarding the issue with paths, I'm researching that a bit more right now.

Thanks!

tough/src/transport.rs
```diff
@@ -139,6 +139,7 @@ impl Error for TransportError {
 pub struct FilesystemTransport;
 
 impl Transport for FilesystemTransport {
+    #[cfg(not(target_os = "windows"))]
```
webern:
```suggestion
```

I'm working on getting a Windows machine setup, so I'm not sure if this works yet. But the URL function [to_file_path](https://docs.rs/url/2.2.0/url/struct.Url.html#method.to_file_path) might work for us.

Could you try changing line 151 to:

`let f = std::fs::File::open(url.to_file_path().unwrap()).map_err(|e| {`

and removing the other changes to see if that works?
If it works we will want to handle the error instead of `unwrap()`, but we can see if it works first.

Cytro54: Ok , trying it now

Cytro54: It looks like it works ! (omg I'm sorry I haven't thought about this one...)

Cytro54:
Ok, i'll create a new Issue concerning the CI Workflow, I originally just wrote a GitHub action because I thought it would be a nice replacement for the PowerShell script ; I'll remove it from this PR if you want to.

tough/src/transport.rs
webern: Nice!

.github/workflows/rust.yml
```diff
@@ -8,4 +8,4 @@ jobs:
     steps:
     - uses: actions/checkout@v1
     - run: rustup update stable
-    - run: make ci
+    - run: make ci
```
webern:
```suggestion
    - run: make ci

```

webern (commented):
I think this is very close now. The last thing would be to squash the commits. I think this should have two commits:
1) The fmt commit
2) everything else squashed/fixup'ed

Also there is a stray newline deletion on the rust.yml file.

Thanks!

Cytro54:
I'm having an ... rather interesting bug with the File sizes, it looks like the tests doesn't complain about file size on the GitHub actions runner but when I run it ; with or without the patch on my dev machine under WSL/Ubuntu ; the same test that passed on github CI breaks locally. 
If this doesn't bother you, i'd like to investigate this before calling the patch mature.

webern:
> If this doesn't bother you, i'd like to investigate this before calling the patch mature.

Yes that sounds important. Do you mean that targets fail to validate for being the wrong size?

Cytro54:
targets.json and snapshot.json fails to load for being the wrong size ; I honestly don't know why but, it looks like the file are too big on my machine (e.g: snapshot.json - expected: 1024 bytes ; on disk : 1234 bytes)

Cytro54:
*all bytes counts provided come from the file sizes found in the 'tough/tests/data/expired-repository'

webern:
What is the name of the failing test?

In the test, I suppose the failure happens from an `unwrap()` statement in the test code. If so, toward the top of the panic printout we should be able to see the Debug text of the error that was unwrapped. Could you paste that?

tough/src/transport.rs
```diff
@@ -147,7 +147,7 @@ impl Transport for FilesystemTransport {
             ));
         }
 
-        let f = std::fs::File::open(url.path()).map_err(|e| {
+        let f = std::fs::File::open(url.to_file_path().unwrap()).map_err(|e| {
```
webern: I forgot we will also need to add an error variant to `error.rs` for this and use snafu's context function like this (instead of unwrapping)

```suggestion
        let f = std::fs::File::open(url.to_file_path().context(error::UrlToFilePath{ url })?).map_err(|e| {
```

Cytro54:
The falling test is `test_expiration_enforcement_unsafe` ; it panics at the `assert!` a the end of the test ; i've added a `dbg!` statement with the result; and i lead a little investigation and instead of the `Ok()` response that the assertion was expecting, it found this : `
ParseMetadata {
        role: Snapshot,
        source: Error("Maximum size 1204 (specified by timestamp.json) exceeded", line: 27, column: 514),
        backtrace:    0:     0x7ff7cb306268 - backtrace::backtrace::dbghelp::trace
                                       at C:\Users\Enzo\.cargo\registry\src\github.com-1ecc6299db9ec823\backtrace-0.3.55\src\backtrace\dbghelp.rs:98     
           1:     0x7ff7cb306268 - backtrace::backtrace::trace_unsynchronized<closure-0>
                                       at C:\Users\Enzo\.cargo\registry\src\github.com-1ecc6299db9ec823\backtrace-0.3.55\src\backtrace\mod.rs:66
           2:     0x7ff7cb306110 - backtrace::backtrace::trace<closure-0>
                                       at C:\Users\Enzo\.cargo\registry\src\github.com-1ecc6299db9ec823\backtrace-0.3.55\src\backtrace\mod.rs:53
           3:     0x7ff7cb300e78 - backtrace::capture::Backtrace::create
                                       at C:\Users\Enzo\.cargo\registry\src\github.com-1ecc6299db9ec823\backtrace-0.3.55\src\capture.rs:176
           4:     0x7ff7cb300dd9 - backtrace::capture::Backtrace::new
                                       at C:\Users\Enzo\.cargo\registry\src\github.com-1ecc6299db9ec823\backtrace-0.3.55\src\capture.rs:140
           5:     0x7ff7cb2feb31 - snafu::{{impl}}::generate
                                       at C:\Users\Enzo\.cargo\registry\src\github.com-1ecc6299db9ec823\snafu-0.6.10\src\lib.rs:648
           6:     0x7ff7cab2f640 - tough::error::{{impl}}::into_error<tough::schema::RoleType>        
                                       at C:\Users\Enzo\Projets\Rust\tough\tough\src\error.rs:20      
           7:     0x7ff7cabce0aa - snafu::{{impl}}::context::{{closure}}<tough::schema::Signed<tough::schema::Snapshot>,serde_json::error::Error,tough::error::ParseMetadata<tough::schema::RoleType>,tough::error::Error>
                                       at C:\Users\Enzo\.cargo\registry\src\github.com-1ecc6299db9ec823\snafu-0.6.10\src\lib.rs:318
           8:     0x7ff7cab15890 - core::result::Result<tough::schema::Signed<tough::schema::Snapshot>, serde_json::error::Error>::map_err<tough::schema::Signed<tough::schema::Snapshot>,serde_json::error::Error,tough::error::Error,closure-0>
                                       at C:\Users\Enzo\.rustup\toolchains\stable-x86_64-pc-windows-msvc\lib\rustlib\src\rust\library\core\src\result.rs:595
           9:     0x7ff7cab1a503 - snafu::{{impl}}::context<tough::schema::Signed<tough::schema::Snapshot>,serde_json::error::Error,tough::error::ParseMetadata<tough::schema::RoleType>,tough::error::Error>
                                       at C:\Users\Enzo\.cargo\registry\src\github.com-1ecc6299db9ec823\snafu-0.6.10\src\lib.rs:318
          10:     0x7ff7cab32153 - tough::load_snapshot
                                       at C:\Users\Enzo\Projets\Rust\tough\tough\src\lib.rs:775       
          11:     0x7ff7caa0657f - tough::Repository::load<std::fs::File>
                                       at C:\Users\Enzo\Projets\Rust\tough\tough\src\lib.rs:325       
          12:     0x7ff7caa07721 - tough::RepositoryLoader<std::fs::File>::load<std::fs::File>        
                                       at C:\Users\Enzo\Projets\Rust\tough\tough\src\lib.rs:187       
          13:     0x7ff7caa09e54 - expiration_enforcement::test_expiration_enforcement_unsafe
                                       at C:\Users\Enzo\Projets\Rust\tough\tough\tests\expiration_enforcement.rs:46
          14:     0x7ff7ca9f6bde - expiration_enforcement::test_expiration_enforcement_unsafe::{{closure}}
                                       at C:\Users\Enzo\Projets\Rust\tough\tough\tests\expiration_enforcement.rs:44
          15:     (snipped)
          38:     0x7ffc71b7d0d1 - RtlUserThreadStart
        ,
    },
`
First, I did not understand ; why the file is too big, but it's okay when executed on GitHub servers ?
I think it's a problem with my machine, let me explain : 
 - I looked at the file size of the problematic file in the folder in which I cloned in the repo. (it was too big)
 - I then tried to download the file directly from the `raw.githubusercontent.com` address, and this time, it was the right size.

Cytro54:
Uh, the dump is unreadable, let me extract the `source` entry : `Error("Maximum size 1204 (specified by timestamp.json) exceeded", line: 27, column: 514)`

Cytro54:
But the thing that bothers me the most is that when I tried to automate the CI process ; the same test failed when GitHub tried to run make ci on Windows ; I really don't know

Cytro54:
Okay, I think I found out the culprit ; I'll test something, and then get back in touch

Cytro54:
I finished tracking the issue; and I think that the patch is ready for a new review ; `make ci` now fully works under windows, (including integrations test via docker desktop) ; all test pass under windows, and some documentation was added to help new people to get started and build / test the library under windows.

tough/src/transport.rs
```diff
@@ -46,6 +46,8 @@ pub enum TransportErrorKind {
     /// transports it might be less obvious, but the intent of `FileNotFound` is to indicate that
     /// the file probably doesn't exist.
     FileNotFound,
+    /// The conversion from the "file://" URL path into a disk path failed.
+    UrlConversionToPathFailed,
```
webern:
```suggestion
```

I'm sorry I misled you when I told you to add an `error` type. Actually we should return `TransportErrorKind::Other` for this.

tough/src/transport.rs
```diff
@@ -56,6 +58,8 @@ impl Display for TransportErrorKind {
             f,
             "{}",
             match self {
+                TransportErrorKind::UrlConversionToPathFailed =>
+                    "URL conversion to file path failed",
```
webern:
```suggestion
```

Same as above, we should return `TransportErrorKind::Other`.

tough/src/transport.rs
```diff
@@ -140,14 +144,21 @@ pub struct FilesystemTransport;
...
             return Err(TransportError::new(
                 TransportErrorKind::UnsupportedUrlScheme,
                 url,
             ));
         }
 
-        let f = std::fs::File::open(url.path()).map_err(|e| {
+        // Convert the file URL into a file path
+        let file_path = &url.to_file_path().map_err(|_e| {
+            TransportError::new(TransportErrorKind::UrlConversionToPathFailed, &url)
```
webern:
```suggestion
            TransportError::new_with_cause(TransportErrorKind::Other, &url, e)
```

When I told you to add an error, I was wrong. I was thinking the error type was crate::error::Error. Since this is a TransportError, we want to avoid creating new variants.

tough/tests/http.rs
```diff
@@ -121,18 +134,25 @@ mod http_integ {
...
-        let output = Command::new("bash")
-            .arg(
-                integ_dir()
-                    .join("failure-server")
-                    .join("run.sh")
-                    .into_os_string(),
-            )
+
+        // Get the "run.sh" path
+        let failure_server_path_str = integ_dir()
```
webern:
```suggestion
        let script_path = integ_dir()
```

tough/tests/http.rs
```diff
@@ -121,18 +134,25 @@ mod http_integ {
...
+            .join("failure-server")
+            .join("run.sh")
+            .into_os_string()
+            .into_string()
+            .unwrap();
+
+        // Run it under BASH
+        let output = bash_base()
+            .arg(OsString::from(failure_server_path_str))
```
webern:
```suggestion
            .arg(OsString::from(script_path))
```

tuftool/src/source.rs
```diff
@@ -56,6 +56,18 @@ pub(crate) fn parse_key_source(input: &str) -> Result<Box<dyn KeySource>> {
         .base_url(Some(&pwd_url))
         .parse(input)
         .context(error::UrlParse { url: input })?;
+    dbg!(&url);
+    dbg!(format!("{}:{}", url.scheme(), url.path()));
+
+    // If we are on on windows , the scheme passed might be the drive letter, so check if the file exists before checking url scheme.
```
webern:
```
    // If we are on on windows , the scheme passed might be the drive letter, so check if the file exists before checking url scheme.
```

I'm not sure about this. On *nix a bare path will also fail (because it has no scheme), shouldn't Windows users also provide a valid `file://` URL? I think Windows users just need to provide a proper URL, e.g. `file:///c:/path/to/the%20file.txt` (which I took from [here](https://en.wikipedia.org/wiki/File_URI_scheme#Windows)).

On the other hand, it might be a pain for users to have to convert. Maybe we should support scheme-less paths for both *nix and Windows.

I think what we should do is treat Windows support for `tuftool` as separate from `tough` and open an issue for it. (We could also open a separate issue for schemeless paths in tuftool). That would give us a place to discuss the best way to handle path arguments for Windows. I also suspect that there may be other lurking issues to address in `tuftool`.

.github/workflows/rust.yml
webern: Reminder to add a newline (not CLRF!) to the end of this so that it doesn't change with this PR.

.gitconfig
```diff
@@ -0,0 +1,2 @@
+[core]
```
webern: I think we should accomplish this with a file at `tough/tests/data/.gitattributes` instead. The gitattributes file should declare that everything is binary, which I think will disable autocrlf.

Can you try it? I think the gitattributes file could look like this

`tough/tests/data/.gitattributes`

```text
# Denote all files as binary so they will not be modified by autocrlf.
* binary
```

The reason I think this is better is that it would not interfere with a Windows developer's preferred workflow.

README.md
```diff
@@ -4,6 +4,14 @@
...
 **tuftool** is a Rust command-line utility for generating and signing TUF repositories.
 
+## ❗ Warnings
+### Windows
+- Tests *will* break under windows if ; before cloning ; you don't set the git flag `core.autocrlf = input`. This is due to the fact that some tests require some test files to have a *precise* byte size and that `core.autocrlf` makes some files bigger while converting from LF to CLRF.
+- The build process requires [Cygwin](https://cygwin.com/install.html) to work properly.
+- Some integration tests require [Docker Desktop](https://www.docker.com/products/docker-desktop) to work properly.
+### UNIX/Linux
+- Integration tests require `docker` to work.
+
```
webern: I think the requirement for `docker` should be clear enough for users of any platform.

I think the idea of warning Windows developers about the potential for line ending changes to cause failing tests is a good one. I changed the working to say that we have mitigated it with a .gitattributes file (as suggested above), but I think the warning is still valid because I could see how the issue could pop-up somewhere else over time.

```suggestion
## Integration Testing

Integration tests require `docker`.

### Windows❗ Warning

Tests can break on Windows if Git's `autocrlf` feature changes line endings.
This is due to the fact that some tests require files to have a *precise* byte size and hash signature.
We have mitigated this with a `.gitattributes` file in the test data directory.
```

tough/tests/http.rs
```diff
@@ -96,7 +96,20 @@ mod http_integ {
...
-        p.join("integ").canonicalize().unwrap()
+        p = p.join("integ");
+        p
+    }
+
+    /// Returns a command object that runs the provided script under BASH , wether we are under cygwin or unix.
+    pub fn bash_base() -> Command {
+        // if under cygwin, run the bash script under cygwin64 bash
+        if cfg!(target_os = "windows") {
+            let mut command = Command::new("c:\\cygwin64\\bin\\bash");
```
webern, line 107:
```
            let mut command = Command::new("c:\\cygwin64\\bin\\bash");
```

The absolute path here causes me concern. It may be the most likely place for cygwin bash to be installed but it could actually be anywhere. Is it typical to not have cygwin's bin dir in the machine's PATH?

webern (commented):
Nice job figuring out what was causing the test failures! With my comments I'm generally trying to steer us in the direction of smaller scope with follow up issues rather than trying to solve all Windows problems in one PR.

tough/tests/http.rs
Cytro54: I looked a bit into it, but it doesn't seem like they are any way to retrieve from the cmd shell the cygwin64 home path :(, when I looked up integration with cygwin from other apps, it looked like hard-coding the cygwin root path wasn't a that bad thing.

Cytro54: ,and `cygpath` isn't available from `cmd` or `powershell`, so ... :(

README.md
Cytro54: I'll edit that, but I think that still warning users that cygwin is required for integration tests to run is important ; because we are still using `run.sh` and `teardown.sh` for integration tests and installing only make for windows wouldn't work because of the .sh files that need a bash-like environment to run.

tough/tests/http.rs
Cytro54: https://github.com/awslabs/tough/pull/342#discussion_r576441363
they *aren't

README.md
Cytro54: I moved this info into a comment in the `Cargo.toml` file, under the `integ` feature comment.

.gitconfig
Cytro54: It works, commited in https://github.com/awslabs/tough/pull/342/commits/670fdf817b249ee126c1bf61cd6ffbe3150eb823

tough/tests/http.rs
Cytro54: Done in https://github.com/awslabs/tough/pull/342/commits/ab5c715717c0c4db6df09bc66f377e965bc18226

tough/tests/http.rs
Cytro54: Done in https://github.com/awslabs/tough/pull/342/commits/ab5c715717c0c4db6df09bc66f377e965bc18226

README.md
webern: Sorry I removed the cygwin statement from this readme suggestion by mistake. It can be added back to the Windows warning here.

tough/src/transport.rs
Cytro54: No problem, fixing that in https://github.com/awslabs/tough/pull/342/commits/2a2950ea84ffb218132d4c81790acf27d27f11ab

tuftool/src/source.rs
Cytro54: We can create a new Issue for that (after we finish this one, I'm bad doing much at the same time), I remember writing that because tests failed and tests call `tuftool` with the raw Windows path and that the `url` crate interpreted that as a prefix, and didn't think a lot about it.

IMO, I think that raw path support should be added in both platforms because it's established in other CLI tools, but again, I think that discussing this in another Issue is a good idea.

README.md
Cytro54: Oh, no problem, putting it back in https://github.com/awslabs/tough/pull/342/commits/cad8d347008ffaf834caf270177b617052f9c94a

tough/src/transport.rs
```diff
@@ -140,14 +140,26 @@ pub struct FilesystemTransport;
...
-        let f = std::fs::File::open(url.path()).map_err(|e| {
+        // Convert the file URL into a file path
+        let file_path = &url.to_file_path().map_err(|_e| {
+            TransportError::new(TransportErrorKind::Other, &url)
+            // We return no source because :
+            // - The source must implement Error, but url.to_file_path() only returns a () as a error return value.
+            // - We can't write an oneoff error like anyhow!("Failed to convert the URL into a file path") because anyhow isn't listed as a dep, but if added ;
+            //   that would be replaced with :
+            // TransportError::new_with_cause((TransportErrorKind::Other, url, anyhow::anyhow!("Conversion from URL to disk path failed"))
+        })?;
```
webern: OK, thank you for explaining that in the comment. The problem is that there would be no good message. Rust implements Error for String, so I think this would work, and would provide an error message to the user.

```suggestion
        // Convert the file URL into a file path
        let file_path = &url.to_file_path().map_err(|_| {
            TransportError::new_with_source(TransportErrorKind::Other, &url, "unable to get filepath from URL".to_string())
        })?;
```

tough/tests/http.rs
webern: OK, then perhaps this is the cleanest solution. I'll leave this conversation 'unresolved' so other reviewers will see it.

tuftool/src/source.rs
webern: Yeah, it might feel 'wrong' to submit a PR where some tests are failing. But in this case, since we're working on support for a new platform, and since we have multiple Rust crates in the repo, I think it's fine if `tuftool` tests are failing for now.

I've opened #347 and #348 for tracking, and I think we should back out the `tuftool` changes from this PR.

tough/src/transport.rs
Cytro54: Ok, committing that, I didn't think about this one, idk why.

Cytro54:
Ok, I removed the `tuftool` fix, and it builds under *nix. I think that, if everything is ok to you, we can call this PR complete so that we can focus on the `tuftool` path problem and the CI one.

webern (commented):
It looks good. There's one more thing though, and that's to squash the commits.
We don't want all of these transient commits in the history, and if I click the squash and merge button then your authorship will be lost (see [this](https://github.com/isaacs/github/issues/1303))

If you're not familiar with interactive rebase, then the following series of commands should do the trick. Be careful though, maybe back up the repo before doing this depending on your git expertise level.

```sh
git checkout windows-fix
git branch -m windows-fix-old
git checkout develop
git checkout -b windows-fix
git merge --squash windows-fix-old
git commit -m'tough: fix build on windows'
git push origin -f windows-fix:windows-fix
```

Cytro54:
Done, every commit was squashed into one  ♪(^∇^*)

.gitignore
```diff
@@ -1,2 +1,3 @@
 /target
 **/*.rs.bk
+.vscode
```
srgothi92: I think we should get rid of IDE specific files.
I personally set global git ignore for them in my machine so they don't bother me with every repo. 
You can modify `~/.gitconfig` to include:
```
[core]
    excludesfile = ~/.gitignore
```

and create ~/.gitignore with
```
.idea
.vscode
```

webern:
> Done, every commit was squashed into one ♪(^∇^*)

Yes that looks good. One more thing :)

The commit message should tell us what the commit is about. Right now when looking at git history we would see "ReSquashing all fixes required" which wouldn't help if we were trying to debug something or whatever.

In this repo we usually use a the naming convention `crate: message`, so a command to fix it would look like this:

```sh
git commit --amend -m'tough: fix windows build'
git push origin -f windows-fix
```

tough/tests/data/.gitattributes
```diff
@@ -0,0 +1,2 @@
+# Denote all files as binary so they will not be modified by autocrlf.
+* binary
```
srgothi92: nit: new line

tough/tests/expiration_enforcement.rs
```diff
@@ -21,6 +21,7 @@ fn test_expiration_enforcement_safe() {
         dir_url(base.join("targets")),
     )
     .load();
+
```
srgothi92: Do we need this ? It would appear as changed file in this commit which is not necessary.

tuftool/src/source.rs
```diff
@@ -56,6 +56,18 @@ pub(crate) fn parse_key_source(input: &str) -> Result<Box<dyn KeySource>> {
         .base_url(Some(&pwd_url))
         .parse(input)
         .context(error::UrlParse { url: input })?;
+    dbg!(&url);
+    dbg!(format!("{}:{}", url.scheme(), url.path()));
```
srgothi92: do we need this ? 
```suggestion
 
```

tuftool/src/source.rs
```diff
@@ -56,6 +56,18 @@ pub(crate) fn parse_key_source(input: &str) -> Result<Box<dyn KeySource>> {
...
+
+    // If we are on on windows , the scheme passed might be the drive letter, so check if the file exists before checking url scheme.
+    if cfg!(target_os = "windows") {
+        let windows_disk_path = PathBuf::from(format!("{}:{}", url.scheme(), url.path()));
+        if windows_disk_path.exists() {
+            return Ok(Box::new(LocalKeySource {
```
srgothi92: Are we supporting only `LocalKeySource` for windows ? If yes, should be document this somewhere ?

tough/tests/http.rs
srgothi92: This make me sad as well. Did you try running this tests from cygwin ? I think bash command will be available there directly. If tests works on cygwin we can document to run this tests only from cygwin or related unix style terminal for windows because of shell script dependency. 
Alternative could be to re-write `run.sh` as powershell script if that is possible 😅

integ/failure-server/run.sh
```diff
@@ -14,6 +14,14 @@ set -eo pipefail
 DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
 TUF_REFERENCE_REPO="${DIR}/../../tough/tests/data/tuf-reference-impl"
 
+# if we are under cygwin , set a windows style path
+# (docker for windows requires windows style paths to work ,it doesn't understand cygwin-style paths)
+if [[ "$OSTYPE" == "cygwin" ]]; then
```
srgothi92, line 19: even`MinGW` is famous for windows. However, i think there is no generic way to detect windows [OS](https://stackoverflow.com/a/8597411)
@webern  should we consider to support MinGW ?

srgothi92 (commented):
Great work ! Its good to get windows support 🎉 
I got few minor questions and nit, otherwise PR looks good.

tuftool/src/source.rs
webern: We had agreed not to make any changes to tuftool in this PR, I think it might have been a mistake that this ended up back in the PR?

.gitignore
webern: I personally prefer if projects allow me to .gitignore my IDE in the project itself. The reason is that some projects may want to check-in shared IDE files (complicated run configs, for example). I have dealt with this by writing a script that can globally ignore and unignore things, but I'd rather just throw `.idea` and `.run` in the project's `.gitignore`.

I may be outnumbered by the team, but I don't see the harm...

integ/failure-server/run.sh
webern: I think if it fixes the script in the contributor's workflow then I'm happy with it. Mostly we will need to lock things down with CI (#349) and document what tools are needed for the developer workflow on Windows.

webern (changes requested):
Sorry, somehow the `tuftool` changes came back!

.gitignore
srgothi92: Yeah I also don't see any harm. But, Its not added in any of our repos, so I thought it would be worth to point out.

tough/tests/http.rs
Cytro54: I always test `make ci` under cygwin when testing windows support, but the thing is that I'm scared that on some windows installs, calling simply `bash` even from cygwin might load the WSL bash, (after all, cygwin supports windows commands, and, if you have WSL installed, `bash` is a valid command under `cmd` \ `powershell` to drop you in a wsl prompt). 
This never happened on my machine, but I prefer being safe on this one. Plus, on a dev machine with WSL, make for windows and cygwin installed, if someone tries to run `make ci` from their cmd, calling `bash` would, at best crash ; and at worst, run tests under WSL, which would lead some to confusing results. 
So I prefer hard-coding the path, but I'm not happy about this.

tuftool/src/source.rs
Cytro54: Yes, I don't know how, but I made some mistake with git, I'll restore the file to its original state

.gitignore
Cytro54: I'll remove it ; I don't think this project will be using IDE specific files soon, it already contains a complete makefile for CI.

Cytro54 (commented):
Thanks for the review! 
I answered here some of your questions ; applied requested changes and squashed all commits into one.
One last thing, I asked git to now treat `run.sh` and `teardown.sh` as binary, autocrlf did break integration tests under a WSL bash when the repo was cloned under windows because WSL\Ubuntu bash doesn't like CRLF endings.

tough/tests/http.rs
srgothi92: Okay, I agree with you. What is the error message if `cygwin` does not exist at path `c:\\cygwin64\\bin\\bash` ? If it already does not say, can we have a way to inform user that `cygwin` is expected at `c:\\cygwin64` or something similar. It's well documented in Readme, may be an overkill to check location. But it seems helpful according to me, If error message is misleading and actual reason was `cygwin` not found at expected location.

integ/failure-server/run.sh
srgothi92: I don't think it would. We don't need it as of now.

srgothi92 (approved):

webern (approved):
Looks good to me. @iliana and @zmrow, any concerns?

zmrow (approved):
Nothing looks too concerning here.  Thanks for the contribution!

:musical_note:

iliana:
No concerns.

webern:
Woohoo, nice job @Cytro54!

PR #355, opened 2021-02-23
speed up ci with caching

### LINKED ISSUE #259: ci is slow
~~I don't understand this, but it seems like each `- run` step in the GitHub actions results in a full rebuild of all the crates. A workaround would be to create separate jobs that run in parallel. A better solution would be to figure out what causes cargo to start over on subsequent `- run` steps.~~ It's slow because it's a small machine CPU-wise and there's a lot of Rust to build.

While we're at it, we can switch to checkout v2 and use the GitHub actions caching feature to cache the `targets` directory. It seems intuitively plausible that the using the caching feature could fix the problem if it causes some different filesystem to be mounted there (really no idea what the problem is).

webern:
*Issue #, if available:*

Closes #259

*Description of changes:*

- Cache build and certain CARGO_HOME directories for CI speedup.
- Split CI into two parts
  - one that does all of the crate builds, fmt, clippy and unit tests,
  - another that runs the docker/script-based HTTP integ test.

The cache behavior is such that, when Cargo.lock changes, a cache miss happens, but we restore the most recent cache and create a new cache after the build. There is a file named `cache_bust` that, if we make a change to it, we will start a fresh build with no cache restore.

One _could_ get the same `cache_bust` behavior by changing some prefix in the cache key in the `.gitignore` file, but the edit would need to be made in more than one place. I like the `cache_bust` file because it is a single edit that will discard all caches.

*Testing*

- `make ci` still works locally.
- GitHub actions run in about 4 minutes. Previously actions took about 11 or 12 minutes.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -1,11 +1,26 @@
 name: Rust
 on:
   pull_request:
+    paths-ignore:
+      - '**.md'
     branches: [develop]
 jobs:
   build:
+    strategy:
+      matrix:
+        make_target: ["build", "integ"]
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@v2
+    - uses: actions/cache@v2
+      with:
+        path: |
+          ~/.cargo/registry
+          ~/.cargo/git
+          target
+        # you can edit the .github/cache_bust file if you need to clear the cache
+        key: ${{ hashFiles('.github/cache_bust') }}-${{ runner.os }}-${{ matrix.make_target }}-${{ hashFiles('**/Cargo.lock') }}
+        restore-keys: |
+          ${{ hashFiles('.github/cache_bust') }}-${{ runner.os }}-${{ matrix.make_target }}-
     - run: rustup update stable
-    - run: make ci
+    - run: make ${{ matrix.make_target }}
--- a/.github/cache_bust
+++ b/.github/cache_bust
@@ -1,0 +1,4 @@
+# this file provides a manual way to clear out github actions caches. any change
+# to this file will cause all github action caches to miss. increment the number
+# below by 1 if you need to clear the caches.
+1
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,10 @@
+# the series of builds, tests and checks that runs for pull requests. requires docker.
 .PHONY: ci
-ci: ## The series of builds, tests and checks that runs for pull requests. Requires docker.
+ci: build integ
+
+# builds each crate, runs unit tests at the workspace level, and runs linting tools.
+.PHONY: build
+build:
 	set +e
 	cargo fmt -- --check
 	cargo clippy --locked -- -D warnings
@@ -9,5 +14,10 @@ ci: ## The series of builds, tests and checks that runs for pull requests. Requi
 	cargo build --locked -p tough-kms
 	cargo build --locked -p tuftool
 	cargo test --locked
+
+# checks tough tests with and without the http feature. http testing requires docker.
+.PHONY: integ
+integ:
+	set +e
 	cd tough && cargo test --features '' --locked
 	cd tough && cargo test --all-features --locked


webern:
Draft while I push things to make it work...

webern:
Weird, actions aren't starting.

.github/workflows/rust.yml
```diff
@@ -6,6 +6,20 @@ jobs:
...
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@v2
+
+    - name: Cargo Home Cache
+      uses: actions/cache@v2
+      env:
+        cache-name: cargo-home
+      with:
+        path:|
+          .cargo
```
tjkirch: Do we want this whole directory?  Here's the recommended Rust setup: https://github.com/actions/cache/blob/main/examples.md#rust---cargo

Maybe https://github.com/marketplace/actions/rust-cache is worth investigating?  It offers an arbitrary `key` that it seems like you could use as a buster.

webern:
I think the markeplace action overcomplicates slightly, but thank you for pointing out the recommended setup, which I am now following. I'm going to try another push that tries running 'integ' and non-integ in parallel to see if that is any faster.

webern:
Parallelizing 'integ' vs 'unit/build' got it down another minute and a half... to 3:51.

webern:
> webern force-pushed the webern:cicache branch from e68e51c to 13d629e

Rebase.

This rebase took: 4m 9sec
Before these changes builds took from 10-13.5 minutes.

webern:
> webern force-pushed the webern:cicache branch from e68e51c to 13d629e
> webern force-pushed the webern:cicache branch from 13d629e to aef8b55
> TBD

Rebase, Improve naming and comments, rebase again because was showing a commit that isn't mine.

webern:
> webern force-pushed the webern:cicache branch from e68e51c to 13d629e
> webern force-pushed the webern:cicache branch from 13d629e to aef8b55
> webern force-pushed the webern:cicache branch from aef8b55 to e8ab58e

Rebase struggles and some renaming and comment tweakery.

webern:
> webern force-pushed the webern:cicache branch from e68e51c to 13d629e
> webern force-pushed the webern:cicache branch from 13d629e to aef8b55
> webern force-pushed the webern:cicache branch from aef8b55 to e8ab58e
> webern force-pushed the webern:cicache branch from e8ab58e to 25c28ec

Rebase struggles and some renaming and comment tweakery.

.github/workflows/rust.yml
```diff
@@ -1,11 +1,40 @@
...
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            target
+          # you can edit the .github/cache_bust file if you need to clear the cache
+          key: ${{ hashFiles('.github/cache_bust') }}-${{ runner.os }}-integ-${{ hashFiles('**/Cargo.lock') }}
+          restore-keys: |
+            ${{ hashFiles('.github/cache_bust') }}-${{ runner.os }}-integ-
+      - run: rustup update stable
```
tjkirch: Can we make rustup run once before both of the main jobs?  Seems weird / dangerous to try it in parallel.

.github/workflows/rust.yml
```diff
@@ -1,11 +1,40 @@
...
+      - run: make integ
```
tjkirch: Could we use an Actions `matrix` for the make targets rather than repeating the whole job setup?

tjkirch (commented):
It's a little weird, but balancing the time over two jobs makes sense for now.

.github/workflows/rust.yml
webern: Maybe, I'll look into it.

.github/workflows/rust.yml
webern: The jobs are isolated in in separate VMs or containers.

webern:
> webern force-pushed the webern:cicache branch from bc09f38 to cf44e1a now

This is a rebase, by branch looked weird. After this I will consolidate commits and cleanup.

webern:
OK, this is ready. I think we don't need to worry about the `Waiting for status to be reported`. That should go away in future PRs (and if it doesn't then it's a UI/settings thing in GitHub).

tjkirch:
> OK, this is ready. I think we don't need to worry about the `Waiting for status to be reported`. That should go away in future PRs (and if it doesn't then it's a UI/settings thing in GitHub).

In the repo settings, in the branch protection rules for develop, we have only "build" set as required to pass before merging, because that was the only check name before this PR.  After this PR, we should switch it to the two new status check names. :white_check_mark:

tjkirch (approved):

Issue #380, opened 2021-06-08
Add license checker to Makefile


jpculp:
Tough crates should have the licenses of their dependencies checked at build time to avoid mishaps.

This can be accomplished with the `cargo-deny` crate.

jpculp:
Merged!

PR #385, opened 2021-06-29
Add license check

jpculp:
*Issue number*

#380

*Description of changes:*

Adds a `check-licenses` target to `ci` which leverages `cargo-deny`
to verify that all crates and their dependencies have compatible licenses.

Creates a workspace **CARGO_HOME**, which will be cached in the GitHub Actions runner.

Replaces `mockito` with `httptest` to stay within existing license set.

*Testing done:*

Checked licenses for all crates. Ran and passed all unit tests.

---


jpculp:
The git actions are a little slow. Let me see if I can make any optimizations.

tuftool/tests/download_command.rs
```diff
@@ -95,19 +107,18 @@ fn download_command(metadata_base_url: Url, targets_base_url: Url) {
...
-    let targets_base_url = base_url.join("targets").unwrap();
+    let server = Server::run();
+    server.expect(create_successful_get("metadata/role1.json"));
+    server.expect(create_successful_get("metadata/role2.json"));
+    server.expect(create_successful_get("metadata/snapshot.json"));
+    server.expect(create_successful_get("metadata/targets.json"));
+    server.expect(create_successful_get("metadata/timestamp.json"));
+    server.expect(create_successful_get("targets/file1.txt"));
+    server.expect(create_successful_get("targets/file2.txt"));
+    // TODO - server.expect(create_successful_get("targets/file3.txt")); when delegate support lands.
```
srgothi92: where is this TODO coming form ? Do we need it ?

srgothi92 (approved):
Looks good to me.

tuftool/tests/download_command.rs
webern: It was there before these changes. It can be deleted. I think it might be true that this test has to be changed if we support delegates, but I'm not sure. Or maybe delegates are already supported? I forget. Anyway it's probably not a helpful TODO statement.

jpculp:
- Moved `cargo-deny` install to the Makefile to speed-up git actions.
- Removed left over TODO's from `download_command.rs`.

Makefile
```diff
@@ -1,6 +1,19 @@
+# Use directory-local cargo root to install version-specific executables into.
+export CARGO_HOME = $(shell pwd)/.cargo
```
jpculp, line 2: I suspect by exporting this I'm breaking some of the GitHub Actions caching. Should I update the cache paths, or limit this `CARGO_HOME` to the `cargo-deny` and `check-licenses` targets?

jahkeup: It looks that way. Depending on how folks feel about it, you could update the invocation to supply a specific path for the CI build:

```yaml
# in the github actions yaml:
# ...
    - run: make ${{ matrix.make_target }} CARGO_HOME=$HOME/.cargo
```

Makefile
```diff
@@ -1,6 +1,19 @@
...
+
 # the series of builds, tests and checks that runs for pull requests. requires docker.
 .PHONY: ci
-ci: build integ
+ci: check-licenses build integ
+
+# installs cargo-deny
+.PHONY: cargo-deny
+cargo-deny:
+	-cargo install --version 0.6.2 cargo-deny --no-default-features
```
jpculp: On second thought, I'll remove these `-` prefixes on my next push.

Makefile
tjkirch: We want the distinct cargo home for users running this in local checkouts, not just GitHub Actions.  We shouldn't infect their registry or installed binaries just because we have specific needs.

What are we seeing that makes us think this broke GitHub Actions caching?  The cache action would have to be pretty fragile for that, and I don't think it even uses Rust/cargo?

tjkirch: Discussed offline, I learned that the issue is just that the Actions yaml file is listing the old cargo home, so if we update it to this new cargo home we should be OK.

jpculp:
- Removed unwanted `-` prefixes.
- Changed caching logic to cache the new workspace CARGO_HOME instead.
- Tested GitHub Actions performance (`check-licenses` went from ~6m to ~11s).

Makefile
```diff
@@ -1,6 +1,19 @@
...
+
+# installs cargo-deny
+.PHONY: cargo-deny
+cargo-deny:
+	cargo install --version 0.6.2 cargo-deny --no-default-features
+
+# checks each crate, and evaluates licenses. requires cargo-deny.
+.PHONY: check-licenses
+check-licenses: cargo-deny
+	cargo deny check --disable-fetch licenses
```
webern, line 16: This seemed weird so I looked at it and it says:

```
    -d, --disable-fetch
            Disable fetching of the advisory database

            When running the `advisories` check, the configured advisory
            database will be fetched and opened. If this flag is passed, the
            database won't be fetched, but an error will occur if it doesn't
            already exist locally.
```

This is not applicable since we are running `licenses` and not `advisories`. So this flag can be removed:

```suggestion
	cargo deny check licenses
```

webern (approved):

Makefile
jpculp: From what I can tell nothing changes if this is removed, so it is _most likely_ safe to do so. That said, it's on the TODO list to take a look at the newer versions of `cargo-deny` and it might make more sense to re-evaluate things like `--disable-fetch` and `--no-default-features` then.

tjkirch (approved):

PR #404, opened 2021-08-13
tuftool: Add `clone` subcommand

### LINKED ISSUE #401: `tuftool`: provide an option to download metadata files
`tuftool download` allows a user to download either an entire repository's targets or specific target files.

It would be nice if `tuftool` provided a way to download the repository metadata files: `{snapshot, targets, timestamp}.json`.

zmrow:
*Issue #, if available:*
Fixes #401 

*Description of changes:*
This PR does a small amount of refactoring before the main event, a new `clone` subcommand!  This subcommand gives the user the opportunity to clone a repository and it's metadata, including all, none or some of the targets.  It can be considered a bit more "low-level" than `tuftool download` as the files end up on disk identically named as they are in the TUF repository, including the SHA256 prefix.  `tuftool download` also does not give the user the opportunity to download metadata.

Reading through by commit is probably easiest as they're fairly small.  

```
    tuftool: Add `download_root` module
    
    This change adds a `download_root` module, which abstracts the logic of
    downloading `root.json` out of `download` (the subcommand).  It also
    makes the HTTP request a bit more robust by handling the initial request
    error as well as a bad response.  Additionally, it avoids creating the
    file until after the request has succeeded to avoid creating cruft on
    the user's system.
```

```
    tuftool: use `download_root` module in `download`
    
    This change updates the `download` module/subcommand to make use of the
    previously added `download_root` function.
    
    It also defines a default of "1" for the `root_version` argument.
    Previously, we effectively had this default in code by using
    `1.root.json` in the event the argument wasn't passed.  It also has the
    nice side effect of not needing to deal with an `Option` for this
    argument
```

```
    tuftool: Add `clone` subcommand
    
    This adds a `clone` subcommand to `tuftool`, allowing a user to download
    a fully functioning TUF repository.  A user has the option to download a
    full repository, a subset of the targets, or just metadata
```

*Testing:*
* All new/old unit tests pass
* Manual testing:
  * Download just a repository's metadata  (with `root.json` and with the flag `--allow-root-download`)
```
$ cargo run --bin tuftool clone --metadata-only --metadata-url "https://updates.bottlerocket.aws/2020-07-07/aws-k8s-1.20/x86_64/" --allow-root-download --metadata-dir ./foo/metadata

# also ran with a downloaded root.json
$ curl -O "https://cache.bottlerocket.aws/root.json
$ cargo run --bin tuftool clone --metadata-only --metadata-url "https://updates.bottlerocket.aws/2020-07-07/aws-k8s-1.20/x86_64/" --root ./root.json --metadata-dir ./foo/metadata

 $ tree ./foo/
./foo/
└── metadata
    ├── 1628878016.snapshot.json
    ├── 1628878016.targets.json
    ├── 1.root.json
    ├── 2.root.json
    └── timestamp.json


```
  * Download an entire repository to `./foo` (with `root.json` and with the flag `--allow-root-download`)
```
$ cargo run --bin tuftool clone --metadata-url "https://updates.bottlerocket.aws/2020-07-07/aws-k8s-1.20/x86_64/" --targets-url "https://updates.bottlerocket.aws/targets/" --allow-root-download --metadata-dir ./foo/metadata --targets-dir ./foo/targets

# also ran with a downloaded root.json
$ curl -O "https://cache.bottlerocket.aws/root.json
$ cargo run --bin tuftool clone --metadata-url "https://updates.bottlerocket.aws/2020-07-07/aws-k8s-1.20/x86_64/" --targets-url "https://updates.bottlerocket.aws/targets/" --root ./root.json --metadata-dir ./foo/metadata --targets-dir ./foo/targets

$ tree ./foo
 ./foo/                                                                                                                                            
├── metadata                                                                                                                                      
│   ├── 1628878016.snapshot.json                                                                                                                  
│   ├── 1628878016.targets.json                                                                                                                   
│   ├── 1.root.json                                                                                                                               
│   ├── 2.root.json                                                                                                                               
│   └── timestamp.json                                                                                                                            
└── targets                                                                                                                                       
    ├── 0c8485912fd928a185f32a74705f8e5efa692177375c27eac7ee5c11219d259e.migrate_v1.1.0_kubelet-registry-qps-registry-burst.lz4
    ├── 148f1c10a790313c19e4967c0136c58cf433c20a41676bffb5f3c976a1170d4a.bottlerocket-aws-k8s-1.20-x86_64-1.1.1-28c4e013-root.verity.lz4
...
```

  * Download a subset of targets (just `manifest.json`)  (with `root.json` and with the flag `--allow-root-download`)
```
$ cargo run --bin tuftool clone --metadata-url "https://updates.bottlerocket.aws/2020-07-07/aws-k8s-1.20/x86_64/" --targets-url "https://updates.bottlerocket.aws/targets/" --allow-root-download --metadata-dir ./foo/metadata --targets-dir ./foo/targets --target-names manifest.json

# also ran with a downloaded root.json
$ curl -O "https://cache.bottlerocket.aws/root.json
$ cargo run --bin tuftool clone --metadata-url "https://updates.bottlerocket.aws/2020-07-07/aws-k8s-1.20/x86_64/" --targets-url "https://updates.bottlerocket.aws/targets/" --root ./root.json --metadata-dir ./foo/metadata --targets-dir ./foo/targets --target-names manifest.json

$ tree ./foo
./foo/
├── metadata
│   ├── 1628878016.snapshot.json
│   ├── 1628878016.targets.json
│   ├── 1.root.json
│   ├── 2.root.json
│   └── timestamp.json
└── targets
    └── aabc9db0c44ff6811aaf0f847c3bba6830e16bb0ecb6d2ec5eef837f68440cd2.manifest.json
```
* Ensure that targets arguments collide with the `--metadata-only` flag
```
$ cargo run --bin tuftool clone --metadata-only --metadata-url "https://updates.bottlerocket.aws/2020-07-07/aws-k8s-1.20/x86_64/" --allow-root-download --metadata-dir ./foo/metadata --target-names foo
...
error: The argument '--metadata-only' cannot be used with '--target-names <target-names>...'

USAGE:
    tuftool clone --allow-root-download --metadata-url <metadata-base-url> --metadata-dir <metadata-dir> --metadata-only --root-version <root-version> --target-names <target-names>... --targets-url <targets-base-url> --targets-dir <targets-dir>
----------------------------------
$ cargo run --bin tuftool clone --metadata-only --metadata-url "https://updates.bottlerocket.aws/2020-07-07/aws-k8s-1.20/x86_64/" --allow-root-download --metadata-dir ./foo/metadata --targets-dir foo
...
error: The argument '--targets-dir <targets-dir>' cannot be used with '--metadata-only'

USAGE:
    tuftool clone --allow-root-download --metadata-url <metadata-base-url> --metadata-dir <metadata-dir> --metadata-only --root-version <root-version> --targets-url <targets-base-url> --targets-dir <targets-dir>
```

### DIFF
--- a/tuftool/src/clone.rs
+++ b/tuftool/src/clone.rs
@@ -1,0 +1,147 @@
+// Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+use crate::common::UNUSED_URL;
+use crate::download_root::download_root;
+use crate::error::{self, Result};
+use snafu::ResultExt;
+use std::fs::File;
+use std::num::NonZeroU64;
+use std::path::PathBuf;
+use structopt::StructOpt;
+use tough::{ExpirationEnforcement, RepositoryLoader};
+use url::Url;
+
+#[derive(Debug, StructOpt)]
+pub(crate) struct CloneArgs {
+    /// Path to root.json file for the repository
+    #[structopt(
+        short = "r",
+        long = "root",
+        required_if("allow-root-download", "false")
+    )]
+    root: Option<PathBuf>,
+
+    /// Remote root.json version number
+    #[structopt(short = "v", long = "root-version", default_value = "1")]
+    root_version: NonZeroU64,
+
+    /// TUF repository metadata base URL
+    #[structopt(short = "m", long = "metadata-url")]
+    metadata_base_url: Url,
+
+    /// TUF repository targets base URL
+    #[structopt(short = "t", long = "targets-url", required_unless = "metadata-only")]
+    targets_base_url: Option<Url>,
+
+    /// Allow downloading the root.json file (unsafe)
+    #[structopt(long)]
+    allow_root_download: bool,
+
+    /// Allow repo download for expired metadata (unsafe)
+    #[structopt(long)]
+    allow_expired_repo: bool,
+
+    /// Download only these targets, if specified
+    #[structopt(short = "n", long = "target-names", conflicts_with = "metadata-only")]
+    target_names: Vec<String>,
+
+    /// Output directory of targets
+    #[structopt(long, required_unless = "metadata-only")]
+    targets_dir: Option<PathBuf>,
+
+    /// Output directory of metadata
+    #[structopt(long)]
+    metadata_dir: PathBuf,
+
+    /// Only download the repository metadata, not the targets
+    #[structopt(long, conflicts_with_all(&["target-names", "targets-dir", "targets-base-url"]))]
+    metadata_only: bool,
+}
+
+#[rustfmt::skip]
+fn expired_repo_warning() {
+    eprintln!("\
+=================================================================
+WARNING: repo metadata is expired, meaning the owner hasn't verified its contents lately and it could be unsafe!
+=================================================================");
+}
+
+impl CloneArgs {
+    pub(crate) fn run(&self) -> Result<()> {
+        // Use local root.json or download from repository
+        let root_path = if let Some(path) = &self.root {
+            PathBuf::from(path)
+        } else if self.allow_root_download {
+            let outdir = std::env::current_dir().context(error::CurrentDir)?;
+            download_root(&self.metadata_base_url, self.root_version, outdir)?
+        } else {
+            eprintln!("No root.json available");
+            std::process::exit(1);
+        };
+
+        // Structopt won't allow `targets_base_url` to be None when it is required.  We require the
+        // user to supply `targets_base_url` in the case they actually plan to download targets.
+        // When downloading metadata, we don't ever need to access the targets URL, so we use a
+        // fake URL to satisfy the library.
+        let targets_base_url = self
+            .targets_base_url
+            .as_ref()
+            .unwrap_or(&Url::parse(UNUSED_URL).context(error::UrlParse {
+                url: UNUSED_URL.to_owned(),
+            })?)
+            .clone();
+
+        // Load repository
+        let expiration_enforcement = if self.allow_expired_repo {
+            expired_repo_warning();
+            ExpirationEnforcement::Unsafe
+        } else {
+            ExpirationEnforcement::Safe
+        };
+        let repository = RepositoryLoader::new(
+            File::open(&root_path).context(error::OpenRoot { path: &root_path })?,
+            self.metadata_base_url.clone(),
+            targets_base_url,
+        )
+        .expiration_enforcement(expiration_enforcement)
+        .load()
+        .context(error::RepoLoad)?;
+
+        // Clone the repository, downloading none, all, or a subset of targets
+        if self.metadata_only {
+            println!("Cloning repository metadata to {:?}", self.metadata_dir);
+            repository
+                .cache_metadata(&self.metadata_dir, true)
+                .context(error::CloneRepository)?;
+        } else {
+            // Similar to `targets_base_url, structopt's guard rails won't let us have a
+            // `targets_dir` that is None when the argument is required.  We only require the user
+            // to supply a targets directory if they actually plan on downloading targets.
+            let targets_dir = self.targets_dir.as_ref().expect(
+                "Developer error: `targets_dir` is required unless downloading metadata only",
+            );
+
+            println!(
+                "Cloning repository:\n\tmetadata location: {:?}\n\ttargets location: {:?}",
+                self.metadata_dir, targets_dir
+            );
+            if self.target_names.is_empty() {
+                repository
+                    .cache(&self.metadata_dir, &targets_dir, None::<&[&str]>, true)
+                    .context(error::CloneRepository)?;
+            } else {
+                repository
+                    .cache(
+                        &self.metadata_dir,
+                        &targets_dir,
+                        Some(self.target_names.as_slice()),
+                        true,
+                    )
+                    .context(error::CloneRepository)?;
+            }
+        };
+
+        Ok(())
+    }
+}
--- a/tuftool/src/download_root.rs
+++ b/tuftool/src/download_root.rs
@@ -1,0 +1,51 @@
+// Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: MIT OR Apache-2.0
+//! The `download_root` module owns the logic for downloading a given version of `root.json`.
+
+use crate::error::{self, Result};
+use snafu::ResultExt;
+use std::fs::File;
+use std::num::NonZeroU64;
+use std::path::{Path, PathBuf};
+use url::Url;
+
+/// Download the given version of `root.json`
+/// This is an unsafe operation, and doesn't establish trust. It should only be used for testing!
+pub(crate) fn download_root<P>(
+    metadata_base_url: &Url,
+    version: NonZeroU64,
+    outdir: P,
+) -> Result<PathBuf>
+where
+    P: AsRef<Path>,
+{
+    let name = format!("{}.root.json", version);
+
+    let path = outdir.as_ref().join(&name);
+    let url = metadata_base_url.join(&name).context(error::UrlParse {
+        url: format!("{}/{}", metadata_base_url.as_str(), name),
+    })?;
+    root_warning(&path);
+
+    let mut root_request = reqwest::blocking::get(url.as_str())
+        .context(error::ReqwestGet)?
+        .error_for_status()
+        .context(error::BadResponse { url })?;
+
+    let mut f = File::create(&path).context(error::OpenFile { path: &path })?;
+    root_request.copy_to(&mut f).context(error::ReqwestCopy)?;
+
+    Ok(path)
+}
+
+/// Print a very noticeable warning message about the unsafe nature of downloading `root.json`
+/// without verification
+fn root_warning<P: AsRef<Path>>(path: P) {
+    #[rustfmt::skip]
+    eprintln!("\
+=================================================================
+WARNING: Downloading root.json to {}
+This is unsafe and will not establish trust, use only for testing
+=================================================================",
+              path.as_ref().display());
+}
--- a/tuftool/src/error.rs
+++ b/tuftool/src/error.rs
@@ -243,6 +243,13 @@ pub(crate) enum Error {
         backtrace: Backtrace,
     },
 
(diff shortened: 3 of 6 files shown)


zmrow:
^ Fix the clippy complaint - I didn't see it when I ran the `make build` command locally...

tuftool/src/download_root.rs
```diff
@@ -0,0 +1,51 @@
...
+    version: NonZeroU64,
+    outdir: P,
+) -> Result<PathBuf>
+where
+    P: AsRef<Path>,
+{
+    let name = format!("{}.root.json", version);
+
+    let path = outdir.as_ref().join(&name);
+    let url = metadata_base_url.join(&name).context(error::UrlParse {
```
tjkirch, line 25: I know this is an existing line of code, so not critical, but I think using UrlParse as an error variant here is misleading, because we're not showing the base URL and the part we're trying to join onto it (which actually fails), we're just showing the base URL and basically telling the user it's invalid, which isn't necessarily true.  I think we could either make a new variant for failed join, or just String-join the base and name and give that as the failed url for UrlParse..?

tuftool/src/main.rs
```diff
@@ -77,14 +78,16 @@ impl Program {
...
-    /// Download a TUF repository's resources
+    /// Download a TUF repository's targets
     Download(download::DownloadArgs),
     /// Update a TUF repository's metadata and optionally add targets
     Update(Box<update::UpdateArgs>),
     /// Manipulate a root.json metadata file
     Root(root::Command),
     /// Delegation Commands
     Delegation(Delegation),
+    /// Clone a fully functional TUF repository
```
tjkirch:
```suggestion
    /// Clone a TUF repository, including metadata and some or all targets
```
It may not be fully functional depending on their given arguments :)

tuftool/tests/clone_command.rs
```diff
@@ -0,0 +1,190 @@
...
+    let mut cmd = Command::cargo_bin("tuftool").unwrap();
+    clone_base_command(&mut cmd, &repo_paths)
+        .args(&["--metadata-only", "--target-names", "foo"])
+        .assert()
+        .failure();
+}
+
+#[test]
+// Ensure that, even when provided with arguments for targets, the `--metadata-only` argument
+// _only_ downloads metadata files
```
tjkirch: Should we make `--metadata-only` conflict with targets-related arguments instead?

tuftool/tests/clone_command.rs
```diff
@@ -0,0 +1,190 @@
...
+    ] {
+        assert_metadata_match(&metadata_dir, f)
+    }
+}
+
+/// Given a `Command`, attach all the base args necessary for the `clone` subcommand
+fn clone_base_command<'a>(cmd: &'a mut Command, repo_paths: &RepoPaths) -> &'a mut Command {
+    cmd.args(&[
+        "clone",
+        "--root",
```
tjkirch, line 83: Nice tests!  Should we have one for `--allow-root-download` as well?

tuftool/src/clone.rs
```diff
@@ -0,0 +1,144 @@
...
+use std::num::NonZeroU64;
+use std::path::{Path, PathBuf};
+use structopt::StructOpt;
+use tough::{ExpirationEnforcement, RepositoryLoader};
+use url::Url;
+
+#[derive(Debug, StructOpt)]
+pub(crate) struct CloneArgs {
+    /// Path to root.json file for the repository
+    #[structopt(short = "r", long = "root")]
```
tjkirch: Could we use `required_if("allow-root-download", false)`?

tuftool/src/clone.rs
```diff
@@ -0,0 +1,144 @@
...
+
+    /// TUF repository targets base URL
+    #[structopt(short = "t", long = "targets-url", required_unless = "metadata-only")]
+    targets_base_url: Option<Url>,
+
+    /// Allow downloading the root.json file (unsafe)
+    #[structopt(long)]
+    allow_root_download: bool,
+
+    /// Allow repo download for expired metadata
```
webern:
```suggestion
    /// Allow repo download for expired metadata (unsafe)
```

tuftool/src/clone.rs
```diff
@@ -0,0 +1,144 @@
...
+    #[structopt(long, required_unless = "metadata-only")]
+    targets_dir: Option<PathBuf>,
+
+    /// Output directory of metadata
+    #[structopt(long)]
+    metadata_dir: PathBuf,
+
+    /// Only download the repository metadata, not the targets
+    #[structopt(long)]
+    metadata_only: bool,
```
tjkirch, line 59: Could we `conflicts_with_all(&["target-names", "targets-dir", "targets-base-url"])`?

tuftool/src/clone.rs
```diff
@@ -0,0 +1,144 @@
...
+}
+
+fn expired_repo_warning<P: AsRef<Path>>(path: P) {
+    #[rustfmt::skip]
+    eprintln!("\
+=================================================================
+Downloading repo to {}
```
tjkirch: Shouldn't we always log a message about where we're downloading the repo, not just in the case where it's expired?

tuftool/src/clone.rs
```diff
@@ -0,0 +1,144 @@
...
+WARNING: `--allow-expired-repo` was passed; this is unsafe and will not establish trust, use only for testing!
```
tjkirch:
```suggestion
WARNING: repo metadata is expired, meaning the owner hasn't verified its contents lately and it could be unsafe!
```
The existing message matches the root download warning, but they don't quite mean the same...

tuftool/src/error.rs
```diff
@@ -10,9 +10,24 @@ use std::path::PathBuf;
 
 pub(crate) type Result<T> = std::result::Result<T, Error>;
 
+// Extracts the status code from a reqwest::Error and converts it to a string to be displayed
+fn get_status_code(source: &reqwest::Error) -> String {
```
webern: Nit: I think I would put this private helper function below the pub and pub(crate) types.

webern (approved):
Nice testing.

tuftool/src/clone.rs
```diff
@@ -0,0 +1,144 @@
...
+            if self.target_names.is_empty() {
+                println!(
+                    "Cloning repository:\n\tmetadata location: {:?}\n\ttargets location: {:?}",
+                    self.metadata_dir, targets_dir
+                );
+                // The None turbofish is required to satisfy the compiler, sorry.
+                repository.cache(&self.metadata_dir, &targets_dir, None::<&[&str]>, true)
+            } else {
+                println!(
+                    "Cloning repository with a subset of targets:\n\tmetadata location:{:?}\n\ttargets location: {:?}",
```
tjkirch: nit: it doesn't feel like this slightly different message is super valuable; the user knows they requested a subset.  Should we just always pass the list of targets?  Could remove the Option to solve #405..?

tuftool/src/clone.rs
```diff
@@ -0,0 +1,144 @@
...
+                repository.cache(
+                    &self.metadata_dir,
+                    &targets_dir,
+                    Some(self.target_names.as_slice()),
+                    true,
+                )
+            }
+        };
+
+        clone_result.context(error::CloneRepository)
```
tjkirch: nit: I think this context, long after the storage of the Result, is throwing off the flow for readers more than it's helping; I'd vote for putting the context on each call, with a `?` so points of failure are more obvious, and letting the function end "normally."

tuftool/tests/clone_command.rs
zmrow: I don't see the harm in ignoring targets related arguments should the user decide to pass them.  Failing in the event a user passes targets related arguments and `--metadata-only` could avoid some confusion?

tuftool/tests/clone_command.rs
zmrow: Good point - I'll add one.

tuftool/src/clone.rs
zmrow: We do!

tuftool/src/clone.rs
zmrow: Makes sense - I went back and forth on it before going this way. :)

tuftool/tests/clone_command.rs
tjkirch: Yeah, the harm is the potential confusion, and it seems like structopt gives us the convenient 'conflicts' arguments to avoid it...

tuftool/src/clone.rs
tjkirch: Oh, did I miss that somewhere else?  Is this message duplicative, then?  This is only called when the user allows expired metadata.

zmrow: We log it further down right before we call `repository.cache()`.  I'll remove it from this message as it's duplicative.

tuftool/src/clone.rs
zmrow: Oh gosh, I was being silly.  I'll always pass the list of targets and remove the nested `if` altogether.  #405 can happen separately.

tuftool/tests/clone_command.rs
zmrow: :( Bad news everyone!  We use `reqwest` to get the `root.json`, and `reqwest` doesn't support `file:///` URLs.

tuftool/src/clone.rs
zmrow: Oops - I can't actually do that.  `None` and an empty list are treated differently so I'll need to keep the extra `if` statement for now.  I'd rather not  fix #405 as a part of this PR, especially because that "fix" changes the API of `cache()` which effectively breaks anyone who is using the library.

All that is to say I'll change my code to a single print message, but keep the `if` statement.

zmrow:
^ addressed all the comments

zmrow:
^ removed a dangling `println!()`

tuftool/tests/clone_command.rs
tjkirch: I think what we normally do is add a branch, where if url.scheme() is file, we just std::fs::copy it instead.

tjkirch (approved):

webern (approved):

tuftool/tests/clone_command.rs
zmrow: Opened a new issue to track this: https://github.com/awslabs/tough/issues/406

PR #188, opened 2020-07-17
tuftool root sign performs check of root.json

ecpullen:
*Issue #, if available:*
Closes #119

*Description of changes:*
Adds a sanity check in `tuftool root sign` to make sure it's possible for all roles to be adequately signed
Adds a signature check in `tuftool root sign` to make sure `root.json` has enough signatures

*Testing*
root_command contains tests, but must be run locally with directory `root_test` added to `tuftool/tests`

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #412, opened 2021-09-09
Add ignore threshold flag

arnaldo2792:
*Issue #, if available:*
N / A

*Description of changes:*

```
240d1613 tuftool: add `ignore-threshold` flag
```

With [`18b7f740`](https://github.com/awslabs/tough/pull/188), a check was added to verify that the root.json file was signed with enough keys as specified in the thresholds. This is a problem when a cloud provider is used to store keys, and they are in different regions, since tuftool doesn't have the notion of regions while it fetches the keys.

With the `ignore-threshold` flag, `tuftool` will be more "permissive" and won't error out if a user attempts to sign the root.json file with fewer keys than specified in the thresholds.

### DIFF
--- a/tuftool/src/root.rs
+++ b/tuftool/src/root.rs
@@ -6,6 +6,7 @@ use crate::error::{self, Result};
 use crate::source::parse_key_source;
 use crate::{load_file, write_file};
 use chrono::{DateTime, Timelike, Utc};
+use log::warn;
 use maplit::hashmap;
 use ring::rand::SystemRandom;
 use snafu::{ensure, OptionExt, ResultExt};
@@ -106,6 +107,9 @@ pub(crate) enum Command {
         ///Optional - Path of older root.json that contains the key-id
         #[structopt(short = "c", long = "cross-sign")]
         cross_sign: Option<PathBuf>,
+        ///Ignore the threshold when signing with fewer keys
+        #[structopt(short = "i", long = "ignore-threshold")]
+        ignore_threshold: bool,
     },
 }
 
@@ -153,7 +157,8 @@ impl Command {
                 path,
                 key_sources,
                 cross_sign,
-            } => Command::sign(&path, &key_sources, cross_sign),
+                ignore_threshold,
+            } => Command::sign(&path, &key_sources, cross_sign, ignore_threshold),
         }
     }
 
@@ -302,6 +307,7 @@ impl Command {
         path: &Path,
         key_source: &[Box<dyn KeySource>],
         cross_sign: Option<PathBuf>,
+        ignore_threshold: bool,
     ) -> Result<()> {
         let root: Signed<Root> = load_file(path)?;
         // get the root based on cross-sign
@@ -309,7 +315,7 @@ impl Command {
             None => root.clone(),
             Some(cross_sign_root) => load_file(&cross_sign_root)?,
         };
-        //sign the root
+        // sign the root
         let mut signed_root = SignedRole::new(
             root.signed.clone(),
             &KeyHolder::Root(loaded_root.signed),
@@ -325,14 +331,27 @@ impl Command {
                 .context(error::SignRoot { path })?;
         }
 
-        // Quick check that root is signed by enough key IDs
+        // Quick check that root is signed by enough key IDs, in all its roles.
         for (roletype, rolekeys) in &signed_root.signed().signed.roles {
-            if rolekeys.threshold.get() > rolekeys.keyids.len() as u64 {
-                return Err(error::Error::UnstableRoot {
-                    role: *roletype,
-                    threshold: rolekeys.threshold.get(),
-                    actual: rolekeys.keyids.len(),
-                });
+            let threshold = rolekeys.threshold.get();
+            let keyids = rolekeys.keyids.len();
+            if threshold > keyids as u64 {
+                // Return an error when the referenced root.json isn't compliant with the
+                // threshold. The referenced file could be a root.json used for cross signing,
+                // which wasn't signed with enough keys.
+                if !ignore_threshold {
+                    return Err(error::Error::UnstableRoot {
+                        role: *roletype,
+                        threshold,
+                        actual: keyids,
+                    });
+                }
+                // Print out a warning to let the user know that the referenced root.json
+                // file isn't compliant with the threshold specified for the role type.
+                warn!(
+                    "Loaded unstable root, role '{}' contains '{}' keys, expected '{}'",
+                    *roletype, threshold, keyids
+                );
             }
         }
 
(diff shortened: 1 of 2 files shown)


tuftool/src/root.rs
```diff
@@ -106,6 +107,9 @@ pub(crate) enum Command {
         ///Optional - Path of older root.json that contains the key-id
         #[structopt(short = "c", long = "cross-sign")]
         cross_sign: Option<PathBuf>,
+        ///Force signing with fewer keys than the specified threshold
+        #[structopt(short = "f")]
+        force: bool,
```
tjkirch: I think we should call this something more specific, because "force" could mean overriding any number of checks, and root roles are critical, so it's safer to override specific checks you know are OK to skip.  Something like `--ignore-threshold` or `--allow-keys-under-threshold` perhaps?

tuftool/tests/root_command.rs
```diff
@@ -84,7 +84,7 @@ fn add_key_all_roles(key: &str, root_json: &str) {
 fn sign_root_json(key: &str, root_json: &str) {
     Command::cargo_bin("tuftool")
         .unwrap()
-        .args(&["root", "sign", root_json, "-k", key])
+        .args(&["root", "sign", root_json, "-f", "-k", key])
```
tjkirch: Can we update testing so we have cases with fewer than the required number of keys, one without the new flag that we ensure fails, and one with the new flag that we ensure passes?

tuftool/src/root.rs
```diff
@@ -327,12 +333,23 @@ impl Command {
...
+            if threshold > keyids as u64 {
+                // Return an error when the "force" flag wasn't set
+                if !force {
+                    return Err(error::Error::UnstableRoot {
+                        role: *roletype,
+                        threshold: threshold,
+                        actual: keyids,
+                    });
+                }
+                // Print out a warning letting the user that the generated file isn't complient with
```
tjkirch: spelling: "compliant"

Can we clarify in the comments the difference between this check/warning and the one a few lines down?

.gitignore
```diff
@@ -1,3 +1,3 @@
-/target
+**/target
```
tjkirch: I don't understand this change.  The tough repo has a Cargo.toml declaring the subdirectories to be in a workspace, and when you build (even from a subdirectory) the results go into `/target` in the root of the repo, not a subdirectory.  What are you doing when you see `target` somewhere else?

arnaldo2792: I ran `cargo run` inside `tuftool/`, and that generated `tuftool/target`:

```bash
Untracked files:
  (use "git add <file>..." to include in what will be committed)
        tough/target/
        tuftool/target/
```

tjkirch: I don't have that behavior - `build` and `run` always put artifacts in the top-level `target`.  Let's talk offline about our setups.

tuftool/tests/root_command.rs
arnaldo2792: There is a test called `append_signature_root`, which initializes the root.json with root threshold 2, and uses `sign_root_json` with `-f`, which covers the the second part of the test that you ask. I'll the missing test case.

.gitignore
arnaldo2792: I re-ran `cargo run` inside `tuftool`, I didn't see the folder. I'll remove the commit that updates the `.gitignore` file.

tuftool/src/root.rs
webern: > I think we should call this something more specific, because "force" could mean overriding any number of checks, and root roles are critical, so it's safer to override specific checks you know are OK to skip. Something like `--ignore-threshold` or `--allow-keys-under-threshold` perhaps?

`--ignore-threshold` works for me (over `--allow-keys-under-threshold`).

tuftool/src/root.rs
```diff
@@ -106,6 +107,9 @@ pub(crate) enum Command {
         ///Optional - Path of older root.json that contains the key-id
         #[structopt(short = "c", long = "cross-sign")]
         cross_sign: Option<PathBuf>,
+        ///Force signing with fewer keys than the specified threshold
+        #[structopt(short = "f")]
```
webern: ...just making sure: this is optional and defaults to `false`, right?

webern (commented):
Looks good. Your CI checks failed due to a clippy lint.

`threshold: threshold` can be re-written as `threshold,` and clippy says you must...

tuftool/src/root.rs
```diff
@@ -352,10 +369,16 @@ impl Command {
...
-                signature_count,
-            });
+            // Return an error when the "force" flag wasn't set
+            if !force {
+                return Err(error::Error::SignatureRoot {
+                    threshold,
+                    signature_count,
+                });
+            }
+            // Print out a warning letting the user that the generated file isn't complient with
```
zmrow: This should probably say "letting the user *know* that the..."

Also - compliant is misspelled. :)

arnaldo2792:
Forced push includes:

- Explanation about the first "quick check"
- Fixed typos
- Changed `force` to  `ignore-threshold`
- Add unit test to check for failures when `-i` isn't present.

arnaldo2792:
Forced push includes:

- Fix linting failure

webern (approved):

tuftool/src/root.rs
tjkirch: I see the top check was clarified to `Quick check that root is signed by enough key IDs, in all its roles`, but I'm still not sure what the lower check is for, that says `Signature check for root`?

tuftool/tests/root_command.rs
```diff
@@ -84,11 +84,19 @@ fn add_key_all_roles(key: &str, root_json: &str) {
 fn sign_root_json(key: &str, root_json: &str) {
     Command::cargo_bin("tuftool")
         .unwrap()
-        .args(&["root", "sign", root_json, "-k", key])
+        .args(&["root", "sign", root_json, "-i", "-k", key])
```
tjkirch, line 88:
```suggestion
        // We don't have enough signatures to meet the threshold, so we have to pass `-i` to succeed
        .args(&["root", "sign", root_json, "-i", "-k", key])
```

tuftool/tests/root_command.rs
```diff
@@ -84,11 +84,19 @@ fn add_key_all_roles(key: &str, root_json: &str) {
...
         .assert()
         .success();
 }
 
+fn sign_root_json_failure(key: &str, root_json: &str) {
+    Command::cargo_bin("tuftool")
+        .unwrap()
+        .args(&["root", "sign", root_json, "-k", key])
```
tjkirch, line 97:
```suggestion
        // We don't have enough signatures to meet the threshold, so we should fail
        .args(&["root", "sign", root_json, "-k", key])
```

tuftool/src/root.rs
tjkirch: That _seems_ to be the case from [structopt docs](https://docs.rs/structopt/0.3.23/structopt/#type-magic)... "`true` if the flag is present"...

tjkirch (approved):

arnaldo2792:
Forced push fixes nits in unit tests

zmrow (approved):
:volleyball:

tjkirch (approved):

PR #425, opened 2021-10-26
Fix parsing of `root.json` file when `ecdsa-sha2-nistp256` keys are used

flavio:
[According to the TUF spec](https://theupdateframework.github.io/specification/v1.0.25/index.html#keytype), the `key_type` value for  ECDSA-sha2-nistp256 keys is `ecdsa-sha2-nistp256`.
    
Prior to this commit, the expected `key_type` value for these keys was `ecdsa`.
    
That made impossible the parsing of `root.json` files using one or more of this type of keys.

### DIFF
--- a/tough/src/schema/key.rs
+++ b/tough/src/schema/key.rs
@@ -58,6 +58,7 @@ pub enum Key {
         _extra: HashMap<String, Value>,
     },
     /// An EcdsaKey
+    #[serde(rename = "ecdsa-sha2-nistp256")]
     Ecdsa {
         /// The Ecdsa key.
         keyval: EcdsaKey,


flavio:
BTW, I ran into this issue while dealing with [Sigstore's TUF repository](https://sigstore-tuf-root.storage.googleapis.com/root.json).

Right now I still cannot parse the repository properly, because the `public` attribute of all the `ecdsa-sha2-nistp256` keys is not PEM encoded (as described by the official spec), but it's hex encoded. I've reached out to the Sigstore folks to understand whether the TUF spec is wrong or their `root.json` is bugged.

trishankatdatadog:
> Right now I still cannot parse the repository properly, because the `public` attribute of all the `ecdsa-sha2-nistp256` keys is not PEM encoded (as described by the official spec), but it's hex encoded. I've reached out to the Sigstore folks to understand whether the TUF spec is wrong or their `root.json` is bugged.

As we discussed in the sigstore Slack, the TUF spec does not mandate any strong requirements on public key values. It makes some suggestions at best, but implementations can differ in how they encode them. For example, python-tuf seems to require PEM-encoded ECDSA pubkeys, but go-tuf seems to encode them the same way as Ed25519 pubkeys.

You have a few options:
1. Be compatible with only go-tuf.
2. Be compatible with only python-tuf (makes no sense for you here).
3. Be compatible with both: slightly more work, but the best option IMHO.

WDYT?

flavio:
I made a quick attempt at implementing the third option proposed by @trishankatdatadog . You can see it here: https://github.com/awslabs/tough/pull/426

webern (approved):
LGTM

https://github.com/theupdateframework/specification/blob/master/tuf-spec.md#file-formats-general-principles--file-formats-general-principles

(BTW, should someone fix the markdown formatting of the spec?)

trishankatdatadog:
> LGTM
> 
> https://github.com/theupdateframework/specification/blob/master/tuf-spec.md#file-formats-general-principles--file-formats-general-principles
> 
> (BTW, should someone fix the markdown formatting of the spec?)

We have much better formatting [here](https://theupdateframework.github.io/specification/latest/index.html#file-formats-general-principles) thanks to @joshuagl

zmrow (approved):
Thanks!

PR #426, opened 2021-10-27
Make ECDSA key parsing more flexible

flavio:
> NOTE WELL: this PR is built on top of https://github.com/awslabs/tough/pull/425

TUF specification suggests to encode ECDSA keys using PEM. This is the behaviour of the python reference implementation.

However, `go-tuf` encodes the keys as Hex numbers.

This commit makes the decoding of ECDSA keys more relaxed, the code will try to decode them as PEM strings and it will resort to Hex when the first decode fails.

### DIFF
--- a/tough/src/schema/de.rs
+++ b/tough/src/schema/de.rs
@@ -86,4 +86,15 @@ mod tests {
         ))
         .is_err());
     }
+
+    /// Ensure that we can deserialize a root.json file that has hex-encoded ECDSA keys. This uses
+    /// sigstore's root.json file taken from here:
+    /// https://sigstore-tuf-root.storage.googleapis.com/2.root.json
+    #[test]
+    fn ecdsa_hex_encoded_keys() {
+        assert!(serde_json::from_str::<Signed<Root>>(include_str!(
+            "../../tests/data/hex-encoded-ecdsa-sig-keys/root.json"
+        ))
+        .is_ok());
+    }
 }
--- a/tough/src/schema/decoded.rs
+++ b/tough/src/schema/decoded.rs
@@ -126,6 +126,32 @@ impl Encode for EcdsaPem {
     }
 }
 
+/// [`Decode`]/[`Encode`] implementation for ECDSA public keys.
+/// This is a flexible implementation, it will try to decode the key assuming
+/// it is PEM encoded, if the decode fails it will then try to decode it
+/// assuming it's Hex encoded.
+/// The official TUF specification suggests ECDSA keys to be PEM encoded,
+/// however the go-tuf implementation encodes them as Hex numbers.
+/// This flexible decoder tries to cover both cases in a transparent way.
+#[derive(Debug, Clone, Copy)]
+pub struct EcdsaFlex {}
+
+impl Decode for EcdsaFlex {
+    fn decode(s: &str) -> Result<Vec<u8>, Error> {
+        if s.starts_with("-----BEGIN ") {
+            EcdsaPem::decode(s)
+        } else {
+            Hex::decode(s)
+        }
+    }
+}
+
+impl Encode for EcdsaFlex {
+    fn encode(b: &[u8]) -> String {
+        EcdsaPem::encode(b)
+    }
+}
+
 // =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
 
 impl<'de, T: Decode> Deserialize<'de> for Decoded<T> {
--- a/tough/src/schema/key.rs
+++ b/tough/src/schema/key.rs
@@ -2,7 +2,7 @@
 
 //! Handles cryptographic keys and their serialization in TUF metadata files.
 
-use crate::schema::decoded::{Decoded, EcdsaPem, Hex, RsaPem};
+use crate::schema::decoded::{Decoded, EcdsaFlex, Hex, RsaPem};
 use crate::schema::error::{self, Result};
 use olpc_cjson::CanonicalFormatter;
 use ring::digest::{digest, SHA256};
@@ -121,7 +121,7 @@ pub enum EcdsaScheme {
 #[derive(Debug, Clone, Deserialize, Serialize, PartialEq)]
 pub struct EcdsaKey {
     /// The public key.
-    pub public: Decoded<EcdsaPem>,
+    pub public: Decoded<EcdsaFlex>,
 
     /// Any additional fields read during deserialization; will not be used.
     #[serde(flatten)]
@@ -204,7 +204,7 @@ impl FromStr for Key {
             } else {
                 Err(KeyParseError(()))
             }
-        } else if let Ok(public) = serde_plain::from_str::<Decoded<EcdsaPem>>(s) {
+        } else if let Ok(public) = serde_plain::from_str::<Decoded<EcdsaFlex>>(s) {
             Ok(Key::Ecdsa {
                 keyval: EcdsaKey {
                     public,
--- a/tough/tests/data/hex-encoded-ecdsa-sig-keys/root.json
+++ b/tough/tests/data/hex-encoded-ecdsa-sig-keys/root.json
@@ -1,0 +1,144 @@
+{
+	"signatures": [
+		{
+			"keyid": "2f64fb5eac0cf94dd39bb45308b98920055e9a0d8e012a7220787834c60aef97",
+			"sig": "3046022100d3ea59490b253beae0926c6fa63f54336dea1ed700555be9f27ff55cd347639c0221009157d1ba012cead81948a4ab777d355451d57f5c4a2d333fc68d2e3f358093c2"
+		},
+		{
+			"keyid": "bdde902f5ec668179ff5ca0dabf7657109287d690bf97e230c21d65f99155c62",
+			"sig": "304502206eaef40564403ce572c6d062e0c9b0aab5e0223576133e081e1b495e8deb9efd02210080fd6f3464d759601b4afec596bbd5952f3a224cd06ed1cdfc3c399118752ba2"
+		},
+		{
+			"keyid": "eaf22372f417dd618a46f6c627dbc276e9fd30a004fc94f9be946e73f8bd090b",
+			"sig": "304502207baace02f56d8e6069f10b6ff098a26e7f53a7f9324ad62cffa0557bdeb9036c022100fb3032baaa090d0040c3f2fd872571c84479309b773208601d65948df87a9720"
+		},
+		{
+			"keyid": "f40f32044071a9365505da3d1e3be6561f6f22d0e60cf51df783999f6c3429cb",
+			"sig": "304402205180c01905505dd88acd7a2dad979dd75c979b3722513a7bdedac88c6ae8dbeb022056d1ddf7a192f0b1c2c90ff487de2fb3ec9f0c03f66ea937c78d3b6a493504ca"
+		},
+		{
+			"keyid": "f505595165a177a41750a8e864ed1719b1edfccd5a426fd2c0ffda33ce7ff209",
+			"sig": "3046022100c8806d4647c514d80fd8f707d3369444c4fd1d0812a2d25f828e564c99790e3f022100bb51f12e862ef17a7d3da2ac103bebc5c7e792237006c4cafacd76267b249c2f"
+		}
+	],
+	"signed": {
+		"_type": "root",
+		"consistent_snapshot": false,
+		"expires": "2022-05-11T19:09:02.663975009Z",
+		"keys": {
+			"2f64fb5eac0cf94dd39bb45308b98920055e9a0d8e012a7220787834c60aef97": {
+				"keyid_hash_algorithms": [
+					"sha256",
+					"sha512"
+				],
+				"keytype": "ecdsa-sha2-nistp256",
+				"keyval": {
+					"public": "04cbc5cab2684160323c25cd06c3307178a6b1d1c9b949328453ae473c5ba7527e35b13f298b41633382241f3fd8526c262d43b45adee5c618fa0642c82b8a9803"
+				},
+				"scheme": "ecdsa-sha2-nistp256"
+			},
+			"b6710623a30c010738e64c5209d367df1c0a18cf90e6ab5292fb01680f83453d": {
+				"keyid_hash_algorithms": [
+					"sha256",
+					"sha512"
+				],
+				"keytype": "ecdsa-sha2-nistp256",
+				"keyval": {
+					"public": "04fa1a3e42f2300cd3c5487a61509348feb1e936920fef2f83b7cd5dbe7ba045f538725ab8f18a666e6233edb7e0db8766c8dc336633449c5e1bbe0c182b02df0b"
+				},
+				"scheme": "ecdsa-sha2-nistp256"
+			},
+			"bdde902f5ec668179ff5ca0dabf7657109287d690bf97e230c21d65f99155c62": {
+				"keyid_hash_algorithms": [
+					"sha256",
+					"sha512"
+				],
+				"keytype": "ecdsa-sha2-nistp256",
+				"keyval": {
+					"public": "04a71aacd835dc170ba6db3fa33a1a33dee751d4f8b0217b805b9bd3242921ee93672fdcfd840576c5bb0dc0ed815edf394c1ee48c2b5e02485e59bfc512f3adc7"
+				},
+				"scheme": "ecdsa-sha2-nistp256"
+			},
+			"eaf22372f417dd618a46f6c627dbc276e9fd30a004fc94f9be946e73f8bd090b": {
+				"keyid_hash_algorithms": [
+					"sha256",
+					"sha512"
+				],
+				"keytype": "ecdsa-sha2-nistp256",
+				"keyval": {
+					"public": "04117b33dd265715bf23315e368faa499728db8d1f0a377070a1c7b1aba2cc21be6ab1628e42f2cdd7a35479f2dce07b303a8ba646c55569a8d2a504ba7e86e447"
+				},
+				"scheme": "ecdsa-sha2-nistp256"
+			},
+			"f40f32044071a9365505da3d1e3be6561f6f22d0e60cf51df783999f6c3429cb": {
+				"keyid_hash_algorithms": [
+					"sha256",
+					"sha512"
+				],
+				"keytype": "ecdsa-sha2-nistp256",
+				"keyval": {
+					"public": "04cc1cd53a61c23e88cc54b488dfae168a257c34fac3e88811c55962b24cffbfecb724447999c54670e365883716302e49da57c79a33cd3e16f81fbc66f0bcdf48"
+				},
+				"scheme": "ecdsa-sha2-nistp256"
+			},
+			"f505595165a177a41750a8e864ed1719b1edfccd5a426fd2c0ffda33ce7ff209": {
+				"keyid_hash_algorithms": [
+					"sha256",
+					"sha512"
+				],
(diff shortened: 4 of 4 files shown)


flavio:
The code works fine against Sigstore's `root.json`, meaning it's not generating an error while parsing the file. However, none of the 5 keys defined inside of the `root.json` file can be used to verify its contents :thinking: 

I tried to understand what is going on, but I couldn't understand it :(

I'm using a local build of the `tuftool` to download the repository:

```console
tuftool download \
   -m https://sigstore-tuf-root.storage.googleapis.com \
   -t https://sigstore-tuf-root.storage.googleapis.com/targets \
   --allow-root-download \
   sigstore-tuf-root
```

tough/src/schema/decoded.rs
```diff
@@ -126,6 +126,35 @@ impl Encode for EcdsaPem {
...
+/// assuming it's Hex encoded.
+/// The official TUF specification suggests ECDSA keys to be PEM encoded,
+/// however the go-tuf implementation encodes them as Hex numbers.
+/// This flexible decoder tries to cover both cases in a transparent way.
+#[derive(Debug, Clone, Copy)]
+pub struct EcdsaFlex {}
+
+impl Decode for EcdsaFlex {
+    fn decode(s: &str) -> Result<Vec<u8>, Error> {
+        match spki::decode(
```
trishankatdatadog: I would do stricter pattern-matching to be safe:

If I seem the PEM header, I will decode PEM. If I see the hex encoding, I will decode it as such. Otherwise, error out. Makes sense?

trishankatdatadog (changes requested):

trishankatdatadog:
> The code works fine against Sigstore's `root.json`, meaning it's not generating an error while parsing the file. However, none of the 5 keys defined inside of the `root.json` file can be used to verify its contents 🤔

Great start, thank you! Not sure why the signature verification is failing, but I'm sure it's missing some cryptographic parameters. Have you tried looking at [the go-tuf implementation](https://github.com/theupdateframework/go-tuf/blob/5908a16987fa24b6b7b60d71e0dfc043783af77c/pkg/keys/ecdsa.go)?

tough/src/schema/decoded.rs
flavio: Yes, I was just trying to get something quick put together... My goal would be to understand why none of the keys can successfully verify the contents of the file :(

flavio:
`tough` uses the [ring](https://briansmith.org/rustdoc/ring/index.html) crate to perform verification. This crate doesn't provide fine tuning of the different parameters, it's pretty straightforward.

I've made the following experiment:
  * I created a small Rust program, that uses `ring` to create a ECDSA keypair, sign a message with that and print the public key, signature and message on the stdout
  * I create a Go program that uses the same code as the one you linked above to verify the signature

Everything worked as expected, the Go program was able to verify the signature produced by the Rust library. If you want I can share the code of the programs.

Now, I know this testing something slightly different: we should produce a signature with Go, verify that with Rust. However I didn't have the time to look into how go-tuf is producing the signature.

However, I find reassuring to see everything work as expected out of the box.

I'm getting really lost on this issue, any help? :pray:

trishankatdatadog:
I'm sure it's something really small/silly, but I don't have enough time to look into this. Let me see what @asraa and I can do with some downtime...

webern:
Thank you for working on this!

Are we sure that eagerly trying both encodings does not introduce a security risk?

For this sort of interoperability functionality it would be great to see a test repository, created by the incompatible go-tuf implementation, added to here https://github.com/awslabs/tough/tree/develop/tough/tests/data and a test added here: https://github.com/awslabs/tough/blob/develop/tough/tests/interop.rs

This might aide your development cycles as well since it should serve as a red/green testing scenario.

trishankatdatadog:
> Are we sure that eagerly trying both encodings does not introduce a security risk?

Agreed, which is why I recommended that we try only one encoding depending on the format.

asraa:
I just tried the reverse check, where I created a ecdsa signature with go-tuf (underlyingly using crypto/ecdsa SignASN1) and it verified properly with `ring::signature`, the code here.

I'm skeptical now that the message isn't being canonicalized the same way before checking signature validation.

flavio:
Thanks for having looked into that!

> I'm skeptical now that the message isn't being canonicalized the same way before checking signature validation.

I'm starting to have doubts too about that. I've recently used the [olpc-cjson](https://github.com/awslabs/tough/tree/develop/olpc-cjson) crate of tough while verifying a cosign bundle returned by Rekor and it worked. Dunno if Rekor is using the same canonicalization code as go-tuf...

asraa:
> while verifying a cosign bundle returned by Rekor and it worked

Nope, the bundle is just a normal marshalled json struct. The TUF code is using a cjson canonicalizer package, for the metadata

Edit: the easiest way for me to test this is to disable any funky indentations we do to make the json nice and see if tuftool can handle a go-tuf generated manifest

flavio:
> Nope, the bundle is just a normal marshalled json struct. The TUF code is using a cjson canonicalizer package, for the metadata

Yes, but then `cosign` internally converts it to cjson before verifying the signature (https://github.com/sigstore/cosign/blob/main/pkg/cosign/verify.go#L335)

flavio:
> Edit: the easiest way for me to test this is to disable any funky indentations we do to make the json nice and see if tuftool can handle a go-tuf generated manifest

Cool, I'm totally new to TUF and its tooling. It would be great if you could look into that. If not, let me know and I'll experiment :)

trishankatdatadog:
> Dunno if Rekor is using the same canonicalization code as go-tuf...

Probably not

flavio:
@trishankatdatadog, @asraa : I finally figured out what is causing tough to choke on Sigstore's TUF reposotory.

The root file is violating the TUF specification. All the date-time objects have been serialized using RFC 3339, which includes the timezone information.
The [TUF specification](https://theupdateframework.github.io/specification/v1.0.26/index.html#date-time) clearly states:

> Metadata date-time follows the ISO 8601 standard. The expected format of the combined date and time string is "YYYY-MM-DDTHH:MM:SSZ". Time is always in UTC, and the "Z" time zone designator is attached to indicate a zero UTC offset. An example date-time string is "1985-10-21T01:21:00Z".

This is what the Sigstore TUF repository root file has inside of it:
```json
"expires": "2021-12-18T13:28:12.99008-06:00"
```

The tough crate deserializes this data into a series of Rust structs, which are rightfully using a `DateTime<Utc>` as internal representation. During this process the Timezone information is lost.

Later, the Rust structs are serialized, during this step the date time object becomes:

```json
 "expires": "2021-12-18T19:28:12.990080Z",
```

The document produced during this process is then used to verify the validity of the signatures of the trusted keys.

The document signed by the Sigstore maintainers is different from the one the Rust client verifies due to this different formatting of the date-time objects. As a result, none of the signatures is verified -> the download stops.

I did a quick fix to address this issue (see commit f06b7a226eb90096889dd90e422f5875c4af9c4c). With this patch I can successfully use the `tuftool` to download the Sigstore TUF repository:

```console
tuftool download -m 'https://sigstore-tuf-root.storage.googleapis.com/' -t 'https://sigstore-tuf-root.storage.googleapis.com/targets' --allow-root-download sigstore-tuf-root`
=================================================================
WARNING: Downloading root.json to /home/flavio/hacking/sigstore/tough/tuftool/1.root.json
This is unsafe and will not establish trust, use only for testing
=================================================================
Downloading targets to "sigstore-tuf-root"
	-> fulcio_v1.crt.pem
	-> ctfe.pub
	-> fulcio.crt.pem
	-> artifact.pub
	-> rekor.pub
```

However, I think my fix is not needed. The problem must be fixed inside of the Sigstore TUF repository. 

This PR now includes two commits:
  * 07228f29d9c795d01f42b264bdfd2035971721cd - we can address the review feedback expressed above OR we can make the TUF specification more explicit about how ECDSA keys must be encoded (right now there's just a recommendation) and fix Sigstore's TUF repository metadata accordingly
  * f06b7a226eb90096889dd90e422f5875c4af9c4c - I would not merge it because the TUF specification is clear about how datetime objects should be formatted. The issue has to be fixed inside of Sigstore's TUF repository

asraa:
> However, I think my fix is not needed. The problem must be fixed inside of the Sigstore TUF repository.

Joshua also pointed this out, this was fixed in the next v2 root (see https://github.com/sigstore/root-signing/blob/9a627a0fc51c202868e0ea10b74007fa4b9e9416/repository/repository/2.root.json#L27). 

Why does expiration verification need to happen on 1.root.json? 

Also I'm sorry that you were still working on this! It turns out that also the signature verification was fixed in v2 and other issues for tuftool https://github.com/sigstore/root-signing/pull/51

EDIT: I think we will always need to pin trust on the initial root being 2.root.json for rust since the sign verification won't work anyway on 1.root.json. This works without any patch when I was testing v2 root signing.
```
tuftool download -m 'https://sigstore-tuf-root.storage.googleapis.com/' -t 'https://sigstore-tuf-root.storage.googleapis.com/targets' --root 2.root.json test/
Downloading targets to "test/"
```

trishankatdatadog:
> Why does expiration verification need to happen on 1.root.json?

Correct: you should [ignore](https://theupdateframework.github.io/specification/v1.0.26/index.html#update-root) expiration of any interim roots...

trishankatdatadog:
> EDIT: I think we will always need to pin trust on the initial root being 2.root.json for rust since the sign verification won't work anyway on 1.root.json. This works without any patch when I was testing v2 root signing.

Agree: this is the simplest fix for now: to bundle v2 with any tough-based client that will fetch/verify targets.

webern:
> the TUF specification is clear about how datetime objects should be formatted

I wonder if we are still going to have issues based on different precisions in fractions of a second, e.g.
- `"expires": "2021-12-18T19:28:12.990080Z"`
- `"expires": "2021-12-18T19:28:12.9900800Z"`

What you quoted from the specification might imply that fractions of a second must not be included, but it doesn't say that explicitly.

I think it makes sense to remove the time serialization changes from this PR and to focus on the ECDSA decoding issue. This may preclude us from adding an interoperability test as I recommended in this PR until the time format issues are addressed.

trishankatdatadog:
> What you quoted from the specification might imply that fractions of a second must not be included, but it doesn't say that explicitly.

Might be a good idea to clarify this in the spec either way. PRs are welcome!

flavio:
Hi everybody, I'm back from the End Year holidays and I'm ready to get back to this issue :muscle: 

I've clarified the situation with @trishankatdatadog. The date format issues inside of the Sigstore TUF repositories have been sorted out (thanks @asraa !). 
That means there's just one issue to be addressed: make the parsing of the ECDSA keys more flexible.

My initial code works, but as @trishankatdatadog and @webern pointed out (see [[1]](https://github.com/awslabs/tough/pull/426#discussion_r737681994) and [[2]](https://github.com/awslabs/tough/pull/426#issuecomment-954054718)), we should do things in a more secure way.

Both @trishankatdatadog and @webern suggested to check the string and:
  * If the PEM header is found -> handle it as PEM format
  * If the string is hexadecimal -> handle as Hex format
  * Otherwise just report an error

My initial thoughts have been:
  * To check if the string is hexadecimal: validate the string using a regular expression
  * To check if the string is PEM encoded: use Rust's `str.starts_with` and look for the `-----BEGIN ` string

I'm not convinced by both approaches. I fear they might not be secure... Do you have better recommendations?

I could also change the current code like that:
  * Attempt to decode the input string assuming it's PEM encoded
  * Check the error type raised by the function
    * if it's a PEM encoding error then do some other code (attempt a straight hex conversion, or do hex conversion only after we know for sure that the input is hexadecimal)
    * If it's not a PEM encoding error -> bubble up the error

Once this aspect is clarified I'll go ahead and implement the new code and extend the test suite as recommended by @webern

webern:
> * To check if the string is hexadecimal: validate the string using a regular expression

This might not be necessary. Maybe you could do the starts_with check for PEM, and if it's not PEM then assume it is hexadecimal. Something didn't sit right with me (earlier) about trying, failing, and trying the other. But once you've determined that it's not PEM, we can assume it is hexadecimal and it won't parse as such if it would have failed a regex anyway.

flavio:
@webern thanks for the feedback. I've updated the PR to reflect what you suggested. I've also added a new unit test covering this specific case.

tough/src/schema/decoded.rs
```diff
@@ -126,6 +126,36 @@ impl Encode for EcdsaPem {
...
+pub struct EcdsaFlex {}
+
+impl Decode for EcdsaFlex {
+    fn decode(s: &str) -> Result<Vec<u8>, Error> {
+        if s.starts_with("-----BEGIN ") {
+            spki::decode(
+                spki::OID_EC_PUBLIC_KEY,
+                Some(spki::OID_EC_PARAM_SECP256R1),
+                s,
+            )
```
webern: Can this re-use EcdsaPem::decode?
```suggestion
            EcdsaPem::decode(s)
```

tough/src/schema/decoded.rs
```diff
@@ -126,6 +126,36 @@ impl Encode for EcdsaPem {
...
+/// it is PEM encoded, if the decode fails it will then try to decode it
+/// assuming it's Hex encoded.
+/// The official TUF specification suggests ECDSA keys to be PEM encoded,
+/// however the go-tuf implementation encodes them as Hex numbers.
+/// This flexible decoder tries to cover both cases in a transparent way.
+#[derive(Debug, Clone, Copy)]
+pub struct EcdsaFlex {}
+
+impl Decode for EcdsaFlex {
+    fn decode(s: &str) -> Result<Vec<u8>, Error> {
```
webern, line 140: Hmm, unfortunately the trait definition doesn't allow us to know which type of encoding was encountered. If I understand this correctly it means that if we deserialize a key with hex encoding we would later serialize it with PEM encoding.

webern (commented):
The first thing I noticed is that, due to our trait definition, the information about what encoding was encountered is lost. This got me thinking about whether there is a general change that should be made to the trait to describe what encoding format was encountered. This led me to realize I don't know much about key encoding formats so I started poking around...

Then I thought, well I need to look at the spec, and... uh oh, It appears that:
- `ed25519`: PUBLIC :: 64-byte hex encoded string.
- `ecdsa-sha2-nistp256`: PUBLIC :: PEM format and a string.

If I'm understanding things correctly, that means the root.json from your test case is not valid per the spec because it states `ecdsa-sha2-nistp256` and uses a hex encoding.

I apologize for missing this earlier, but are we trying to make the `tough` library accept root.json files that are not formed correctly per the specification?

Here's the section that I'm looking at: 

https://github.com/theupdateframework/specification/blob/c43045a85a637a5e2922af2dc1af08691e496780/tuf-spec.md#file-formats-general-principles--file-formats-general-principles

I guess the spec also says that

> adopters can define and use any particular keytype, signing scheme, and cryptographic library

but I would interpret that to mean that you can create new `keytype` strings, not that you can use an existing `keytype` differently than what the spec defines.

Help me out here if I'm behind the curve! Thanks.

trishankatdatadog:
> I guess the spec also says that
> 
> > adopters can define and use any particular keytype, signing scheme, and cryptographic library
> 
> but I would interpret that to mean that you can create new `keytype` strings, not that you can use an existing `keytype` differently than what the spec defines.

No, it means that you are free to redefine to use different schemas for the proposed `keytype`-s.

If this is unclear or undesirable, please do open a PR upstream (cc @lukpueh @joshuagl)

flavio:
> If I'm understanding things correctly, that means the root.json from your test case is not valid per the spec because it states `ecdsa-sha2-nistp256` and uses a hex encoding.

Good catch, I didn't notice that.
 
> I apologize for missing this earlier, but are we trying to make the `tough` library accept root.json files that are not formed correctly per the specification?

That was not my intention.

> Help me out here if I'm behind the curve! Thanks.

As @trishankatdatadog suggested, I've submitted https://github.com/theupdateframework/specification/issues/201 to shed some light on that.

webern:
> > If I'm understanding things correctly, that means the root.json from your test case is not valid per the spec because it states `ecdsa-sha2-nistp256` and uses a hex encoding.
> 
> Good catch, I didn't notice that.
> 
> > I apologize for missing this earlier, but are we trying to make the `tough` library accept root.json files that are not formed correctly per the specification?
> 
> That was not my intention.

Actually per discussion in theupdateframework/specification#201, it sounds like the spec allows for your root.json to be considered valid. Did you create it with go-tuf? In general, is it go-tuf that we aim to interoperate with by making these changes?

I'm in the `cloud-native.slack.com` as `@Matt Briggs` if you want to DM me on this. The main thing missing from my brain with your implementation is what happens when we re-serialize that root.json. Do we encode the key back as a PEM even though when we deserialized it was hex? I can't quite see the answer to that question by inspecting the traits, so I'm going to check out your branch and try it to see what happens.

webern:
Ok, this did at least reserialize the keys in their original hex encoding. I think I want to take it further though and have an example repository created by go-tuf, with targets and metadata, that we can load, add a target to, re-sign, then load with go-tuf. i.e. true interop. Let's chat about it via DM soon.

```rust
fn hex_encoded_ecdsa_sig_keys() {
    let path = test_data()
        .join("hex-encoded-ecdsa-sig-keys")
        .join("root.json");
    let mut f = File::open(path).unwrap();

    let root: Signed<Root> = serde_json::from_reader(f).unwrap();
    println!("{}", serde_json::to_string_pretty(&root).unwrap())
}
```

tough/src/schema/decoded.rs
webern: > Hmm, unfortunately the trait definition doesn't allow us to know which type of encoding was encountered. If I understand this correctly it means that if we deserialize a key with hex encoding we would later serialize it with PEM encoding.

I'm less concerned with this since @flavio doesn't need to write hex-encoded Ecds keys anyway. Right now we just need to support reading them. Also, I believe reserializing works fine because the `Decoded<T>` struct has the original string in it.

tough/src/schema/de.rs
```diff
@@ -86,4 +86,12 @@ mod tests {
         ))
         .is_err());
     }
+
+    #[test]
+    fn ecdsa_hex_encoded_keys() {
```
webern, line 94:
```suggestion
    /// Ensure that we can deserialize a root.json file that has hex-encoded ECDSA keys. This uses
    /// sigstore's root.json file taken from here:
    /// https://sigstore-tuf-root.storage.googleapis.com/2.root.json
    #[test]
    fn ecdsa_hex_encoded_keys() {
```

webern (commented):
This looks good @flavio, sorry it took me so long. The only recommended changes are to document the test's purpose and to re-use EcdsaPem::decode in EcdsaFlex::decode

webern:
Needs a `git rebase develop` to pick up some clippy fixes.

flavio:
@webern I've rebased my PR against `develop`. The PR is now ready for review

webern (approved):
Nice :rocket:

zmrow (approved):
Thanks again for all the work on this! 

:+1:

trishankatdatadog (approved):
LGTM, thanks!

jku:
> What you quoted from the specification might imply that fractions of a second must not be included, but it doesn't say that explicitly.

Spec is explicit about time formats ( this is not in the expiry sections themselves but it is in there):

> Metadata [date-time](https://theupdateframework.github.io/specification/latest/#date-time) follows the ISO 8601 standard. The expected format of the combined date and time string is "YYYY-MM-DDTHH:MM:SSZ". Time is always in UTC, and the "Z" time zone designator is attached to indicate a zero UTC offset. An example date-time string is "1985-10-21T01:21:00Z".

This is very clear: timezones and microseconds are not allowed

webern:
> This is very clear: timezones and microseconds are not allowed

I created #447 for this.

PR #435, opened 2022-01-04
Upgrade SNAFU to 0.7

shepmaster:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1571,9 +1571,9 @@ checksum = "c307a32c1c5c437f38c7fd45d753050587732ba8628319fbdf12a7e289ccc590"
 
 [[package]]
 name = "snafu"
-version = "0.6.10"
+version = "0.7.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "eab12d3c261b2308b0d80c26fffb58d17eba81a4be97890101f416b478c79ca7"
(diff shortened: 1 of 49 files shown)


shepmaster:
Ran rustfmt and force-pushed.

jpculp:
🎸 Thank you again for taking care of this!

It looks like there is a small clippy warning about a redundant closure, but other than that I think this is great! Would you prefer to fix it or would you rather we take care of it?

zmrow (approved):
:+1: 

LGTM minus clippy's little tantrum.

Thanks again for doing this!

webern (approved):
Wow, this is a very interesting update to the Snafu library. Adding the `*Snafu` suffix should help out IDE tooling quite a bit, although JetBrains *almost* had it handled by now. The ergonomics of having a ready type `Whatever` and for making stringly types on the fly are very interesting. I'll have to try these out as a replacement for `anyhow` when my complaint is "I don't want to deal with defining variants."

Thank you for updating us with the new version.

shepmaster:
The Clippy lint appears unrelated to these changes, so I added it as a separate commit before the change to SNAFU and force-pushed the branch.

jpculp (approved):
🚀

cbgbt (approved):
Thanks for doing this!

cbgbt (approved):

shepmaster:
Thanks for using SNAFU!

PR #451, opened 2022-03-24
do not pin tokio version in Cargo.toml

webern:
*Issue #, if available:*

N/A

*Description of changes:*

I ran into some issues when attempting to update some Bottlerocket Rust dependencies (in the tools workspace). Generally speaking, we should not lock a dependency minor version in a library that is to be consumed by other projects. Doing so can cause unresolvable build issues if a customer's dependencies lock to a different minor version.

No versions were changed in Cargo.lock in this PR, we just changing Cargo.toml to allow cargo to do its thing when others are using the library.

### DIFF
--- a/tough-kms/Cargo.toml
+++ b/tough-kms/Cargo.toml
@@ -21,7 +21,7 @@ rusoto_core = { version = "0.47", optional = true, default-features = false }
 rusoto_credential = { version = "0.47", optional = true }
 rusoto_kms = { version = "0.47", optional = true, default-features = false }
 snafu = { version = "0.7", features = ["backtraces-impl-backtrace-crate"] }
-tokio = "~1.8"  # LTS
+tokio = "1.8"
 pem = "1.0.2"
 
 [dev-dependencies]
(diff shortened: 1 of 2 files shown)


tuftool/Cargo.toml
```diff
@@ -35,7 +35,7 @@ simplelog = "0.11"
 snafu = { version = "0.7", features = ["backtraces-impl-backtrace-crate"] }
 structopt = "0.3"
 tempfile = "3.3.0"
-tokio = "~1.8"  # LTS
+tokio = "1.8"
```
bcressey: Since it's a binary target, it seems like we should leave `tuftool` pinned to ensure that it gets the dependencies we expect.

arnaldo2792 (approved):

tuftool/Cargo.toml
webern: We could. I'll make the change.

webern:
^ undo tuftool change

jpculp (approved):

PR #464, opened 2022-05-11
feat: add workflow responsible for notifying of new TUF spec releases

rdimitrov:
Signed-off-by: Radoslav Dimitrov <dimitrovr@vmware.com>

*Description of changes:*

The following PR adds a workflow that keeps track and creates an issue for notifying of new TUF specification releases.

The version (v1.0.29) of the spec which is currently used for the verification is added without verifying if this is indeed the version that the `tough` project is compliant with. In that sense, please advise which version to use and I'll update the PR accordingly.

**Note:** Dependent on - https://github.com/theupdateframework/specification/pull/224

### DIFF
--- a/.github/workflows/specification-version-check.yml
+++ b/.github/workflows/specification-version-check.yml
@@ -1,0 +1,14 @@
+on:
+  schedule:
+    - cron: "0 13 * * *"
+  workflow_dispatch:
+name: Specification version check
+jobs:
+  # Get the latest TUF specification release and open an issue (if needed)
+  specification-bump-check:
+    permissions:
+      contents: read
+      issues: write
+    uses: theupdateframework/specification/.github/workflows/check-latest-spec-version.yml@master
+    with:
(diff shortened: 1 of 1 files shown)


webern:
This seems like a good idea to me, any thoughts @zmrow or @iliana ?

iliana:
Seems fine. I also have no idea if tough is v1.0.29-compliant, so maybe we want to figure out the delta of what the spec does to what tough does?

webern:
> no idea if tough is v1.0.29-compliant

True, but it's as good a tag as any for a starting point I suppose. This way at least we'll generate some issues telling us to go figure it out. We can comment on the first one and say we don't know quite which patch version is our starting point.

webern (approved):

rdimitrov:
Hey, just so you know the dependent PR is now merged in the specification repo 👍

zmrow (approved):
Seems fine

Issue #428, opened 2021-11-04
tough-ssm and tough-kms: use the new aws rust sdk


webern:
The new Rust SDK is available. When this is no longer considered alpha/beta, we should switch to it because rusoto is in maintenance mode.

- https://crates.io/crates/aws-sdk-ssm
- https://crates.io/crates/aws-sdk-kms

jpculp:
#469 was merged.

Issue #213, opened 2020-08-13
Switch to async / std::future::Future across codebase


zmrow:
`tough-ssm` creates its own `tokio` runtime because it needs to use the `rusoto` which is async.  However, this causes a problem if you're trying to use `tough-ssm` anywhere else that is async... (because multiple runtimes)

We've seen errors like this:
```
...
thread 'main' panicked at 'Cannot start a runtime from within a runtime. 
This happens because a function (like `block_on`) attempted to block the current thread 
while the thread is being used to drive asynchronous tasks.', 
...
```

We should probably change the `KeySource` trait to be async using [the `async-trait` crate](https://crates.io/crates/async-trait)

iliana:
It would probably be more accurate to say "convert tough to async / `std::future::Future` code" than to identify tough-ssm as what needs to change :)

iliana:
@webern can I grab this issue as well? The new `Datastore` trait proposed in #417 should probably also be `#[async_trait]`.

iliana:
(I guess more specifically: I am rashly assuming this is still something we want to do before tough 1.0; is that correct?)

webern:
Yes, we should do this.

Now that I understand the Rust `async` world slightly better, and since `reqwest::blocking` is very problematic (since it instantiates a tokio runtime), here's what I think we should/shouldn't do:
* We **should** change over (pretty much) all existing functions and traits to async ASAP. (This will happen naturally because of the [red function/blue function](https://journal.stuffwithstuff.com/2015/02/01/what-color-is-your-function/) nature of async)
* We should **not** attempt to provide a backward-compatible blocking API. (Based on the reqwest example it seems to do so is fraught and makes foot-guns hard to see)

We should get input from @tjkirch and @zmrow on these points.

iliana:
Progress update: I have tough building, except for the editor module. `trait TargetsWalker`'s use of a Fn trait for what will become an async function is going to take some refactoring. https://github.com/awslabs/tough/blob/446615b715f01afa1bf2149db5a3f38b70a82e66/tough/src/editor/signed.rs#L592-L600

webern:
Hi @iliana it sounds like you got pretty close with this. I wonder if there is a way to collaborate on it. By its nature, a partial implementation cannot be merged, so it's tricky.

iliana:
I'll push my WIP branch later today, I shifted onto using tough as-is (synchronously) for expediency's sake a few weeks back.

webern:
[...]

PR #469, opened 2022-06-09
Aws sdk rust

rpkelly:
*Issue #, if available:*
https://github.com/awslabs/tough/issues/428

*Description of changes:*
* Changed tough-ssm to use aws sdk rust
* Changed tough-kms to use aws sdk rust
* Changed tuftool to use aws sdk rust
* Updated root directory to support above changes

### DIFF
--- a/tough-kms/Cargo.toml
+++ b/tough-kms/Cargo.toml
@@ -9,24 +9,25 @@ keywords = ["TUF", "KMS"]
 edition = "2018"
 
 [features]
-default = ["rusoto"]
-rusoto = ["rusoto-rustls"]
-rusoto-native-tls = ["rusoto_core/native-tls", "rusoto_credential", "rusoto_kms/native-tls"]
-rusoto-rustls = ["rusoto_core/rustls", "rusoto_credential", "rusoto_kms/rustls"]
+default = ["aws-sdk-rust"]
+aws-sdk-rust = ["aws-sdk-rust-rustls"]
+aws-sdk-rust-tls = ["aws-config/native-tls", "aws-sdk-kms/native-tls"]
+aws-sdk-rust-rustls = ["aws-config/rustls", "aws-sdk-kms/rustls"]
 
 [dependencies]
 tough = { version = "0.12.2", path = "../tough", features = ["http"] }
 ring = { version = "0.16.16", features = ["std"] }
-rusoto_core = { version = "0.48", optional = true, default-features = false }
-rusoto_credential = { version = "0.48", optional = true }
-rusoto_kms = { version = "0.48", optional = true, default-features = false }
+aws-sdk-kms = "0.16.0"
+aws-config = "0.46.0"
 snafu = { version = "0.7", features = ["backtraces-impl-backtrace-crate"] }
-tokio = "1.8"
+tokio = { version = "1", features = ["fs", "io-util", "time", "macros", "rt-multi-thread"] }
 pem = "1.0.2"
 
 [dev-dependencies]
+aws-smithy-client = {version = "0.46.0", features = ["test-util"]}
+aws-smithy-http = "0.46.0"
 base64 = "0.13"
 bytes = "1"
-rusoto_mock = { version = "0.48", default-features = false }
+http = "0.2"
 serde = "1.0.125"
 serde_json = "1.0.63"
--- a/tough-kms/src/client.rs
+++ b/tough-kms/src/client.rs
@@ -2,32 +2,43 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
 use crate::error::{self, Result};
-use rusoto_core::{HttpClient, Region};
-use rusoto_credential::ProfileProvider;
-use rusoto_kms::KmsClient;
+use aws_sdk_kms::Client as KmsClient;
 use snafu::ResultExt;
-use std::str::FromStr;
+use std::thread;
 
 /// Builds a KMS client for a given profile name.
 pub(crate) fn build_client_kms(profile: Option<&str>) -> Result<KmsClient> {
-    Ok(if let Some(profile) = profile {
-        let mut provider = ProfileProvider::new().context(error::RusotoCredsSnafu)?;
-        provider.set_profile(profile);
-        let region = provider
-            .region_from_profile()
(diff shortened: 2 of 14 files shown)


tough-kms/src/client.rs
```diff
@@ -2,32 +2,31 @@
...
-            HttpClient::new().context(error::RusotoTlsSnafu)?,
-            provider,
-            match region {
-                Some(region) => {
-                    Region::from_str(&region).context(error::RusotoRegionSnafu { region })?
-                }
-                None => Region::default(),
-            },
-        )
+    let config = aws_config::from_env();
```
webern, line 25: Does `from_env` already use the profile if environment variables are not present? Although the behavior was previously not commented, I think it would help to have some comments here explaining why `from_env` happens first, then things are overridden by the profile.

tough-kms/src/client.rs
```diff
@@ -2,32 +2,31 @@
...
+            )
+            .load()
     } else {
-        KmsClient::new(Region::default())
-    })
+        config.load()
+    };
+    let client_config = tokio::runtime::Runtime::new()
+        .context(error::RuntimeCreationSnafu)?
+        .block_on(client_config_future);
```
webern: To actually make this safe, I think we need to first create a thread, then inside that thread we create the tokio runtime and block on.

Basically the problem here is that someone could call this function from within their tokio runtime, which would panic: https://stackoverflow.com/questions/62536566/how-can-i-create-a-tokio-runtime-inside-another-tokio-runtime-without-getting-th

The real solution is to make this trait, and all of `tough`, async: https://github.com/awslabs/tough/issues/213

tough-kms/src/lib.rs
```diff
@@ -83,10 +86,10 @@ impl KeySource for KmsKeySource {
...
         // Get the public key from AWS KMS
-        let fut = kms_client.get_public_key(rusoto_kms::GetPublicKeyRequest {
-            key_id: self.key_id.clone(),
-            ..rusoto_kms::GetPublicKeyRequest::default()
-        });
+        let fut = kms_client
+            .get_public_key()
+            .key_id(self.key_id.clone())
+            .send();
         let response = tokio::runtime::Runtime::new()
```
webern, line 93: This possible panic already existed, so I guess we don't need to change this, but above I have suggested using a thread to avoid the multiple tokio runtime panic problem

tough-kms/Cargo.toml
```diff
@@ -9,24 +9,25 @@ keywords = ["TUF", "KMS"]
...
 tough = { version = "0.12.2", path = "../tough", features = ["http"] }
 ring = { version = "0.16.16", features = ["std"] }
-rusoto_core = { version = "0.48", optional = true, default-features = false }
-rusoto_credential = { version = "0.48", optional = true }
-rusoto_kms = { version = "0.48", optional = true, default-features = false }
+aws-sdk-kms = "0.12.0"
+aws-config = "0.12.0"
 snafu = { version = "0.7", features = ["backtraces-impl-backtrace-crate"] }
-tokio = "1.8"
+tokio = { version = "1", features = ["full"] }
```
webern: Do we actually need full features?

tough-ssm/src/client.rs
```diff
@@ -1,33 +1,32 @@
...
-            HttpClient::new().context(error::RusotoTlsSnafu)?,
-            provider,
-            match region {
-                Some(region) => {
-                    Region::from_str(&region).context(error::RusotoRegionSnafu { region })?
-                }
-                None => Region::default(),
-            },
-        )
+    let config = aws_config::from_env();
```
webern, line 26: Request same documentation here about why we do `from_env` first then override values with the profile.

tough-ssm/src/lib.rs
```diff
@@ -58,15 +58,17 @@ impl KeySource for SsmKeySource {
...
+            .put_parameter()
+            .name(self.parameter_name.to_owned())
+            .description(key_id_hex.to_owned())
+            .set_key_id(self.key_id.as_ref().cloned())
+            .overwrite(true)
+            .set_type(Some(aws_sdk_ssm::model::ParameterType::SecureString))
+            .value(value.to_owned())
+            .send();
+
         tokio::runtime::Runtime::new()
```
webern, line 72: As I mentioned elsewhere, we should consider spawning a thread before doing this to avoid runtime-in-runtime panic, but I'm don't know that we need to change existing cases of this in the current PR.

webern (commented):
nice work

tough-kms/src/client.rs
rpkelly: https://docs.rs/aws-config/0.13.0/aws_config/fn.from_env.html it creates a ConfigLoader, which is a builder for the SdkConfig. Looking at the source, it's the same as `ConfigLoader::default()`

https://docs.rs/aws-config/0.13.0/aws_config/struct.ConfigLoader.html `.from_env()` is how the docs always refer to creating the builder for the SdkConfig, by default it looks in your environment and default profile, and the overrides are if you're doing anything else.

tough-kms/tests/all_test.rs
```diff
@@ -15,7 +10,7 @@ use tough::schema::key::Key;
 use tough_kms::KmsKeySource;
 use tough_kms::KmsSigningAlgorithm::RsassaPssSha256;
 
-/// Deserialize base64 to `bytes::Bytes`
+/// Deserialize base64 to `byt/es::Bytes`
```
jpculp:
```suggestion
/// Deserialize base64 to `bytes::Bytes`
```
Is this a typo?

jpculp (approved):

etungsten (approved):
Is there anything we can do to test these changes? I believe at a minimum you should try this out in a Bottlerocket host (after https://github.com/bottlerocket-os/bottlerocket/pull/2300 is merged) and see the update flow still works.

tough-kms/src/client.rs
etungsten: ~~I think Matt meant adding code comments for the explanation.~~

edit: Since `ConfigLoader` does document that the default chain implementation goes through `from_env` first, I think it's ok to document that here.

etungsten (approved):
Richard pointed out to me offline that github actions runs the integration tests. After looking through the integration test, I'm satisfied with the coverage.

PR #514, opened 2022-09-26
Update dependencies

jpculp:
*Description of changes:*

Updates various dependencies, including AWS SDK for Rust (0.18.0) and `cargo-deny` (0.12.2).

Replaced a deprecated field (**CustomerMasterKeySpecCustomerMasterKeySpec**) in `tough-kms`

*Testing done:*
Downloaded a Bottlerocket OVA with `tuftool`.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -9,12 +9,13 @@ default = "deny"
 # We want really high confidence when inferring licenses from text
 confidence-threshold = 0.93
 
+# Commented license types are allowed but not currently used
 allow = [
     "Apache-2.0",
-    #"BSD-2-Clause", # OK but currently unused; commenting to prevent warning
+    # "BSD-2-Clause",
     "BSD-3-Clause",
     "BSL-1.0",
-    #"CC0-1.0",  # OK but currently unused; commenting to prevent warning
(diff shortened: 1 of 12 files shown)


ecpullen (approved):
LGTM

stmcginnis (approved):
Looks fine to me and all checks have passed.

deny.toml
```diff
@@ -58,11 +59,8 @@ multiple-versions = "deny"
 wildcards = "deny"
 
 skip-tree = [
-    { name = "h2", version = "0.3.11" }, 
-    # aws sdk depends on hyper, which pulls in an older version of itoa
-    { name = "hyper", version = "0.14.16" },
```
arnaldo2792, line 63: Nit: Probably an explanation in the commit message about why this was removed could be helpful for reviewers :+1:

arnaldo2792 (approved):

deny.toml
jpculp: It might be a bit noisy for a commit message, but it's probably a good idea to call it out at least in the PR. In this instance the SDK is no longer pulling those older versions so the exceptions are no longer required.

PR #516, opened 2022-09-27
Replace structopt with clap in tuftool

jpculp:
*Description of changes:*

`structopt` has been superseded by `clap`, which is more or less a drop-in replacement.

The only functional change is that the short arg for delegated **path-hash-prefixes** (previously 'hp') is now 'x'.

*Testing done:*

Downloaded a Bottlerocket OVA with `tuftool`.

### DIFF
--- a/tuftool/src/add_role.rs
+++ b/tuftool/src/add_role.rs
@@ -6,81 +6,81 @@ use crate::datetime::parse_datetime;
 use crate::error::{self, Result};
 use crate::source::parse_key_source;
 use chrono::{DateTime, Utc};
+use clap::Parser;
 use snafu::{OptionExt, ResultExt};
 use std::num::NonZeroU64;
 use std::path::PathBuf;
-use structopt::StructOpt;
 use tough::editor::{targets::TargetsEditor, RepositoryEditor};
 use tough::key_source::KeySource;
 use tough::schema::{PathHashPrefix, PathPattern, PathSet};
 use url::Url;
 
-#[derive(Debug, StructOpt)]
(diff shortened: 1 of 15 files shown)


ecpullen (approved):
LGTM!

tuftool/src/add_role.rs
```diff
@@ -6,81 +6,81 @@ use crate::datetime::parse_datetime;
...
     outdir: PathBuf,
 
     /// The delegated paths
-    #[structopt(short = "p", long = "paths", conflicts_with = "path-hash-prefixes")]
+    #[clap(short = 'p', long = "paths", conflicts_with = "path-hash-prefixes")]
     paths: Option<Vec<PathPattern>>,
 
     /// The delegated paths hash prefixes
-    #[structopt(short = "hp", long = "path-hash-prefixes")]
+    #[clap(short = 'x', long = "path-hash-prefixes")]
```
stmcginnis, line 62: Seems like we should call this out somewhere. But I don't know where that would be. Since we're the main consumer, it probably isn't too important though.

stmcginnis (approved):
lgtm

tuftool/src/add_role.rs
jpculp: I called it out in the commit message, but we should also mention it in the next release notes. My understanding is delegated paths don't really get used though. I also suspect "hp" was never a valid short (being that it was two characters).

arnaldo2792 (approved):

jpculp:
Just a rebase.

arnaldo2792 (approved):

PR #544, opened 2022-12-06
clippy and cargo update

webern:
*Issue #, if available:*

N/A

*Description of changes:*

Fix or ignore new clippy lints and update dependencies. Restores develop (currently build is broken).

*Testing*

CI only.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -19,9 +19,9 @@ checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
 name = "aho-corasick"
-version = "0.7.19"
+version = "0.7.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b4f55bd91a0978cbfd91c457a164bab8b4001c833b7f323132c0a4e1922dd44e"
+checksum = "cc936419f96fa211c1b9166887b38e5e40b19958e5b895be7c1f93adec7071ac"
 dependencies = [
  "memchr",
 ]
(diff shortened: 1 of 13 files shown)


Cargo.lock
```diff
@@ -48,11 +48,11 @@ dependencies = [
 
 [[package]]
 name = "assert_cmd"
-version = "2.0.5"
+version = "2.0.7"
```
jpculp, line 51: We might want to skip this one unless we have a need, just to avoid multiple versions of `bstr`.

webern: Is it fix-able in the cargo deny configuration? Should I just fix it?

jpculp: You certainly could. I guess it just depends if we want/need anything from 2.0.7 or if we are bumping just to bump. If it's the latter I'd say probably wait.

webern:
push a fix for multiple versions of a couple of sub-crates

https://github.com/awslabs/tough/compare/4fdb00711c9b95e214036ec3a1d38f203998bfbc..620a81691e39e95a666bada4ae6f2961f6a375dc

webern:
^ fix a mistake

https://github.com/awslabs/tough/compare/620a81691e39e95a666bada4ae6f2961f6a375dc..8d16cfea139e2e9f7d1712da71a70618b76ed87f

Cargo.lock
webern: I think it would be better to fix the problem and keep our deps up-to-date, so I'm attempting to fix it.

webern:
^ cargo fmt after fixing a deprecation warning from utc

https://github.com/awslabs/tough/compare/8d16cfea139e2e9f7d1712da71a70618b76ed87f..d4ebe8ec7fd665843a3647954be3a7d10b997bc1

zmrow (approved):
:zebra:

Cargo.lock
jpculp: I guess I'm still skeptical since I think the problem is that we are pulling and compiling two versions without a clear reason to justify it (yet).

webern: > I guess I'm still skeptical since I think the problem is that we are pulling and compiling two versions without a clear reason to justify it (yet).

I've always had a hard time understanding this lint. The language fully supports have two major versions. I don't quite see the benefit of *not* having two major versions, especially since `deny.toml` is calling it out for us to keep track of.

deny.toml
```diff
@@ -58,7 +58,17 @@ license-files = [
 multiple-versions = "deny"
 wildcards = "deny"
 
+skip = [
+    # globset and httptest are using an old version of bstr
+    { name = "bstr", version = "=0.2.17" },
```
ecpullen, line 63: I think this is ok for now given that it is small.

ecpullen (approved):

PR #506, opened 2022-09-12
Get rid of RUSTSEC-2020-0071

flavio:
*Issue #, if available:* none avaialable

*Description of changes:*

The `cargo audit` tool used to report this codebase to be affected by the [RUSTSEC-2020-0071](https://rustsec.org/advisories/RUSTSEC-2020-0071) security issue.

The vulnerability is brought by the `chrono` crate. Luckily, it's possible to disable the vulnerable code by enabling only the `chrono` features that are actually needed.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #561, opened 2023-01-23
support compiling for wasm32 architecture

lulf:
*Description of changes:*

Add feature flags for dependencies to work on wasm target architectures, and exclude the editor module when wasm32 target is set since it relies on filesystem operations not available.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -25,4 +25,5 @@ jobs:
       - run: rustup default 1.66.1
       - run: rustup component add rustfmt
       - run: rustup component add clippy
+      - run: rustup target add wasm32-unknown-unknown
       - run: make ${{ matrix.make_target }}
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,7 @@ build:
 	cargo clippy --locked -- -D warnings
 	cargo build --locked -p olpc-cjson
 	cargo build --locked -p tough
+	cargo build --locked -p tough --target wasm32-unknown-unknown
 	cargo build --locked -p tough-ssm
 	cargo build --locked -p tough-kms
 	cargo build --locked -p tuftool
@@ -34,3 +35,4 @@ integ:
 	set +e
 	cd tough && cargo test --features '' --locked
 	cd tough && cargo test --all-features --locked
+	cd tough && cargo test --no-default-features --locked --target wasm32-unknown-unknown
--- a/tough/Cargo.toml
+++ b/tough/Cargo.toml
@@ -15,7 +15,7 @@ globset = { version = "0.4" }
 hex = "0.4"
 log = "0.4"
 olpc-cjson = { version = "0.1", path = "../olpc-cjson" }
-path-absolutize = "3"
+path-absolutize = { version = "3", features = ["use_unix_paths_on_wasm"] }
 pem = "1"
 percent-encoding = "2"
 reqwest = { version = "0.11", optional = true, default-features = false, features = ["blocking"] }
--- a/tough/src/lib.rs
+++ b/tough/src/lib.rs
@@ -32,6 +32,7 @@
 
 mod cache;
 mod datastore;
+#[cfg(not(target_arch = "wasm32"))]
 pub mod editor;
 pub mod error;
 mod fetch;


webern (approved):
Edit: sorry I meant to click the comment button, not the approve button. Which now means I need to click the "needs changes button" which feels more aggressive than I intended...

webern (changes requested):
Thank you for looking into this. I have a few thoughts/questions that fall under the category of "can we actually support this":

- We would definitely need CI for this, or else I'm concerned that we would promptly break it on WASM. The problem is that the maintainers are not using or building anything on WASM currently.
- I'm also curious about the `editor` module. Is it "very far" from wasm compatibility? What are the filesystem features it requires that are unavailable on WASM?
- How would we do testing on WASM? Would we need flags for those tests that are WASM incompatible?

lulf:
> Thank you for looking into this. I have a few thoughts/questions that fall under the category of "can we actually support this":
> 
> * We would definitely need CI for this, or else I'm concerned that we would promptly break it on WASM. The problem is that the maintainers are not using or building anything on WASM currently.

I've added this to the CI now, thanks for catching that.

> * I'm also curious about the `editor` module. Is it "very far" from wasm compatibility? What are the filesystem features it requires that are unavailable on WASM?

It's primarily the reading from the filesystem, and including some unix symlink dependency. I'm not too familiar with tough itself and I am primarily contributing this in order to allow sigstore-rs to work in the browser, so I'm not sure how much of the editor would be used in practice. I can attempt to make this more fine grained though, since the errors when enabling the editor module doesn't look too scary:

```rust
error[E0433]: failed to resolve: could not find `unix` in `os`
  --> tough/src/editor/signed.rs:26:14
   |
26 | use std::os::unix::fs::symlink;
   |              ^^^^ could not find `unix` in `os`

error[E0599]: no function or associated item named `from_file_path` found for struct `Url` in the current scope
   --> tough/src/editor/signed.rs:710:22
    |
710 |                 Url::from_file_path(&dest)
    |                      ^^^^^^^^^^^^^^ function or associated item not found in `Url`

Some errors have detailed explanations: E0433, E0599.
For more information about an error, try `rustc --explain E0433`.
```

> * How would we do testing on WASM? Would we need flags for those tests that are WASM incompatible?

I've tried to addressed this now as well, basically running tests on the wasm32 target.

lulf:
I've rebased and updated this with the last commit showing how the editor module can still be imported, at the expense of some cfg's in the signed.rs file.

Let me know what you think.

webern:
This is making more sense to me with the CI addition and a real-world use case! Running CI.

lulf:
Sorry, I forgot to check the CI results here, I will fix it to only run wasm32 for the tough crate.

webern:
Since the repository editor is probably not usable without the functions currently not-compiled, I think your first version where the module was conditionally compiled was better.

webern:
Looks like we need to compile guard whatever tests rely on the repository editor.

lulf:
@webern Looks like the httptest in dev-dependencies is pulling in some crates that don't compile for wasm32. Since dev-dependencies can't be optional, I'm not sure how to work around it that easily.

jpculp:
`httptest` hasn't been updated in quite some time. I wonder if there is a different crate we can switch to that is more wasm32-friendly.

lulf:
I walked through modifying the tests and ended up having to take a step back: even if we can compile the tests for wasm32, we can't really execute them without a WASM runtime, so all the existing unit tests cannot really run on the wasm32 target using the standard Rust way.

Instead, I propose excluding existing tests from target_arch = wasm32, and create a new test only executing on wasm32 using wasm-bindgen-test  (https://rustwasm.github.io/wasm-bindgen/wasm-bindgen-test/index.html), and cover some of the functionality there.

webern:
> Instead, I propose excluding existing tests from target_arch = wasm32, and create a new test only executing on wasm32 using wasm-bindgen-test (https://rustwasm.github.io/wasm-bindgen/wasm-bindgen-test/index.html), and cover some of the functionality there.

I have concerns about maintaining separate tests for this architecture. I was hoping that at least a subset of the existing tests could be executed in a WASM runtime. Is that not possible?

jpculp:
Is it problematic that we are disabling the wasmbind feature in `chrono`? We're currently doing so to address RUSTSEC-2020-0071 (https://github.com/awslabs/tough/pull/506).

lulf:
Now that sigstore-rs does not have a required dependency on tough, I don't really have a need for this anymore. Given that the browser side of things doesn't have any filesystem capabilities, I'm not sure if it makes sense with this quick fix, and that it should instead be a more thought out effort to support wasm32, along with wasi. If anyone wants to continue on this path though, feel free to take up the torch!

PR #562, opened 2023-01-23
fix all_targets iterator type

### LINKED ISSUE #560: `Repository::all_targets()` cannot actually be used
https://docs.rs/tough/0.12.5/tough/struct.Repository.html#method.all_targets

https://github.com/awslabs/tough/blob/86a0b6e7349bf25da6881b833eca94408c26e532/tough/src/lib.rs#L413-L416

Because the `Item` associated type isn't specified, all a caller is able to know is that this return type is an Iterator, but not what it contains.

The return type should match [`Targets::targets_iter`](https://docs.rs/tough/latest/tough/schema/struct.Targets.html#method.targets_iter).

webern:
*Issue #, if available:*

Fixes #560

*Description of changes:*

Include the type in an `impl Iterator` return so that the return type is usable.

### DIFF
--- a/tough/src/lib.rs
+++ b/tough/src/lib.rs
@@ -411,7 +411,7 @@ impl Repository {
     }
 
     ///return a vec of all targets including all target files delegated by targets
-    pub fn all_targets(&self) -> impl Iterator + '_ {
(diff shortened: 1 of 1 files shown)


stmcginnis (approved):
Looks reasonable, all checks have passed.

etungsten (approved):

PR #563, opened 2023-01-30
[RFC] add a lifetime parameter to Transport::fetch and Repository::read_target

sunshowers:
Looking for feedback on this one -- is this desirable?

*Issue #, if available:*

*Description of changes:*

Add a lifetime parameter to `Transport::fetch` and `Repository::read_target`.

This is useful in cases where you're implementing a transport that borrows from the reader (e.g. if you're reading a TUF repository from a zip file, returning a [`ZipFile<'a>`](https://docs.rs/zip/latest/zip/read/struct.ZipFile.html) as the transport).

This is a breaking change because `read_target` now borrows from `self`. I don't think that's a major issue, because in most cases you're going to keep the `Repository` open for longer than any of the targets you read from it.

### DIFF
--- a/tough/src/cache.rs
+++ b/tough/src/cache.rs
@@ -260,7 +260,7 @@ impl Repository {
         target: &Target,
         digest: &[u8],
         filename: &str,
-    ) -> Result<impl Read> {
+    ) -> Result<impl Read + '_> {
         fetch_sha256(
             self.transport.as_ref(),
             self.targets_base_url
--- a/tough/src/fetch.rs
+++ b/tough/src/fetch.rs
@@ -8,12 +8,12 @@ use snafu::ResultExt;
 use std::io::Read;
 use url::Url;
 
-pub(crate) fn fetch_max_size(
-    transport: &dyn Transport,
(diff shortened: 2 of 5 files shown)


webern:
Thank you for this PR. On the surface it makes sense, I'll take a closer look soon.

webern (approved):
Makes sense to me (to the extent of my ability to reason about lifetimes) and doesn't seem to affect the call sites. Nice!

@zmrow and @iliana

zmrow (commented):
This seems good!

@sunshowers You mentioned this is a breaking change because of the `read_target` signature change?  That's because of the anonymous lifetime?

sunshowers:
@zmrow Yes -- the `Repository` must now stay open longer than any of the readers for the targets read out of it. This is true in common use but I can definitely see realistic users that may have done something like:

1. Open a `Repository`.
2. Create a bunch of readers for targets, and start the reads.
3. Discard the repository without having completed the reads.

Those users will now have to rearrange their code so that the repository stays open longer than any of the readers.

zmrow (approved):
Thanks!

PR #573, opened 2023-02-10
Remove minor/patch versions from Cargo.tomls

jpculp:
*Description of changes:*

In an effort to keep things flexible, this PR removes unnecessary minor/patch versions from being explicitly set in `Cargo.toml`.

_(Also includes Dependabot's #572, but my change doesn't impact the `Cargo.lock` otherwise)_


webern (approved):
yay

stmcginnis (approved):
🎉

PR #593, opened 2023-03-21
tough: Add 1 second tolerance to `system_time` stepped back check

ecpullen:
*Issue #, if available:*
N/A

*Description of changes:*
In a new system, the system time may not initially be monotonic. When downloading many targets across several threads, a time update can cause the previously latest known time to be in the future (according to the system). As a result adding a 1 second tolerance to the time check should resolve this issue.

### DIFF
--- a/tough/src/lib.rs
+++ b/tough/src/lib.rs
@@ -599,9 +599,9 @@ fn system_time(datastore: &Datastore) -> Result<DateTime<Utc>> {
     let sys_time = Utc::now();
 
     if let Some(Ok(latest_known_time)) = poss_latest_known_time {
-        // Make sure the sampled system time did not go back in time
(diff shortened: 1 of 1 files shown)


jpculp (approved):

stmcginnis (approved):
Seems like a safe thing to do.

etungsten (approved):

webern (changes requested):
don't merge this yet please

stmcginnis:
> don't merge this yet please

Are we still waiting on something here?

webern:
> don't merge this yet please

We talked about it, but I don't remember what we decided. It feels very arbitrary to me. Either we allow the system time to change, or we don't... but 1 second?

Issue #602, opened 2023-04-04
Datastore: Read/write operations may not be thread safe


ecpullen:
The datastore uses a `RwLock` to manage reads and writes; however, the `RwLock___Guard` is dropped when `.path()` is called which will unlock the file.

webern:
This is fixed in #687! Closing!

PR #609, opened 2023-04-21
tough datastore: treat system_time as critical section and document

### LINKED ISSUE #604: Datastore: `system_time` should be a critical section
The current implementation of `Datastore` and `system_time` are open to several race conditions when reading targets from many threads. To ensure that each timestamp written to `latest_known_time.json` is read, the entire process of calculating and writing the latest known time should be completed under a RwLock on the datastore.

webern:
*Issue #, if available:*

Closes #604 
Documents which Datastore functions are thread safe and which one is not, see #602

*Description of changes:*

Adds documentation to the Datastore, moves the system_time function into the Datastore, and adds a lock guard to ensure thread safety of that function.

*Testing*

I tested this with our internal monitoring system for Bottlerocket repos and it worked.

### DIFF
--- a/tough/src/datastore.rs
+++ b/tough/src/datastore.rs
@@ -11,6 +11,7 @@ use std::path::{Path, PathBuf};
 use std::sync::{Arc, PoisonError, RwLock, RwLockReadGuard, RwLockWriteGuard};
 use tempfile::TempDir;
 
+/// `Datastore` persists TUF metadata files.
 #[derive(Debug, Clone)]
 pub(crate) struct Datastore(Arc<RwLock<DatastorePath>>);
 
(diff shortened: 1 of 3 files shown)


stmcginnis (approved):

ecpullen (approved):
LGTM

tough/src/datastore.rs
```diff
@@ -68,6 +89,43 @@ impl Datastore {
             },
         }
     }
+
+    /// Ensures that system time has not stepped backward since it was last sampled. This function
+    /// is protected by a lock guard to ensure thread safety.
+    pub(crate) fn system_time(&self) -> Result<DateTime<Utc>> {
+        // Treat this function as a critical section. This lock is not used for anything else.
+        let _lock = self.time_lock.write().map_err(|e| {
```
cbgbt: I always wondered if non-lexical lifetimes meant that this could potentially be dropped prior to the end of the lexical scope.

After reading a few resources, my understanding is that this will be dropped "at the end of its lifetime" which is currently the end of the scope. Possibly it's worth explicitly `drop(_lock)` at the end of the function to make accidental lifetime changes result in a compiler error.

cbgbt (approved):

cbgbt (approved):

PR #645, opened 2023-08-07
tuftool: Allow providing multiple keys to `root add-key`

### LINKED ISSUE #479: tuftool command to add multiple keys for a role
**Problem**

Currently in `tuftool`, to add multiple keys to a role in the `root.json` file the user has to execute the same command per key they want to add, i. e. with 3 keys regardless of the role, the user has to execute:

```bash
tuftool root add-key <>.root.json <key> -r <role>
tuftool root add-key <>.root.json <key> -r <role>
tuftool root add-key <>.root.json <key> -r <role>
```

**Proposal**

Once #469 is merged, it should be possible to have a command as follows:

```bash
tuftool root add-key <>.root.json \
    -k <key> \
    -k <key> \
    -k <key> \
    -r <role>
```

stmcginnis:
**Issue #, if available:**

Closes: #479 

**Description of changes:**

This command is often used to add multiple keys to a role. That currently means calling the command multiple times, once for each key.

Since this is a common scenario, this changes the subcommand to allow providing multiple keys are part of one invocation.

**Note**: This is a breaking change for the CLI. The singular key used to be provided as a positional argument to the command. This now requires passing it as `-k` or `--key` to match the pattern used in other `tuftool` commands.

**Testing done**

Added unit test to cover case of providing multiple keys.

Created two testkeys in KMS and initialized new root (5.root.json). Verified keys could be added and new public key data matches test keys.

```sh
~/MCM-63706822 - $ tuftool root add-key 5.root.json -k aws-kms://default/alias/bottlerocket-root-2023-08-07 -k aws-kms://default/alias/bottlerocket-root-2023-08-07-1 -r root
420d8813216cf5e6f9d7446591d0fcc0d978f8f1e8fb8a1a61d95dc49e00d6fa
2bf157f0eb68996fa0098eef0059127810d4ab302d631583aae224df81567c5f
~/MCM-63706822 - $ jq . 5.root.json 
{
  "signed": {
    "_type": "root",
    "spec_version": "1.0.0",
    "consistent_snapshot": true,
    "version": 5,
    "expires": "2024-08-03T00:00:00Z",
    "keys": {
      "420d8813216cf5e6f9d7446591d0fcc0d978f8f1e8fb8a1a61d95dc49e00d6fa": {
        "keytype": "rsa",
        "keyval": {
          "public": "-----BEGIN PUBLIC KEY-----\nMIIBojANBgkqhkiG9w0BAQEFAAOCAY8AMIIBigKCAYEA7hhgvC82t3xsE+2+eODB\nS5Y7tR+xXWplJgxN/CMkNP1n2f2C0SSrOHHuinWfXZB5HHewwTQ8Bjr++6VhFSj4\nGrLRZsvpDg3g+MBK4qK2Az+30I1zSu6a/L257BoXfKyRP3vV0Eyfb32yZr+8FupN\n7DSyABeag2AIwGelKW+hSp7vIQsL9yYs9zIoLQRkqljl4oNe/VkaGYf2z39EqA6f\n4866YSmQ+YEMIqBTDFICPJfJLJy/nhqvaI66vK5/7D8JKOGenFOj1/L4FjOM5UqK\nN92xFg6gCq48tka89LfgePbUOyYcs/kT4CrwndsC46hwCZCVTOkKr6QMG8Sf111y\nIP5plh6p4PJj/kzTFriAdIBxonoqcFKUOMG9qKk4n5KD+O0LsfTeRmq+AEW5g8sA\nq3gOusAXobvijAz9CjPb8hVwf2PgOAP+zb22diBLv2LAm6PCbYiEk79ifaz0McEC\nDlxxj+R744rb7MIGLkLkt2EgmuyJaZC9Aw+oClICUDGNAgMBAAE=\n-----END PUBLIC KEY-----\n"
        },
        "scheme": "rsassa-pss-sha256"
      },
      "2bf157f0eb68996fa0098eef0059127810d4ab302d631583aae224df81567c5f": {
        "keytype": "rsa",
        "keyval": {
          "public": "-----BEGIN PUBLIC KEY-----\nMIIBojANBgkqhkiG9w0BAQEFAAOCAY8AMIIBigKCAYEAz8z2CZ3A3RLOffrDjenl\ngBzlMxixaupMSIBqBefEjTOOQ89ecqC4FebWh9X8rY105fhwWuXwaoh5wBu3RFJj\nerduBBQGHen+b0d1AAtdaH4/fP1VQ6tKfPU2MXyLmaA8ubnxXwWuCEgr3H2Jj6FH\npi9en4NH7foo01n84ZPTyPhb68SmuHgsbfD/2eIzSUYGeab8Fo985gWcii+P/yPy\n2vre9buc6ezsmpIRiMMdfwJlRS+fwrleJGgSzf7xilQJX9uimUNnlrH+zJkm8Jmm\n2ltwEjsoH/2mGHlv8Z8LQ5lViz4ztYy1KUy2yeZ/CQPZwzkgg1MwKQ/8AAigMMD8\nOAjE/6ZbFcsVIIV0R+8HOZW9mrvZO7Av7l6giWatNI21/m2V5+MzW/201LpX11OL\nxW+L6R1TSpdNeQhl5p1iPM0QXhXrQnITTR4x505HeVI9qPapi4QN3sPaNgCw1BJ2\n1vC3tYtizeJulJRAJybPJTNQ5BXhjkff4D1WVsonV/sTAgMBAAE=\n-----END PUBLIC KEY-----\n"
        },
        "scheme": "rsassa-pss-sha256"
      }
    },
    "roles": {
      "timestamp": {
        "keyids": [],
        "threshold": 1
      },
      "root": {
        "keyids": [
          "420d8813216cf5e6f9d7446591d0fcc0d978f8f1e8fb8a1a61d95dc49e00d6fa",
          "2bf157f0eb68996fa0098eef0059127810d4ab302d631583aae224df81567c5f"
        ],
        "threshold": 2
      },
      "targets": {
        "keyids": [],
        "threshold": 1
      },
      "snapshot": {
        "keyids": [],
        "threshold": 1
      }
    }
  },
  "signatures": []
}
```

### DIFF
--- a/tuftool/src/root.rs
+++ b/tuftool/src/root.rs
@@ -59,13 +59,13 @@ pub(crate) enum Command {
         /// Version number
         version: NonZeroU64,
     },
-    /// Add a key (public or private) to a role
+    /// Add one or more keys (public or private) to a role
     AddKey {
         /// Path to root.json
         path: PathBuf,
-        /// The new key
-        #[clap(parse(try_from_str = parse_key_source))]
-        key_source: Box<dyn KeySource>,
+        /// The new key to be added
+        #[clap(short = 'k', long = "key", parse(try_from_str = parse_key_source))]
+        key_source: Vec<Box<dyn KeySource>>,
         /// The role to add the key to
         #[clap(short = 'r', long = "role")]
         roles: Vec<RoleType>,
@@ -101,13 +101,13 @@ pub(crate) enum Command {
     Sign {
         /// Path to root.json
         path: PathBuf,
-        ///Key source(s) to sign the file with
-        #[clap(short = 'k', long = "key",parse(try_from_str = parse_key_source))]
+        /// Key source(s) to sign the file with
+        #[clap(short = 'k', long = "key", parse(try_from_str = parse_key_source))]
         key_sources: Vec<Box<dyn KeySource>>,
-        ///Optional - Path of older root.json that contains the key-id
+        /// Optional - Path of older root.json that contains the key-id
         #[clap(short = 'c', long = "cross-sign")]
         cross_sign: Option<PathBuf>,
-        ///Ignore the threshold when signing with fewer keys
+        /// Ignore the threshold when signing with fewer keys
         #[clap(short = 'i', long = "ignore-threshold")]
         ignore_threshold: bool,
     },
@@ -225,15 +225,23 @@ impl Command {
     }
 
     #[allow(clippy::borrowed_box)]
-    fn add_key(path: &Path, roles: &[RoleType], key_source: &Box<dyn KeySource>) -> Result<()> {
+    fn add_key(
+        path: &Path,
+        roles: &[RoleType],
+        key_source: &Vec<Box<dyn KeySource>>,
+    ) -> Result<()> {
         let mut root: Signed<Root> = load_file(path)?;
-        let key_pair = key_source
-            .as_sign()
-            .context(error::KeyPairFromKeySourceSnafu)?
-            .tuf_key();
-        let key_id = hex::encode(add_key(&mut root.signed, roles, key_pair)?);
         clear_sigs(&mut root);
-        println!("{key_id}");
+
+        for ks in key_source {
+            let key_pair = ks
+                .as_sign()
+                .context(error::KeyPairFromKeySourceSnafu)?
+                .tuf_key();
+            let key_id = hex::encode(add_key(&mut root.signed, roles, key_pair)?);
+            println!("Added key: {key_id}");
+        }
+
         write_file(path, &root)
     }
 
--- a/tuftool/README.md
+++ b/tuftool/README.md
@@ -57,9 +57,9 @@ tuftool root set-threshold "${ROOT}" timestamp 1
 tuftool root gen-rsa-key "${ROOT}" "${WRK}/keys/root.pem" --role root
 
 # for this example we will re-use the same key for the other standard roles
-tuftool root add-key "${ROOT}" "${WRK}/keys/root.pem" --role snapshot
-tuftool root add-key "${ROOT}" "${WRK}/keys/root.pem" --role targets
-tuftool root add-key "${ROOT}" "${WRK}/keys/root.pem" --role timestamp
+tuftool root add-key "${ROOT}" -k "${WRK}/keys/root.pem" --role snapshot
+tuftool root add-key "${ROOT}" -k "${WRK}/keys/root.pem" --role targets
+tuftool root add-key "${ROOT}" -k "${WRK}/keys/root.pem" --role timestamp
 
 # sign root.json
 tuftool root sign "${ROOT}" -k "${WRK}/keys/root.pem"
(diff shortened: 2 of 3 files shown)


tuftool/src/root.rs
```diff
@@ -59,13 +59,13 @@ pub(crate) enum Command {
...
+    /// Add one or more keys (public or private) to a role
     AddKey {
         /// Path to root.json
         path: PathBuf,
-        /// The new key
-        #[clap(parse(try_from_str = parse_key_source))]
-        key_source: Box<dyn KeySource>,
+        /// The new key to be added
+        #[clap(short = 'k', long = "key", parse(try_from_str = parse_key_source))]
+        key_source: Vec<Box<dyn KeySource>>,
```
etungsten, line 68: > This is a breaking change for the CLI.

Can you please also update the README examples for creating root.json?

I also wonder if we can do some weird serde shenanigans with an enum/custom type to preserve the old positional arguments. Or maybe parse the arguments outside of clap.

tuftool/src/root.rs
```diff
@@ -225,15 +225,23 @@ impl Command {
...
         clear_sigs(&mut root);
-        println!("{key_id}");
+
+        for ks in key_source {
+            let key_pair = ks
+                .as_sign()
+                .context(error::KeyPairFromKeySourceSnafu)?
+                .tuf_key();
+            let key_id = hex::encode(add_key(&mut root.signed, roles, key_pair)?);
+            println!("{key_id}");
```
etungsten: Maybe a more descriptive log message would be helpful for what this output is.

etungsten (changes requested):

tuftool/src/root.rs
stmcginnis: Good call, I missed that that was in the readme.

We could probably figure out some workaround to keep that backwards compatibility, but I think the `-k` makes this more inline with some of the other calls. Let me know if you feel strongly about that and I will explore it some more, but I'd personally rather go with the new flag-based arguments.

tuftool/src/root.rs
stmcginnis: This actually is just bubbling up the [`KeyIdSnafu`](https://github.com/awslabs/tough/blob/develop/tuftool/src/error.rs#L165) or [`KeyDuplicateSnafu`](https://github.com/awslabs/tough/blob/develop/tuftool/src/error.rs#L159) errors from `add_key`: https://github.com/awslabs/tough/blob/develop/tuftool/src/root.rs#L423-L426

I don't think any more useful context can be added from here, but please do let me know if I'm missing anything.

etungsten: I was referring to the `println!("{key_id}");`. I think it would help contextualize what the hexstring being printed there is.

ecpullen (approved):

tuftool/src/root.rs
stmcginnis: Ah, makes sense. Quick update coming.

etungsten (approved):

PR #646, opened 2023-08-08
tuftool: Allow specifying version in `root init`

stmcginnis:
**Issue #, if available:**

N/A

**Description of changes:**

A new repo is created by calling `tuftool root init <path>`. It is a common pattern when renewing an expiring root to then have to call `tuftool root bump-version` multiple times or `tuftool root set-version`.

Since this is so common, this change makes it possible to provide an optional initial version to `root init` to avoid needing to run multiple commmands. This adds an optional `--version` argument that can take a positive integer to set as the initial root version.

**Testing done:**

- [x] Unit tests added covering success and failure cases.
- [x] Viewed `tuftool root init --help` output and verified correct help information
- [x] Ran normal `tuftool root init test.json` and verified created successfully with the normal initial version of 1
- [x] Ran `tuftool root init --version 5 test.json` and verified created successfully with initial version set to 5

### DIFF
--- a/tuftool/src/root.rs
+++ b/tuftool/src/root.rs
@@ -28,6 +28,9 @@ pub(crate) enum Command {
     Init {
         /// Path to new root.json
         path: PathBuf,
+        /// Initial metadata file version
+        #[clap(long)]
+        version: Option<u64>,
     },
     /// Increment the version
     BumpVersion {
@@ -131,7 +134,7 @@ macro_rules! role_keys {
 impl Command {
     pub(crate) fn run(self) -> Result<()> {
         match self {
-            Command::Init { path } => Command::init(&path),
+            Command::Init { path, version } => Command::init(&path, version),
             Command::BumpVersion { path } => Command::bump_version(&path),
             Command::Expire { path, time } => Command::expire(&path, &time),
             Command::SetThreshold {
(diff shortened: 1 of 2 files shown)


tuftool/src/root.rs
```diff
@@ -28,6 +28,9 @@ pub(crate) enum Command {
     Init {
         /// Path to new root.json
         path: PathBuf,
+        /// Initial metadata file version
+        #[clap(short, long)]
+        version: Option<u64>,
```
etungsten: nit: I feel like `-v` as an flag is more commonly reserved for the verbosity flag. Without context, `-v 5` reads as "set log level to trace". Can we just have `long` and no short flag option?

stmcginnis: Yeah, that's fair. I did kind of wonder about that myself. Even `--version` can be convention for printing out the version of the tool. I was thinking about `--init-version`, `--start-version`, and a few others, but didn't really like any of the options I was coming up with. I think if we just go with the long `--version`, it's at least documented in the help output and should (hopefully) be clear from its context.

stmcginnis: Changed to just have the `long` form. New help output:

```
Create a new root.json metadata file

USAGE:
    tuftool root init <PATH>

ARGS:
    <PATH>    Path to new root.json

OPTIONS:
    -h, --help                 Print help information
        --version <VERSION>    Initial metadata file version
```

etungsten (approved):

ecpullen (approved):

Issue #347, opened 2021-02-16
tuftool: support windows


webern:
There are issue with respect to filepaths and failing tests on Windows.

Cytro54:
So, the problem is that, under windows, tests call the `tuftool` binary with raw windows paths (e.g: `c:\foo\bar`), and not URL paths, IMO , there are two solutions to the problem.

- Make tests call tuftool with `file://` like paths.
- Support raw file paths under *nix and Windows.

They all have advantages and problems : 

- If we support raw file paths, it messes up a little of the code, because the cli supports also "aws-ssm://" paths, and it wouldn't be all URLs anymore ; but it will be easier for users to pass a simple path rather than a "file://" address.
- If we simply fix tests, the code would stay uniform, but it would be unusual for users to pass urls to cli instead of file paths.

PR #656, opened 2023-08-18
Fix up path prefix check for windows

jon-zu:
*Issue #, if available:*

#347 (somewhat, I think)

*Description of changes:*

When loading a repository, the absolutized path will start with \ instead of /, so I changed It to use `std::path::MAIN_SEPARATOR_STR`. This allows me to use the Repository struct on Windows properly.

### DIFF
--- a/tough/src/target_name.rs
+++ b/tough/src/target_name.rs
@@ -137,7 +137,7 @@ fn clean_name(name: &str) -> Result<String> {
                 .context(error::TargetNameComponentsEmptySnafu { name })?
                 .as_os_str();
 
-            // If the first component isn't the main separator ( unix `/`, windows '\' ) 
+            // If the first component isn't the main separator ( unix `/`, windows '\' )
             // then there is a bug or behavior change in absolutize_from.
             ensure!(
                 first_component == std::path::MAIN_SEPARATOR_STR,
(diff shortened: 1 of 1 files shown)


stmcginnis (approved):
I don't have a system to test this on, but this looks reasonable to me. Thanks for addressing this!

tough/src/target_name.rs
```diff
@@ -137,10 +137,10 @@ fn clean_name(name: &str) -> Result<String> {
                 .context(error::TargetNameComponentsEmptySnafu { name })?
                 .as_os_str();
 
-            // If the first component isn't `/` then there is a bug or behavior change in
+            // If the first component isn't the main separator ( unix `/`, windows '\' ) then there is a bug or behavior change in
             // absolutize_from.
```
webern: Nit: we wrap comments at 100-char width.

```suggestion
            // If the first component isn't the main separator ( unix `/`, windows '\' ) then there
            // is a bug or behavior change in absolutize_from.
```

webern (approved):

tough/src/target_name.rs
jon-zu: I've updated that. I thought that would be handled by rustfmt.

cbgbt:
I guess `cargo fmt check` doesn't like the blank space at the end of the line. Happy to update the PR or let you do it!

cbgbt (approved):
Thanks! 🚀

PR #660, opened 2023-08-23
Enable windows and macos builds in CI

### LINKED ISSUE #347: tuftool: support windows

### LINKED ISSUE #349: add windows to continuous integration
Add Windows to the continuous integration build.

Needs these:
- [x] #343 
- [x] #347
- [ ] #348 (maybe)

### LINKED ISSUE #663: Add macos to Continuous Integration
If I extend the CI in https://github.com/awslabs/tough/pull/660 to build on macos, the integration tests fail:

```
 failures:

---- http_integ::test_retries stdout ----
thread 'http_integ::test_retries' panicked at 'Failed to run integration test HTTP servers, is docker running?: Failed to find our configured proxy 'toxictuf'

Caused by:
    I/O error: "error sending request for url (http://127.0.0.1:8472/proxies/toxictuf): error trying to connect: tcp connect error: Connection refused (os error 61)"', tough/tests/http.rs:131:14
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

It seems that the `noxious-server` isn't starting, but we don't capture enough output to determine why.

cbgbt:
*Issue #, if available:*
Closes: #349 
Closes: #347 
Closes: #663

*Description of changes:*

```
    github: run CI on windows
    
    This also disables Rust incremental builds in the CI for a few reasons:
    * Incremental builds need a lot of disk space, which we need to protect
      for Windows builds.
    * Only the first invocation for each cache key is stored, so the delta
      between incremental builds grows larger over time.
    
    See https://github.com/dtolnay/rust-toolchain/issues/26 for more
    details.
```

```
    tough: always use unix-style paths in clean_name
    
    Using `std::path::Path` to manipulate `Url` paths only has the correct
    behavior on unix systems where the properties of the paths are nearly
    identical.
    
    This change moves to using unix-style path manipulation regardless of
    platform.
    
    It also adds a utility for safely creating filesystem paths based on
    `Url` paths for both unix and windows (which is complicated by drive
    lettering in absolute paths).
```

```
    Refrain from using docker for integration testing
    
    Github actions on Windows do not allow running Linux containers. In
    order to run the same CI on Windows as Linux, we compile toxic HTTP
    services at test time.
```

```
    github: run CI on macos
```

```
    integ: use retries when applying toxics
    
    The forked noxious-server takes a little longer to start on MacOS. Using
    retries ensures we keep trying until the server is ready.
```


Makefile
```diff
@@ -28,9 +28,18 @@ build:
...
 
-# checks tough tests with and without the http feature. http testing requires docker.
+
+# installs noxious-server
+# We currently build from a forked version, until such a point that the following are resolved:
+# https://github.com/oguzbilgener/noxious/issues/13
+# https://github.com/oguzbilgener/noxious/pull/14
+.PHONY: noxious
+noxious:
+	cargo install --locked --git https://github.com/cbgbt/noxious.git --tag v1.0.5
```
ecpullen, line 38: Can we add a tracking issue to make sure we don't forget about this fork.

ecpullen (approved):

Makefile
cbgbt: Yes, I opened one here: https://github.com/awslabs/tough/issues/671

bcressey (approved):

cbgbt:
Thanks @chipsenkbeil for help with https://github.com/chipsenkbeil/typed-path/issues/8, which made this much easier than it could have been.

PR #668, opened 2023-08-24
clippy: Fix new lints with 1.72.0

stmcginnis:
*Issue #, if available:*

N/A

*Description of changes:*

The new rust clippy enforces more lints. This addresses warnings that will be flagged with the new 1.72.0 toolchain.

### DIFF
--- a/tough-kms/src/lib.rs
+++ b/tough-kms/src/lib.rs
@@ -71,7 +71,7 @@ impl fmt::Debug for KmsKeySource {
         f.debug_struct("KmsKeySource")
             .field("key_id", &self.key_id)
             .field("profile", &self.profile)
-            .finish()
+            .finish_non_exhaustive()
     }
 }
 
@@ -163,7 +163,7 @@ impl fmt::Debug for KmsRsaKey {
             .field("key_id", &self.key_id)
             .field("signing_algorithm", &self.signing_algorithm)
             .field("public_key", &self.public_key)
-            .finish()
+            .finish_non_exhaustive()
     }
 }
 
--- a/tough/src/editor/targets.rs
+++ b/tough/src/editor/targets.rs
@@ -300,7 +300,7 @@ impl TargetsEditor {
                     delegated_role.keyids.extend(keyids.clone());
                 }
             }
-            for delegated_role in self.new_roles.get_or_insert(Vec::new()).iter_mut() {
+            for delegated_role in &mut *self.new_roles.get_or_insert(Vec::new()) {
                 if delegated_role.name == role {
                     delegated_role.keyids.extend(keyids.clone());
                 }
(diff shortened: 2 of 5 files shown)


tough/src/editor/targets.rs
```diff
@@ -300,7 +300,7 @@ impl TargetsEditor {
                     delegated_role.keyids.extend(keyids.clone());
                 }
             }
-            for delegated_role in self.new_roles.get_or_insert(Vec::new()).iter_mut() {
+            for delegated_role in &mut *self.new_roles.get_or_insert(Vec::new()) {
```
cbgbt, line 303: I'm a little surprised at this one. Do you know what the reasoning behind it is?

cbgbt (approved):

tough/src/editor/targets.rs
ecpullen: Looks like it is implicitly creating an iterator, the issue may be `get_or_insert` instead of `unwrap_or_default`. I'm not sure if that creates a cleaner loop or not.

ecpullen (approved):

tough-kms/src/lib.rs
```diff
@@ -71,7 +71,7 @@ impl fmt::Debug for KmsKeySource {
         f.debug_struct("KmsKeySource")
             .field("key_id", &self.key_id)
             .field("profile", &self.profile)
-            .finish()
+            .finish_non_exhaustive()
```
stmcginnis, line 74: The warning for these is:

```
warning: manual `Debug` impl does not include all fields
```

Using `finish_non_exhaustive()` indicates the entire struct is not intentionally used for the debug representation.

tough/src/editor/targets.rs
stmcginnis: The actual warning from the existing code is:

```
warning: it is more concise to loop over references to containers instead of using explicit iteration methods
```

With this being given as the suggested change.

https://rust-lang.github.io/rust-clippy/master/index.html#/explicit_iter_loop

The `&mut *` part looks a bit confusing at first, but I believe this _should_ be correct of iterating over a mutable reference to each element. Please correct me if I am interpreting that wrong.

tough/src/target_name.rs
```diff
@@ -258,15 +258,15 @@ fn resolved_8() {
 
 #[test]
 fn uncleaned_1() {
-    let name = r#"~/\.\."#;
+    let name = r"~/\.\.";
```
stmcginnis, line 261: These are technically fine. The spec calls for "zero or more #" to demarc raw strings, so the existing code is fine (and probably what is expected in most cases). This was something `cargo clippy --fix` changes though, so I decided to include it to make any future uses of `--fix` have less noise. But it really isn't necessary, so if folks would rather see this kept as-is, I can drop this part of the patch.

cbgbt: I'm ok with this, though I admit I always add the `#` charactrers to raw strings out of habit.

PR #670, opened 2023-08-24
use default provider chains for KMS and SSM

### LINKED ISSUE #669: tough kms doesn't support more complex profiles
The following `~/.aws/profile` doesn't work correctly:
```
[profile signing-ops]
credential_process = gimme-creds
region = us-west-2

[profile signing-root-us-west-2]
source_profile = signing-ops
role_arn = arn:aws:iam::000123456789:role/signing
region = us-west-2
```

The goal is for the `signing-ops` profile to vend creds that are allowed to assume the role specified in `signing-root-us-west-2`, but this always fails with the following error:
```
$ tuftool root add-key 99.root.json -k 'aws-kms://signing-root-us-west-2/alias/my-root-2023-08-23' -r root
failed to load credentials from the credentials cache
```

bcressey:
**Issue #, if available:**
Fixes #669

**Description of changes:**
Wrap the credential provider in a provider chain which adds an outer refreshing cache provider. This is required for certain profiles to work, for example a profile that uses `source_profile` to fetch role credentials from a profile that uses `credential_process`, since the presence of a cache is assumed.

Do the same for the region provider, as the region may come from the environment or IMDS rather than being specified in the profile.

**Testing done:**
With these changes, I was able to use the profile and command in the linked issue to add a key to `root.json`.

**Terms of contribution:**

### DIFF
--- a/tough-kms/src/client.rs
+++ b/tough-kms/src/client.rs
@@ -2,6 +2,8 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
 use crate::error::{self, Result};
+use aws_config::default_provider::credentials::DefaultCredentialsChain;
+use aws_config::default_provider::region::DefaultRegionChain;
 use aws_sdk_kms::Client as KmsClient;
 use snafu::ResultExt;
 use std::thread;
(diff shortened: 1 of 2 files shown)


bcressey:
While I expect both changes are fine and match what users would expect, either could potentially be a breaking change since now region or creds could be taken from the environment or IMDS if the profile doesn't specify them. So I'd suggest bumping the crate versions for `tough-kms` and `tough-ssm`. I can do that here if that's easiest or otherwise it can be handled in the release workflow.

stmcginnis (approved):
Looks good, and I have tested locally to confirm this works.

As to bumping versions, I'm fine with either doing that here, or just making sure we bump prior to next release. I think we normally do raise the versions if there have been changes merged, so seems safe to defer, but it might be more clear to have it already indicated in the code if someone were to clone and install from source.

webern:
> While I expect both changes are fine and match what users would expect, either could potentially be a breaking change since now region or creds could be taken from the environment or IMDS if the profile doesn't specify them. So I'd suggest bumping the crate versions for `tough-kms` and `tough-ssm`. I can do that here if that's easiest or otherwise it can be handled in the release workflow.

I think we normally bump these as part of the release process.

webern (approved):

etungsten (approved):

PR #687, opened 2023-09-28
tough: migrate to async

phu-cinemo:
Depend on tokio for all IO access and convert all related interfaces to async.
Replace `std::io::Read` with `futures_core::stream::Stream` for hopefully seamless migration once `std::async_iter` stabilized. Use async-compatible locks from tokio and coincidentally fix a locking issue in datastore.

This also affects the crates tough-kms, tough-ssm and tuftool.

*Issue #, if available:*
#213
#602 
*Description of changes:*


tough/src/lib.rs
```diff
@@ -114,54 +121,55 @@ impl From<ExpirationEnforcement> for bool {
...
+/// # });
 /// ```
 #[derive(Debug, Clone)]
-pub struct RepositoryLoader<R>
-where
-    R: Read,
-{
-    root: R,
+pub struct RepositoryLoader<'a> {
+    root: &'a [u8],
```
webern, line 172: I'm wondering if this is the right interface.

What I'm thinking about here is whether there should be a `Root` object (maybe one of the existing ones, need to looks at those types) that deserializes itself from a `Transport`.

```
pub struct RepositoryLoader {
    root: DeserializedRoot,
...
```

```rust
impl DeserializedRoot {
    pub async fn from_https(url: Url) -> Result<Self> {}
    pub async fn from_file(path: impl AsRef<Path>) -> Result<Self> {}
    pub async fn from_bytes(path: &[u8]) -> Result<Self> {}
}
```

Basically a convenience to the user. Just a thought...

tough/src/lib.rs
```diff
@@ -541,17 +556,37 @@ impl Repository {
...
+        create_dir_all(filepath_dir)
+            .await
+            .context(error::DirCreateSnafu {
                 path: &filepath_dir,
             })?;
-        std::io::copy(&mut reader, &mut f).context(error::FileWriteSnafu { path: &f.path() })?;
+
+        // Create a new temporary file.
+        let tmp_path = filepath_dir.to_owned();
+        let tmp = tokio::task::spawn_blocking(move || NamedTempFile::new_in(tmp_path))
```
webern, line 572: Why is the thread spawn needed?

tough/src/datastore.rs
```diff
@@ -28,34 +28,26 @@ impl Datastore {
...
+        self.path_lock.write().await
     }
 
-    /// Get a reader to a file in the datastore. Caution, this is *not* thread safe. A lock is
-    /// briefly created on the datastore when the read object is created, but it is released at the
-    /// end of this function.
+    /// Get contents of a file in the datastore. This function is thread safe.
     ///
     /// TODO: [provide a thread safe interface](https://github.com/awslabs/tough/issues/602)
     ///
```
webern, line 46: Looks to me like it's thread safe now! @cbgbt do you agree?

```suggestion
```

tough/src/datastore.rs
```diff
@@ -28,34 +28,26 @@ impl Datastore {
...
-    pub(crate) fn reader(&self, file: &str) -> Result<Option<impl Read>> {
-        let path = self.read().path().join(file);
-        match File::open(&path) {
+    pub(crate) async fn bytes(&self, file: &str) -> Result<Option<Vec<u8>>> {
```
webern, line 47: Stating the obvious: returning the bytes here instead of a reader or stream solves our thread safety problems.

tough/src/http.rs
```diff
@@ -125,93 +132,222 @@ pub struct HttpTransport {
 }
 
 /// Implement the `tough` `Transport` trait for `HttpRetryTransport`
+#[async_trait]
 impl Transport for HttpTransport {
-    /// Send a GET request to the URL. Request will be retried per the `ClientSettings`. The
-    /// returned `RetryRead` will also retry as necessary per the `ClientSettings`.
```
webern, line 130: Seems like this description is still relevant.

tough/src/http.rs
```diff
@@ -125,93 +132,222 @@ pub struct HttpTransport {
...
+        let r = RetryState::new(self.settings.initial_backoff);
+        Ok(fetch_with_retries(r, &self.settings, &url).boxed())
+    }
+}
+
+enum RequestState {
+    /// A response is streaming.
+    Streaming(BoxStream<'static, reqwest::Result<bytes::Bytes>>),
+    /// A request is pending.
+    Executing(BoxFuture<'static, reqwest::Result<reqwest::Response>>),
```
webern: Better match to the description?

```suggestion
    Pending(BoxFuture<'static, reqwest::Result<reqwest::Response>>),
```

webern (commented):
This is an amazing PR. I've gotten through 33 / 54 file :sweat_smile:!

tough/src/lib.rs
phu-cinemo: Because `NamedTempFile::new_in` does synchronous file I/O which might block

tough/src/lib.rs
phu-cinemo: Since the root is supposed to be "shipped with your software", I figured having it in a file or in memory would be the most common case. Having `Read`/`Stream`-like access to a file doesn't make much sense as there are no async interfaces for that in `serde_json`. So it has to be completely in memory anyway.

You can read a file into memory with a single call already, so that would stay the same with your approach in terms of convenience.
If users keep the root in memory already (because it's linked in the binary or because they get it by means we don't support directly), that would require an extra step now.

But maybe `from_https` is a common thing to do?
Also, your proposal adds the slight benefit of an early deserialization error.

phu-cinemo:
updated according to suggestions

tuftool/src/main.rs
```diff
@@ -92,70 +91,104 @@ enum Command {
...
+        move |mut rx| async move { Some((rx.recv().await?, rx)) },
+    )
+    .filter_map(|entry| {
+        let indir = indir.clone();
+        async move {
+            match entry {
+                Ok(entry) => {
+                    if entry.file_type().is_file() {
+                        let future = async move { process_target(entry.path()).await };
+                        Some(tokio::task::spawn(future).await.unwrap())
```
webern: Why do you have an unwrap here? We generally try to avoid panicking and favor errors.
Maybe a failure to spawn is catastrophic, but maybe there's a way to produce a SpawnErrSnafu.

tuftool/src/main.rs
```diff
@@ -92,70 +91,104 @@ enum Command {
...
+                    }
                 }
+                Err(err) => Some(Err(err).context(error::WalkDirSnafu { directory: indir })),
             }
-            Err(err) => Some(Err(err).context(error::WalkDirSnafu { directory: indir })),
-        })
-        .collect()
+        }
+    })
+    .try_collect()
```
webern: Note to myself: double check parallelization.

tough/src/http.rs
```diff
@@ -251,65 +389,29 @@ impl RetryState {
...
 
 /// Much of the complexity in the `fetch_with_retries` function is in deciphering the `Result`
 /// we get from `reqwest::Client::execute`. Using this enum we categorize the states of the
 /// `Result` into the categories that we need to understand.
 enum HttpResult {
-    /// We got a response with an HTTP code that indicates success.
-    Ok(reqwest::blocking::Response),
+    Ok(reqwest::Response),
+    Err(ErrorClass),
+}
```
webern: Could be:

```rust
type HttpResult = std::result::Result<reqwest::Response, ErrorClass>;
```

tough/src/http.rs
```diff
@@ -251,65 +389,29 @@ impl RetryState {
...
+        url: url.clone(),
+        request: RequestState::None,
+        done: false,
+        has_range_support: false,
     }
 }
 
 /// Much of the complexity in the `fetch_with_retries` function is in deciphering the `Result`
 /// we get from `reqwest::Client::execute`. Using this enum we categorize the states of the
 /// `Result` into the categories that we need to understand.
```
webern: This documentation is more applicable to `enum ErrorClass` now.

tough/src/io.rs
```diff
@@ -1,133 +1,154 @@
...
     }
 }
 
-impl<'a> Read for DigestAdapter<'a> {
-    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
-        assert!(
-            self.digest.is_some(),
-            "DigestAdapter::read called after end of file"
-        );
+impl Stream for DigestAdapter {
```
webern, line 31: I trust that your Stream implementations are correct as I have never implemented Stream myself.

tough/src/io.rs
```diff
@@ -1,133 +1,154 @@
...
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
-use crate::error;
+use crate::{error, transport::TransportStream, TransportError};
+use futures::StreamExt;
+use futures_core::Stream;
 use ring::digest::{Context, SHA256};
-use std::io::{self, Read};
+use std::{convert::TryInto, path::Path, task::Poll};
+use tokio::fs::metadata;
```
webern:
```suggestion
use tokio::fs;
```

tough/src/io.rs
```diff
@@ -1,133 +1,154 @@
...
-        Self {
-            reader,
-            specifier,
-            max_size,
-            counter: 0,
-        }
-    }
+/// Async analogue of `std::path::Path::is_file`
+pub async fn is_file(path: impl AsRef<Path>) -> bool {
+    metadata(path).await.map(|m| m.is_file()).unwrap_or(false)
```
webern: nit: slightly clearer that this is a tokio function.

```suggestion
    fs::metadata(path).await.map(|m| m.is_file()).unwrap_or(false)
```

tough/src/io.rs
```diff
@@ -1,133 +1,154 @@
...
-                max_size: self.max_size,
-                specifier: self.specifier,
-            }
-            .fail()?;
-        }
-        Ok(size)
-    }
+/// Async analogue of `std::path::Path::is_dir`
+pub async fn is_dir(path: impl AsRef<Path>) -> bool {
+    metadata(path).await.map(|m| m.is_dir()).unwrap_or(false)
```
webern:
```suggestion
    fs::metadata(path).await.map(|m| m.is_dir()).unwrap_or(false)
```

tuftool/src/main.rs
```diff
@@ -92,70 +91,104 @@ enum Command {
...
-        .par_bridge()
-        .filter_map(|entry| match entry {
-            Ok(entry) => {
-                if entry.file_type().is_file() {
-                    Some(process_target(entry.path()))
-                } else {
-                    None
+    let indir = indir.as_ref().to_owned();
+
+    let (tx, rx) = tokio::sync::mpsc::channel(10);
```
webern, line 152: Is `10` the max parallelization? Should it be `num_cpu`?

```
    let (tx, rx) = tokio::sync::mpsc::channel(10);
```

webern (commented):
Again: amazing work.

I have gotten through all of the files. Now I need to test it in Bottlerocket and pubsys.

tough/src/lib.rs
```diff
@@ -50,24 +50,31 @@ use crate::error::Result;
...
 };
 pub use crate::target_name::TargetName;
+use crate::transport::IntoVec;
 pub use crate::transport::{
     DefaultTransport, FilesystemTransport, Transport, TransportError, TransportErrorKind,
 };
 pub use crate::urlpath::SafeUrlPath;
+use async_recursion::async_recursion;
+pub use async_trait::async_trait;
+use bytes::Bytes;
```
webern: This is in the interface:

```suggestion
pub use bytes::Bytes;
```

tough/src/lib.rs
```diff
@@ -50,24 +50,31 @@ use crate::error::Result;
...
 /// An HTTP transport that includes retries.
 #[cfg(feature = "http")]
-pub use crate::http::{HttpTransport, HttpTransportBuilder, RetryRead};
+pub use crate::http::{HttpTransport, HttpTransportBuilder};
+use crate::io::is_dir;
 use crate::schema::{
     DelegatedRole, Delegations, Role, RoleType, Root, Signed, Snapshot, Timestamp,
 };
 pub use crate::target_name::TargetName;
+use crate::transport::IntoVec;
```
webern: Can we implement some convenience so that `download_target("sometarget").to_bytes()` or `to_vec()` is possible?

webern:
I have started the testing and integration process here https://github.com/bottlerocket-os/twoliter/pull/98

tuftool/src/main.rs
phu-cinemo: we'd only panic if the task happened to panic, that's why I opted for `unwrap` here.
but I'm sure I can turn that into an error, no problem.

tough/src/http.rs
phu-cinemo: The orphan rule would keep us from implementing external traits like `From` for it

tuftool/src/main.rs
phu-cinemo: It's just the buffer depth of the producer. `WalkDir` will run on a special thread reserved for blocking tasks. I chose some value >1 to avoid starvation of the consumers while this thread isn't scheduled.
`tokio` with `rt-multi-thread` enabled will automatically scale the number of threads according to the number of cores. As the consuming futures are spawned as independent tasks, they will evenly distribute across those threads.

Edit: Writing this I realized that I failed to spawn all tasks **before** awaiting them. This is fixed now.

tuftool/src/main.rs
phu-cinemo: good instincts! it wasn't really parallel at all.. see my other comment

tough/src/lib.rs
phu-cinemo: I generalized the implementation of `IntoVec` to deal with any error type and made the trait public. It now can be used on streams returned by the public interface, e.g. `read_target`.
I hope that's what you had in mind!

phu-cinemo:
I incorporated another round of suggestions

matze:
@webern is there anything blocking from further review?

webern:
> @webern is there anything blocking from further review?

I have been out on vacation and will resume the work next week. The blocker is that we want to make sure Bottlerocket can be built, run and update using the new code. I have done about 75% of the work thus far.

webern:
Update. I've run into a problem integrating this downstream in Bottlerocket's update system.

The LZ4 decompression library we use in `updog` takes a blocking `Read` object. I am holding a `Stream` of the target, which in this case is the very large `bottlerocket-image.lz4` file.

I am not seeing an async-compatible (ideally stream compatible) LZ4 library at the moment.

The best I can think of is to write the LZ4 file to disk then open that file with `std::fs` to get a blocking `Read` that I can feed to the lz4 decompression lib. But this is going to be slower than what we have now (feeding the `tough` http transport `Read` straight to decompression). The file is also too large to store it in memory. Another options might be somehow spawning a runtime and blocking on each `next` of the Stream to convert to a `Read`?

@bcressey and @ecpullen :eyes: 

https://github.com/bottlerocket-os/bottlerocket/blob/d74bdb9a3758b2d177921c9a26f8fbd4afedc100/sources/updater/updog/src/main.rs#L229

https://github.com/10XGenomics/lz4-rs/blob/418855f339fe229083f6f8cce77a61d8632ab4c6/src/decoder.rs#L23

webern:
> Update. I've run into a problem integrating this downstream in Bottlerocket's update system.

Hmmm, I wonder if I can create a `Read` object that somehow holds a handle to the tokio runtime and uses `block_on`.

webern:
> > Update. I've run into a problem integrating this downstream in Bottlerocket's update system.
> 
> Hmmm, I wonder if I can create a `Read` object that somehow holds a handle to the tokio runtime and uses `block_on`.

Untested but this compiles

```rust
struct StreamRead<S>
where
    S: Stream<Item = std::result::Result<Bytes, tough::error::Error>> + Send + Unpin,
{
    runtime: tokio::runtime::Handle,
    stream: S,
}

impl<S> StreamRead<S>
where
    S: Stream<Item = std::result::Result<Bytes, tough::error::Error>> + Send + Unpin,
{
    fn new(stream: S) -> Self {
        let runtime = tokio::runtime::Handle::current();
        Self { runtime, stream }
    }
}

impl<S> Read for StreamRead<S>
where
    S: Stream<Item = std::result::Result<Bytes, tough::error::Error>> + Send + Unpin,
{
    fn read(&mut self, buf: &mut [u8]) -> std::io::Result<usize> {
        let maybe_next = self.runtime.block_on(self.stream.next());
        let next = match maybe_next {
            None => return Ok(0),
            Some(next) => next,
        };
        let bytes = next.map_err(|e| std::io::Error::new(ErrorKind::Other, e))?;
        let buffer_length = buf.len();
        let bytes_len = bytes.len();
        if bytes_len > buffer_length {
            // TODO - handle this painful case
            return Err(std::io::Error::new(ErrorKind::Other, "Bug"));
        }
        let bytes = bytes.iter().as_slice();
        // TODO - this seems stupid
        for i in 0..bytes_len {
            buf[i] = bytes[i];
        }
        Ok(bytes_len)
    }
}
```

phu-cinemo:
@webern I think you can find the necessary adapters for this in `tokio_util`.
I didn't try it myself, but you should be able to get from `Stream` to `AsyncRead` using [`tokio_util::io::StreamReader`](https://docs.rs/tokio-util/latest/tokio_util/io/struct.StreamReader.html#). Then, you'd use [`tokio_util::io::SyncIoBridge`](https://docs.rs/tokio-util/0.7.10/tokio_util/io/struct.SyncIoBridge.html#) to get a blocking `Read`.

webern (approved):
I have tested this in Bottlerocket and Bottlerocket's release tooling. Nice work!

@phu-cinemo Can you rebase to resolve the Cargo.toml conflict?

webern:
Here is the corresponding Bottlerocket PR https://github.com/bottlerocket-os/bottlerocket/pull/3566

phu-cinemo:
@webern Thanks! I squashed the fixups and rebased.

zmrow (approved):
Nice work!

tough/Cargo.toml
```diff
@@ -9,21 +9,28 @@ keywords = ["tuf", "update", "repository"]
 edition = "2018"
 
 [dependencies]
+async-recursion = "1.0.5"
+async-trait = "0.1.73"
+bytes = "1.5.0"
 chrono = { version = "0.4", default-features = false, features = ["std", "alloc", "serde", "clock"] }
 dyn-clone = "1"
+futures = "0.3.28"
+futures-core = "0.3.28"
```
webern, line 18: Can you reduce the specificity of the dependency versions? We have seen conflicts with other downstream dependencies at times when they require a "no-higher-than X" version.

```suggestion
async-recursion = "1"
async-trait = "0.1"
bytes = "1"
chrono = { version = "0.4", default-features = false, features = ["std", "alloc", "serde", "clock"] }
dyn-clone = "1"
futures = "0.3"
futures-core = "0.3"
```

tuftool/Cargo.toml
```diff
@@ -42,4 +44,6 @@ walkdir = "2"
 
 [dev-dependencies]
 assert_cmd = "2"
+futures = "0.3.28"
+futures-core = "0.3.28"
```
webern, line 48: Can you reduce the specificity of the dependency versions? We have seen conflicts with other downstream dependencies at times when they require a "no-higher-than X" version.

```suggestion
futures = "0.3"
futures-core = "0.3"
```

tuftool/Cargo.toml
```diff
@@ -21,19 +21,21 @@ aws-sdk-kms = "0.28"
 aws-sdk-ssm = "0.28"
 chrono = { version = "0.4", default-features = false, features = ["alloc", "std", "clock"] }
 clap = { version = "4", features = ["derive"] }
+futures = "0.3.28"
```
webern, line 24: Can you reduce the specificity of the dependency versions? We have seen conflicts with other downstream dependencies at times when they require a "no-higher-than X" version.

```suggestion
futures = "0.3"
```

webern (approved):

tuftool/src/main.rs
```diff
@@ -92,70 +91,112 @@ enum Command {
...
 
-fn write_file<T>(path: &Path, json: &T) -> Result<()>
+async fn write_file<T>(path: &Path, json: &T) -> Result<()>
 where
     T: serde::Serialize,
 {
     // Use `tempfile::NamedTempFile::persist` to perform an atomic file write.
     let parent = path.parent().context(error::PathParentSnafu { path })?;
-    let mut writer =
+    let file =
```
bcressey, line 125: Does this need the same `tokio::task::spawn_blocking` handling since `NamedTempFile` can block?

tough/src/http.rs
```diff
@@ -125,93 +132,224 @@ pub struct HttpTransport {
...
+            RequestState::Pending(_) => f.write_str("Executing"),
+            RequestState::None => f.write_str("None"),
+        }
     }
 }
 
-/// This serves as a `Read`, but carries with it the necessary information to do retries.
 #[derive(Debug)]
-pub struct RetryRead {
+pub(crate) struct RetryStream {
```
bcressey, line 165: Does this need to be `pub(crate)`? Not sure why `RetryRead` was `pub` before.

bcressey (approved):

tuftool/src/main.rs
webern: I can do this after merging. https://github.com/awslabs/tough/issues/697

tough/src/http.rs
webern: I will check this after merging https://github.com/awslabs/tough/issues/698

tough/Cargo.toml
webern: Disregard, I will do this in a fast-follow as I work on publishing this! https://github.com/awslabs/tough/issues/699

webern:
@phu-cinemo thank you for this! Merging!

Issue #700, opened 2023-11-03
release the new async version of tough!


webern:
Do a release to get tough async out into the world!
```[tasklist]
### Tasks
- [ ] https://github.com/awslabs/tough/issues/699
- [ ] https://github.com/awslabs/tough/issues/697
- [ ] https://github.com/awslabs/tough/issues/698
- [x] update deps
- [x] update changelog
- [x] publish
```

PR #706, opened 2023-11-07
Fix publish issue

webern:
*Issue #, if available:*

#700 
Arises from #660, 967e02dfd19c74506111b22f600ef9cb360bc270

*Description of changes:*

```
    remove unused lines from deny.toml
```

```
    deny.toml: allow-wildcard-paths

    failure-server is an internal package used only in testing that we do
    not want to publish to crates.io. When including it in tough's dev-
    dependencies, it is necessary to do so without a version number or else
    cargo publish will fail. The lack of version number causes cargo deny
    to ban failure-server as a wildcard-versioned dependency. The following
    line allows us to refer to our own internal packages without a version
    number.
```

*Testing*

Used `cargo publish --dry-run` to verify this fixes the problem.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -73,8 +73,6 @@ skip = [
     { name = "fastrand", version = "=1.9" },
     # several dependencies are using an old version of bitflags
     { name = "bitflags", version = "=1.3" },
-    # transient dependencies are using an older versions of regex-automata
-    { name = "regex-automata", version = "=0.1" },
     { name = "regex-automata", version = "=0.4" },
     # several dependencies are using an old version of socket2
     { name = "socket2", version = "=0.4" },
--- a/deny.toml
+++ b/deny.toml
@@ -59,6 +59,12 @@ license-files = [
 # Deny multiple versions or wildcard dependencies.
 multiple-versions = "deny"
 wildcards = "deny"
+# failure-server is an internal package used only in testing that we do not want to publish to
+# crates.io. When including it in tough's dev dependencies, it is necessary to do so without a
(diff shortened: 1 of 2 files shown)


deny.toml
```diff
@@ -59,10 +59,14 @@ license-files = [
 # Deny multiple versions or wildcard dependencies.
 multiple-versions = "deny"
 wildcards = "deny"
+# failure-server is an internal package used only in testing that we do not want to publish to
+# crates.io. when including it in tough's dev dependencies, it is necessary to do so without a
```
zmrow: s/when/When

deny.toml
```diff
@@ -59,10 +59,14 @@ license-files = [
...
+# version number or else cargo publish will fail. the lack of a version number causes cargo deny
```
zmrow: s/the/The

deny.toml
```diff
@@ -59,10 +59,14 @@ license-files = [
...
+# to ban failure-server as a wildcard-versioned dependency. the following line allows us to refer
```
zmrow: s/the/The

deny.toml
```diff
@@ -59,10 +59,14 @@ license-files = [
...
+# to our own internal packages without a version number.
+allow-wildcard-paths = true
 
 skip = [
-    # http-test is using an old version of bstr
```
zmrow, line 64: Is this not applicable any more?

webern: I need to add it back, actually. For reasons I don't understand I have to use `make check-licenses` locally instead of using `cargo deny` directly.

zmrow (approved):

etungsten (approved):

Issue #662, opened 2023-08-23
tuftool: Support binary releases on Github


cbgbt:
What I'd Like:
I'd like to be able to install pre-built binaries of tuftool.

```[tasklist]
### Tasks
- [ ] Support binary releases on Linux for x86_64/arm64
- [ ] https://github.com/awslabs/tough/issues/349
- [ ] Support binary releases on Windows for x86_64
- [ ] https://github.com/awslabs/tough/issues/663
- [ ] Support binary releases on MacOS for x86_64/arm64
```

I'm not totally committed to that list of platforms and architectures. I'll keep this issue updated as I continue to do some research.

mikn:
Is something missing to move this along? I am swift approaching the same point that others were in when they asked for this and would love if I could just download the tuftool binary to manage the bottlerocket images rather than building from source. I would prefer not to use the Docker image as binaries integrate better with our dependency management systems.

jku:
> I'd like to be able to install pre-built binaries of tuftool.

I'll mention https://github.com/PyO3/maturin here: it's a pretty cheeky way to achieve not only binary packaging but also a robust delivery mechanism: It creates thin Python wrappers for your rust binaries and uploads them to pypi.org.

https://github.com/woodruffw/zizmor is an example of a rust app that can be installed in seconds with `pip install zizmor`: I believe the only things needed in zizmor sources are one GH workflow and a pyproject.toml file

PR #710, opened 2023-11-08
Add tuftool dockerfile

stockholmux:
*Issue #, if available:*

n/a

*Description of changes:*

This PR adds a very simple Dockerfile for `tuftool`

This solves a problem for Bottlerocket users who have been [challenged to download all the dependencies for tuftool just to get an image](https://github.com/bottlerocket-os/bottlerocket/discussions/2779). Additionally, this potentially avoids the need to create a cross compiled binary (see #662)

### DIFF
--- a/tuftool/Dockerfile
+++ b/tuftool/Dockerfile
@@ -2,6 +2,6 @@
 
 FROM rust:slim
 RUN apt update && apt install -y openssl
-RUN cargo install --force tuftool
+RUN cargo install --force --locked tuftool
 RUN mkdir /share
 ENTRYPOINT ["tuftool"]
--- a/tuftool/Dockerfile
+++ b/tuftool/Dockerfile
@@ -1,0 +1,7 @@
+# syntax=docker/dockerfile:1
+
+FROM rust:slim
+RUN apt update && apt install -y openssl
+RUN cargo install --force tuftool
+RUN mkdir /share
+ENTRYPOINT ["tuftool"]
--- a/tuftool/README.md
+++ b/tuftool/README.md
@@ -135,6 +135,24 @@ tuftool download \
 
 `tuftool` respects the `HTTPS_PROXY` and `NO_PROXY` environment variables.
 
+## Container
+
+You can build a simple container image to avoid needing to install the Rust toolchain and dependencies or your local machine.
+
+To build the image run use Docker or Finch (same argument syntax, just replace `finch` for `docker`):
+
+```shell
+finch build -t tuftool .
+```
+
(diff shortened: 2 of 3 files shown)


tuftool/README.md
```diff
@@ -135,6 +135,24 @@ tuftool download \
...
 `tuftool` respects the `HTTPS_PROXY` and `NO_PROXY` environment variables.
 
+## Container
+
+You can build a simple container image to avoid needing to install the Rust toolchain and dependencies or your local machine.
+
+To build the image run use Docker or Finch (same argument syntax, just replace `finch` for `docker`):
+
+```shell
+finch build -t tuftool .
```
webern: We love finch, and it is great that you say "Docker of Finch" above, but Docker is the more common tool and Finch is only for macOS. We should provide the command in the format that most people will use it.

(The cd makes it more clear where the dockerfile is).

```suggestion
cd finch
docker build -t tuftool .
```

tuftool/README.md
```diff
@@ -135,6 +135,24 @@ tuftool download \
...
+```
+
+To use tuftool, mount the host working directory to `/share`.
+
+For example, to mount the current directory for `download` you would do something like:
+
+```shell
+finch run -it -v $(pwd):/share tuftool download "/share/some_directory" ...
```
webern:
```suggestion
docker run -it -v $(pwd):/share tuftool download "/share/some_directory" ...
```

webern (commented):
This seems like a good idea to me. I suspect you have heard from users that they don't like installing Rust to use `tuftool`? I think we do want to provide binary releases in the future, but this is an easy fix for now.

I need to make sure this works on Linux.

tuftool/README.md
stockholmux: Fair enough, thought I'd try it.

stockholmux:
@webern That's right - users don't want the Rust tool chain in every situation. A binary would be more elegant but it's a lot more work and I'd like to give folks a way to use tuftool ASAP.

webern (approved):

tuftool/Dockerfile
```diff
@@ -0,0 +1,7 @@
+# syntax=docker/dockerfile:1
+
+FROM rust:slim
+RUN apt update && apt install -y openssl
+RUN cargo install --force tuftool
```
cbgbt: This should also use `--locked`

stockholmux: Fixed!

cbgbt (approved):

PR #717, opened 2023-11-15
Update ring to 0.17

### LINKED ISSUE #395: bump untrusted from 0.7.1 to 0.9+
Requires some code changes, fails build e.g. #394

### LINKED ISSUE #711: update to ring 0.17
Looks like this update has some interface changes that will require some manual updates.

_Originally posted by @stmcginnis in https://github.com/awslabs/tough/issues/709#issuecomment-1802029219_

stmcginnis:
*Issue #, if available:*

Closes #711 
Closes #395 

*Description of changes:*

This updates the ring dependency to the 0.17.x releases. This also required updating untrusted to 0.9 due to some public interface usage in ring.

These updates also change some other function signatures, so minor tweaks were needed in calling code.


etungsten (approved):

tough/src/schema/spki.rs
```diff
@@ -67,14 +68,24 @@ pub(super) fn decode(
...
                             alg_ident.read_all(ring::error::Unspecified, |input| {
-                                if der::expect_tag_and_get_value(input, der::Tag::OID)?
-                                    != untrusted::Input::from(&asn1_encode_oid(algorithm_oid))
+                                let expected_tag_value =
+                                    der::expect_tag_and_get_value(input, der::Tag::OID)?;
+
+                                let asn1_encode = asn1_encode_oid(algorithm_oid);
+                                let algo_encode_oid: Input<'_> =
+                                    untrusted::Input::from(&asn1_encode);
+                                if expected_tag_value.as_slice_less_safe()
```
cbgbt: For posterity, I was concerned about using functions with `_less_safe` in the name. It seems that the `untrusted::Input` type tends to push you towards using a `Reader` and explicitly doing a byte-for-byte comparison as an acknowledgment that the comparison uses a boundless amount of compute. It also prevents you from parsing the input more than once.

It's easy for us in this context to see that we aren't parsing multiple times, we're just doing a bit comparison. For the problem of unbounded inputs, I suppose here we're always bounded by the size of `algorithm_oid`.

It might be good to add commentary about why this is safe, it makes me a bit squirrely to see code that implies that it could panic.

stmcginnis: Yeah, that's fair. I'll add a comment since I do think being "less safe" here is actually safe for us.

cbgbt (approved):

PR #729, opened 2023-11-29
tuftool: fix root gen-rsa command bug

### LINKED ISSUE #728: tuftool: bug when using path as URL
A bug was introduced in this commit https://github.com/awslabs/tough/commit/55a40cc

These lines do not work for `tuftool gen rsa-key` because the path to the key is for a file that does not yet exist. So we end up parsing the filepath with Url which fails like this:

```rust
    let input_as_path = Path::new(input);
    let url = if input_as_path.exists() {
        Url::from_file_path(input)
            .ok() // dump unhelpful `()` error
            .context(error::FileUrlSnafu {
                path: input_as_path.to_owned(),
            })?
    } else {
        Url::parse(input).context(error::UrlParseSnafu { url: input })?
    };
```

```shell
$ dir=/tmp/url-bug
rm -rf $dir
mkdir $dir

tuftool root init $dir/root.json
tuftool root gen-rsa-key \
   $dir/root.json \
   $dir/default.pem \
   --role root --role snapshot --role targets --role timestamp
```

Output:

```
Failed to parse URL "/tmp/url-bug/default.pem": relative URL without a base
```

webern:
*Issue #, if available:*

Closes #728 

*Description of changes:*

A bug introduced in 55a40cc caused the tuftool tuftool root gen-rsa command to be unusable because it caused file paths to be parsed with Url::parse. This commit only uses URL for parsing if the SSM or KMS schemes are found. Otherwise it is assumed to be a file path and parsed with PathBuf.

*Testing*

This now works (previously would error):

```
dir=/tmp/url-bug
rm -rf $dir
mkdir $dir

cargo run --package tuftool -- root init $dir/root.json
cargo run --package tuftool -- root gen-rsa-key \
   $dir/root.json \
   $dir/default.pem \
   --role root --role snapshot --role targets --role timestamp

dir=/tmp/url-bug
rm -rf $dir
mkdir $dir
```

This also works (using `aws-kms`):

```
dir=/tmp/url-bug
rm -rf $dir
mkdir $dir

cargo run --package tuftool -- root init $dir/root.json
cargo run --package tuftool -- root add-key $dir/root.json \
   --key aws-kms:///alias/bottlerocket \
   --role root

```

### DIFF
--- a/tuftool/src/source.rs
+++ b/tuftool/src/source.rs
@@ -34,10 +34,9 @@
 //! "aws-ssm:///a/key" (notice the 3 slashes after the colon)
 
 use crate::error::{self, Result};
-use snafu::{OptionExt, ResultExt};
-use std::path::Path;
+use snafu::ResultExt;
+use std::path::PathBuf;
 use tough::key_source::{KeySource, LocalKeySource};
-use tough::SafeUrlPath;
 use tough_kms::{KmsKeySource, KmsSigningAlgorithm};
 use tough_ssm::SsmKeySource;
 use url::Url;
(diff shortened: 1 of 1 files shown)


etungsten (approved):

tuftool/src/source.rs
```diff
@@ -51,60 +50,173 @@ use url::Url;
...
+fn test_parse_path_or_url_path_5() {
+    let input = "C:";
+    let expected = PathOrUrl::Path(PathBuf::from(input));
+    let actual = parse_path_or_url(input).unwrap();
+    assert_eq!(expected, actual);
+}
+
+#[test]
+fn test_parse_path_or_url_path_6() {
+    let input = r#"C:\foo.txt"#;
```
cbgbt, line 170: Nice, thanks for testing this

cbgbt (approved):
Nice catch! I think avoiding the machinery that is URL parsing in this case makes a lot of sense and cleans this up quite nicely.

PR #737, opened 2023-12-29
tough: fix ecdsa key reading

Xynnn007:
In the ecdsa public key, there are two OIDs to specify the key type and the curve. In the decode logic OID_EC_PUBLIC_KEY and OID_EC_PARAM_SECP256R1 are specified.

However, in `ring::io::der` every read would only read one OID. Current code does not read the second OID and use the first OID to compared the given OID_EC_PARAM_SECP256R1 one. Thus all legal pem ecdsa keys would be deserialized unsuccessfully.

### DIFF
--- a/tough/src/schema/de.rs
+++ b/tough/src/schema/de.rs
@@ -97,4 +97,15 @@ mod tests {
         ))
         .is_ok());
     }
+
+    /// Ensure that we can deserialize a root.json file that has pem-encoded ECDSA keys. This uses
+    /// sigstore's root.json file taken from here:
+    /// `<https://github.com/sigstore/sigstore-rs/blob/8a269a3/trust_root/prod/root.json>`
+    #[test]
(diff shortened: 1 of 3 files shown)


cbgbt:
Thanks for the PR! I'll have a look, though I have some learning to do to give a useful review.

bcressey (approved):

cbgbt (approved):

Xynnn007:
Hi @cbgbt @bcressey Thanks for reviewing/merging this. When will the next crate release be published including this patch?

Xynnn007:
Kindly ping again @cbgbt @bcressey . When will the next crate release be published including this patch?

webern:
> Kindly ping again @cbgbt @bcressey . When will the next crate release be published including this patch?

This is released, sorry for the delay. Unfortunately #755 isn't in this release so we will need to do another release as soon as we can.

Issue #754, opened 2024-03-13
support "ecdsa" keytype as well as the old "ecdsa-sha2-nistp256"


jku:
The TUF spec has last year fixed the inconsistency in keytypes so "ecdsa" is now the suggested keytype name (instead of "ecdsa-sha2-nistp256"):
https://theupdateframework.github.io/specification/latest/#keytype

For clients it's fine (and the suggested approach) to support both names.

Based on my (limited) understanding of serde this might be an easy fix: 

    /// An EcdsaKey
    #[serde(rename = "ecdsa-sha2-nistp256")]
    Ecdsa {
    
becomes something like

    /// An EcdsaKey
    #[serde(alias = "ecdsa-sha2-nistp256")]
    Ecdsa {

?

jku:
This came up since sigstore root-signing is trying to use "ecdsa" as a keytype: https://github.com/sigstore/root-signing/pull/1164. The actual error from tuftool is:
`Failed to load repository: Failed to parse root metadata: Invalid SubjectPublicKeyInfo document at line 18 column 4`
https://github.com/sigstore/root-signing/actions/runs/8262338514/job/22601461047

So it is still possible that the issue is something not related to the ecdsa change itself but that seemed like the most obvious culprit (as no public key content has actually changed).

Also apologies for how the whole ecdsa keytype evolution has happened: I realize this is not the first or second time things have had to change in tough around this. We (TUF community as whole) could really improve the communication on these issues...

cc @kommendorkapten

kommendorkapten:
👋 
Will there also be a new release of `tuftool` with this update? We are using `tuftool` in Sigstore's TUF repo to verify client compatibility.

jmt-lab:
> 👋 Will there also be a new release of `tuftool` with this update? We are using `tuftool` in Sigstore's TUF repo to verify client compatibility.

Hi there, we haven't planned a new release of tuftool due to there only being changes in tough itself and we did a minor release. If you use cargo install to install tuftool or run a cargo update it should bring in the new version of tough. If it does not or you see any weirdness do please reach out and let us know.

kommendorkapten:
@jmt-lab Perfect, thanks. Verified this too and it works, which can be seen in this PR: https://github.com/sigstore/root-signing/pull/1190

jku:
Seems to be fixed, thanks a lot.

PR #755, opened 2024-03-13
Added a new ecdsa key that uses the new type `ecdsa`.

kommendorkapten:
In the TUF spec v1.0.32 (released 2023-03-02) the key type was updated for ecdsa keys from `ecdsa-sha2-nistp256`.
This PR introduces the new type and keeps support for the older key type.

*Issue #, if available:* https://github.com/awslabs/tough/issues/754

*Description of changes:*

In the TUF spec v1.0.32 (released 2023-03-02) the key type was updated for ecdsa keys from `ecdsa-sha2-nistp256`.
This PR introduces the new type and keeps support for the older key type.

I first tried with `  #[serde(alias = "ecdsa-sha2-nistp256")]` as proposed by @jku but I never got it to work.

### DIFF
--- a/tough/src/schema/de.rs
+++ b/tough/src/schema/de.rs
@@ -108,4 +108,14 @@ mod tests {
         ))
         .is_ok());
     }
+    /// Ensure that we can deserialize a root.json file that has ECDSA keys with new type ecdsa. This uses
+    /// sigstore's root.json file taken from here:
+    /// `<https://github.com/sigstore/root-signing/blob/d3738d62e92580b5b928d6212c927084ada2bfee/repository/repository/9.root.json>`
+    #[test]
+    fn ecdsa_new_type_keys() {
+        assert!(serde_json::from_str::<Signed<Root>>(include_str!(
+            "../../tests/data/ecdsa-new-type-sig-keys/root.json"
+        ))
+        .is_ok());
+    }
 }
--- a/tough/src/schema/key.rs
+++ b/tough/src/schema/key.rs
@@ -57,9 +57,19 @@ pub enum Key {
         #[serde(flatten)]
         _extra: HashMap<String, Value>,
     },
-    /// An EcdsaKey
-    #[serde(rename = "ecdsa-sha2-nistp256")]
+    /// An EcdsaKey.
     Ecdsa {
+        /// The Ecdsa key.
+        keyval: EcdsaKey,
+        /// Denotes the key's signature scheme.
+        scheme: EcdsaScheme,
+        /// Any additional fields read during deserialization; will not be used.
+        #[serde(flatten)]
+        _extra: HashMap<String, Value>,
+    },
+    /// An EcdsaKey with the old key type.
+    #[serde(rename = "ecdsa-sha2-nistp256")]
+    EcdsaOld {
         /// The Ecdsa key.
         keyval: EcdsaKey,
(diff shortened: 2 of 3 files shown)


Hayden-IO:
From the Sigstore community, thanks so much for this PR. We use awslabs/tough in our Rust client, so getting this merged in will help unblock Sigstore users.

jku:
> I first tried with   #[serde(alias = "ecdsa-sha2-nistp256")] as proposed by @jku but I never got it to work.

:+1: 

I think the code uses the Key struct when canonicalizing the signed content: my suggestion would then be wrong (as you must have the original keytype in the canonicalized content). 

I'm sure there is a way to store the original keytype and use that when serialising even if using a single struct but I'm not sure it would be simpler than your suggestion

kommendorkapten:
Some local testing with the soon to be deployed [TUF root](https://github.com/sigstore/root-signing/pull/1164) for sigstore

```shell
kommendorkapten@m1m14:~/git/tough % ./target/debug/tuftool download out --root ~/git/root-signing/repository/repository/5.root.json -t http://localhost:8081/targets -m http://localhost:8081
Downloading targets to "out"
	-> fulcio_v1.crt.pem
	-> trusted_root.json
	-> fulcio_intermediate_v1.crt.pem
	-> artifact.pub
	-> rekor.pub
	-> ctfe.pub
	-> ctfe_2022.pub
	-> fulcio.crt.pem
kommendorkapten@m1m14:~/git/tough % git status
On branch new-ecdsa-key-type
```

Switching to default branch:
```
kommendorkapten@m1m14:~/git/tough % ./target/debug/tuftool download out --root ~/git/root-signing/repository/repository/5.root.json -t http://localhost:8081/targets -m http://localhost:8081
Failed to load repository: Failed to parse root metadata: unknown variant `ecdsa`, expected one of `rsa`, `ed25519`, `ecdsa-sha2-nistp256` at line 9 column 22
kommendorkapten@m1m14:~/git/tough % git status
On branch develop
```

Hayden-IO:
@webern Would you be able to take a look at this? Thanks!

webern:
I don't quite understand. Is the issue that `"ecdsa"` is now a valid `"keytype"` when it used to require the `"keytype"` to be "ecdsa-sha2-nistp256"?

Serde does have `alias` for cases like this, but in that case we would not know which way to serialize so maybe this is the best approach. Can you confirm my understanding of the issue?

Thanks.

jku:
> I don't quite understand. Is the issue that `"ecdsa"` is now a valid `"keytype"` when it used to require the `"keytype"` to be "ecdsa-sha2-nistp256"?

Yes. client should should accept both for backwards compat but "ecdsa" is considered correct.

> Serde does have `alias` for cases like this, but in that case we would not know which way to serialize so maybe this is the best approach. Can you confirm my understanding of the issue?

Agreed: alias would work except if the serialized form is used for anything (because the canonicalized form would then be different and so signatures wouldn't match).

astoycos:
Another data point here we're hitting this in the https://github.com/bpfman/bpfman via https://github.com/sigstore/sigstore-rs/issues/338 and it's currently a build blocker for us at `:main`

webern (approved):

bcressey (approved):

Hayden-IO:
Thanks all!

Issue #771, opened 2024-06-05
client: hashes and length should be optional in (timestamp, snapshot) METAFILEs


jku:
I was sure this issue already existed but now I cannot find it...

tough client does not seem to support METAFILEs without hashes or length within timestamp and snapshot metadata.

The specification is clear in this case: these are optional fields that can be omitted:
> **LENGTH**
> An integer length in bytes of the metadata file at [METAPATH](https://theupdateframework.github.io/specification/latest/#snapshot-metapath). It is OPTIONAL and can be omitted to reduce the snapshot metadata file size. In that case the client MUST use a custom download limit for the listed metadata.
> 
> **HASHES**
> A dictionary that specifies one or more hashes of the metadata file at [METAPATH](https://theupdateframework.github.io/specification/latest/#snapshot-metapath), with the cryptographic hash function as key and the value as HASH, the hexdigest of the cryptographic function computed on the metadata file at [METAPATH](https://theupdateframework.github.io/specification/latest/#snapshot-metapath). For example: { "sha256": HASH, ... }. [HASHES](https://theupdateframework.github.io/specification/latest/#metapath-hashes) is OPTIONAL and can be omitted to reduce the snapshot metadata file size. In that case the repository MUST guarantee that [VERSION](https://theupdateframework.github.io/specification/latest/#metapath-version) alone unambiguously identifies the metadata at [METAPATH](https://theupdateframework.github.io/specification/latest/#snapshot-metapath).

I am currently not including these fields in what is likely to be sigstore TUF repository in future. This lead to https://github.com/sigstore/sigstore-rs/issues/369

flavio:
I'm willing to look into that and eventually build a patch for that. I tried to reproduce the issue you mentioned using the sigstore-staging TUF repository, but this is currently malformed (as you pointed out inside of the sigstore-rs issue).

[...]

PR #769, opened 2024-05-22
chore(deps): update reqwest

flavio:
_Issue #, if available:_

The current version of tough depends on an older version of reqwest which in cause a old version of hyper to be consumed.

This hyper upgrade is massive, having a unified version of hyper inside of projects consuming the tough crate is beneficial. For example, hyper testing frameworks can work only with one version of the library at a time.

_Description of changes:_

Update to latest version of reqwest, v0.12. This version depends on hyper v1, while previous versions (like v0.11) rely on hyper v0.14.

PR #778, opened 2024-07-01
fix: timestamp.json meta can has optional fields

flavio:
_Issue #, if available:_

Fixes issue https://github.com/awslabs/tough/issues/771

_Description of changes:_

According to the TUF specification, the `meta` attribute of
`timestamp.json` must follow the same specification of `METAFILES`.
That means it has optional `LENGTH` and `HASHES`.

See [this](https://theupdateframework.github.io/specification/latest/#file-formats-timestamp) section of the TUF specification.

I've handled a missing `LENGTH` and `HASHES` in the `timestamp.json` file using the same logic used by the library when loading targets.

### DIFF
--- a/tough/src/cache.rs
+++ b/tough/src/cache.rs
@@ -96,7 +96,8 @@ impl Repository {
     {
         self.cache_file_from_transport(
             self.snapshot_filename().as_str(),
-            self.max_snapshot_size()?,
+            self.max_snapshot_size()?
+                .unwrap_or(self.limits.max_snapshot_size),
             "timestamp.json",
             &metadata_outdir,
         )
@@ -237,7 +238,7 @@ impl Repository {
     }
 
     /// Gets the max size of the snapshot.json file as specified by the timestamp file.
-    fn max_snapshot_size(&self) -> Result<u64> {
+    fn max_snapshot_size(&self) -> Result<Option<u64>> {
         let snapshot_meta =
             self.timestamp()
                 .signed
--- a/tough/src/lib.rs
+++ b/tough/src/lib.rs
@@ -250,8 +250,8 @@ impl<'a> RepositoryLoader<'a> {
 /// are set higher than what would reasonably be expected by a repository, but not so high that the
 /// amount of data could interfere with the system.
 ///
-/// `max_root_size` and `max_timestamp_size` are the maximum size for the `root.json` and
-/// `timestamp.json` files, respectively, downloaded from the repository. These must be
+/// `max_root_size`, `max_timestamp_size` and `max_snapshot_size` are the maximum size for the `root.json`,
+/// `timestamp.json` and `snapshot.json` files, respectively, downloaded from the repository. These must be
 /// sufficiently large such that future updates to your repository's key management strategy
 /// will still be supported, but sufficiently small such that you are protected against an
 /// endless data attack (defined by TUF as an attacker responding to clients with extremely
@@ -261,6 +261,7 @@ impl<'a> RepositoryLoader<'a> {
 /// * `max_root_size`: 1 MiB
 /// * `max_targets_size`: 10 MiB
 /// * `max_timestamp_size`: 1 MiB
+/// * `max_snapshot_size`: 1 MiB
 /// * `max_root_updates`: 1024
 #[derive(Debug, Clone, Copy)]
 pub struct Limits {
@@ -275,6 +276,9 @@ pub struct Limits {
     /// The maximum allowable size in bytes for the downloaded timestamp.json file.
     pub max_timestamp_size: u64,
 
+    /// The maximum allowable size in bytes for the downloaded snapshot.json file.
+    pub max_snapshot_size: u64,
+
     /// The maximum number of updates to root.json to download.
     pub max_root_updates: u64,
 }
@@ -285,6 +289,7 @@ impl Default for Limits {
             max_root_size: 1024 * 1024,         // 1 MiB
             max_targets_size: 1024 * 1024 * 10, // 10 MiB
             max_timestamp_size: 1024 * 1024,    // 1 MiB
+            max_snapshot_size: 1024 * 1024,     // 1 MiB
             max_root_updates: 1024,
         }
     }
@@ -360,6 +365,7 @@ impl Repository {
             transport.as_ref(),
             &root,
             &timestamp,
+            limits.max_snapshot_size,
             &datastore,
             &metadata_base_url,
             expiration_enforcement,
@@ -906,10 +912,12 @@ async fn load_timestamp(
 }
 
 /// Step 3 of the client application, which loads the snapshot metadata file.
+#[allow(clippy::too_many_lines)]
 async fn load_snapshot(
     transport: &dyn Transport,
     root: &Signed<Root>,
     timestamp: &Signed<Timestamp>,
+    max_snapshot_size: u64,
     datastore: &Datastore,
     metadata_base_url: &Url,
     expiration_enforcement: ExpirationEnforcement,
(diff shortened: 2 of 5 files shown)


flavio:
JFYI: TUF's reference implementation (written in python) has `LENGTH` and `HASHES` optional also for `timestamp.json`

jku (approved):
Seems correct to me.

The only note I have is that I believe TimestampMeta and SnapshotMeta are now 100% equivalent (as they should be): Having two separate types seems reasonable  but maybe there should be only one struct implementation and then TypeAliases for the actual types?

flavio:
I agree with you @jku, I just wanted to go soft on the initial PR to not scare the maintainers :smile: 

But this change is definitely on the table if the maintainers are fine with that.

jpculp:
This change makes sense to me, although it has triggered a clippy warning. Unless someone disagrees, I think it might be fine to just add a clippy exception to `load_snapshot` for now and we can break up the function in a later PR.

tough/src/cache.rs
```diff
@@ -96,7 +96,8 @@ impl Repository {
     {
         self.cache_file_from_transport(
             self.snapshot_filename().as_str(),
-            self.max_snapshot_size()?,
+            self.max_snapshot_size()?
+                .unwrap_or(self.limits.max_timestamp_size),
```
bcressey: I would prefer adding a separate `max_snapshot_size` limit to the limits struct. It can share the default with `max_timestamp_size` (1 MiB) but it's a different metadata file and clients may want to treat it differently.

tough/src/schema/mod.rs
```diff
@@ -1128,11 +1128,18 @@ pub struct Timestamp {
 /// file, this MUST only include a description of the snapshot.json file.
 #[derive(Debug, Clone, Deserialize, Serialize, Eq, PartialEq)]
 pub struct TimestampMeta {
-    /// The integer length in bytes of the snapshot.json file.
-    pub length: u64,
+    /// The integer length in bytes of the snapshot.json file. It is OPTIONAL and
+    /// can be omitted to reduce the snapshot metadata file size. In that case the client MUST use a
+    /// custom download limit for the listed metadata.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub length: Option<u64>,
```
bcressey: After this change, `TimestampMeta` looks identical to `SnapshotMeta`. If we're taking an API break for this, seems like we should just converge on a single `MetaFile` struct, to better convey the idea that this is actually the same type according to the spec.

bcressey (changes requested):

tough/src/cache.rs
flavio: Yes, good point. I've fixed that with 44740517cbe5a74421b818682f8bb4fff637aa50

I will squash the commit once the review process is done

tough/src/schema/mod.rs
flavio: I agree, do you want me to attempt to perform this change now or would you prefer a dedicated PR later?

jpculp: I think this should probably be in the same PR. Perhaps a separate commit for now and then we squash at the end of the review process?

flavio: I've just pushed a commit that replaces `SnapshotMeta` and `TimestampMeta` with `Metafiles`.

I'm open to suggestions for a better name :sweat_smile:

flavio:
The CI is failing because clippy complains about the [`load_snapshot`](https://github.com/flavio/tough/blob/eb5e25e1609ac30d6c9c5f694ccc608acc35401c/tough/src/lib.rs#L909) function having too many lines.

Should I add a directive to ignore this warning? The function seems ok to me...

jpculp:
> The CI is failing because clippy complains about the [load_snapshot](https://github.com/flavio/tough/blob/eb5e25e1609ac30d6c9c5f694ccc608acc35401c/tough/src/lib.rs#L909) function having too many lines.

I would probably just add a `clippy` exception for now and we can make an issue to split up the function in a separate PR.

flavio:
> I would probably just add a `clippy` exception for now and we can make an issue to split up the function in a separate PR.

Done. BTW, the function length limit is 100, the function is currently 101 lines long :sweat:

jpculp (approved):

tough/src/schema/mod.rs
```diff
@@ -292,7 +292,7 @@ pub struct Snapshot {
 ///    },
 /// ```
 #[derive(Debug, Clone, Deserialize, Serialize, Eq, PartialEq)]
-pub struct SnapshotMeta {
+pub struct Metafiles {
```
bcressey: I think this should be `Metafile` since there's only one. I was thrown at first by the `Hashes` struct immediately after this one, but that's different because it can include multiple hashes, even though only the sha256 one is used.

In the enclosing structs, the `HashMap<String, Metafile>` collection implies the "plural" - there can be 0-N metafiles - so the type name doesn't need to reflect it. If you really wanted a closer correspondence to the spec, you could declare a newtype:
```
struct Metafiles(HashMap<String, Metafile>)
```

And then `Snapshot` and `Timestamp` could say this instead:
```
    pub meta: Metafiles,
```

But I don't think that's necessary.

tough/src/lib.rs
```diff
@@ -360,6 +365,7 @@ impl Repository {
             transport.as_ref(),
             &root,
             &timestamp,
+            limits.max_timestamp_size,
```
bcressey:
```suggestion
            limits.max_snapshot_size,
```

tough/src/lib.rs
```diff
@@ -906,10 +912,12 @@ async fn load_timestamp(
 }
 
 /// Step 3 of the client application, which loads the snapshot metadata file.
+#[allow(clippy::too_many_lines)]
 async fn load_snapshot(
     transport: &dyn Transport,
     root: &Signed<Root>,
     timestamp: &Signed<Timestamp>,
+    max_timestamp_size: u64,
```
bcressey:
```suggestion
    max_snapshot_size: u64,
```

bcressey (commented):
Looks good, just a couple minor quibbles - including the inevitable naming discussion 😅

tough/src/schema/mod.rs
flavio: @bcressey: sorry, should I go ahead and rename `Metafiles` to `Metafile`?

If you want I can also go ahead with the `struct Metafiles` thing you suggested.

bcressey: > sorry, should I go ahead and rename Metafiles to Metafile?

@flavio: yes please rename `Metafiles` to `Metafile`. I wouldn't bother with the second part (newtype) unless you're excited about it 😀

flavio: I’m currently on vacation. I’ll look at that on the 1st of August

flavio: I'm back from holidays. I've done the rename inside of 0274aa528f7f3101ebcf5c7b69971e9522f20be3

I tried to introduce the `struct Metafiles`, but the change required to implement `get` and `insert` methods on it. I don't know if we really want all this extra code.

I'll leave you to mark this comment as solved if you're fine with the change

flavio:
Just a reminder, once everybody is happy about the change I'll squash all commits into a single one.

bcressey (approved):

bcressey:
> Just a reminder, once everybody is happy about the change I'll squash all commits into a single one.

Looks good to me! Thanks!

flavio:
Great, I've squashed all the commits into a single one

jpculp (approved):

arnaldo2792 (approved):

flavio:
Hello, do you have any ETA about when this fix will be made available through an official release?

Thanks!

loosebazooka:
Hi @jpculp. Sorry, just pinging this after the long weekend, hoping it doesn't get lost.  I understand that sigstore work isn't really your concern but, it would be helpful for us to get a release on this, as we're planning some updates to our TUF metadata that break without it. If there's an ETA, I'd love to try to schedule our work around it.

anfredette:
This change broke [bpfman](https://github.com/bpfman/bpfman) today.  Is there a schedule for a release containing this fix?  Thanks, Andre

jpculp:
@flavio / @loosebazooka, sorry for the delay! We were taking a look at some other contributions first, but I suspect they will end up in a separate release. I can't commit to a date yet, but I'd like to get this out sooner rather than later. Thank you for your patience!

@anfredette, can you elaborate on the breakage with `bpfman`?

anfredette:
Thanks for your response, @jpculp!  We just started getting the following error from `sigstore::trust::sigstore::SigstoreTrustRoot::new()`:
```
Failed to parse timestamp metadata: missing field `length` at line 14 column 4
```

loosebazooka:
Oh my bad I guess they already moved the sigstore metadata.

flavio:
On top of having this fix published, we also need https://github.com/awslabs/tough/pull/769 to be merged.

Currently `sigstore` on the `main` branch is affected by [this](https://github.com/awslabs/tough/pull/769#issuecomment-2220898278) issue.
The only way to workaround it is to use a patched version of tough, which is something we cannot do when publishing a new release on crates.io. There we cannot have dependencies pulled from git :cry:

mgsharm:
Hi @flavio / @loosebazooka, this fix is available today through the official release. Thank you for your patience.

flavio:
@mgsharm thanks! now we have to sort out https://github.com/awslabs/tough/pull/769 

We will need another tough release with whatever fix of the linked issue to have sigstore work :(

PR #780, opened 2024-07-08
Bump rust to 1.78.0 in actions runner

jpculp:
*Description of changes:*

Bumps rust to 1.78.0 and addresses the `clippy` warnings that came with it.

### DIFF
--- a/tuftool/src/error.rs
+++ b/tuftool/src/error.rs
@@ -127,9 +127,6 @@ pub(crate) enum Error {
         backtrace: Backtrace,
     },
 
-    #[snafu(display("Can't build URL from path '{}'", path.display()))]
-    FileUrl { path: PathBuf, backtrace: Backtrace },
-
     #[snafu(display("Failed to write to {}: {}", path.display(), source))]
     FileWrite {
         path: PathBuf,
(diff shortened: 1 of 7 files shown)


bcressey:
We need to update `cargo deny` to fix this error:
```
cargo deny --all-features check --disable-fetch licenses bans sources
thread '<unnamed>' panicked at /home/runner/work/tough/tough/.cargo/registry/src/index.crates.io-6f17d22bba15001f/krates-0.14.0/src/builder.rs:791:47:
called `Option::unwrap()` on a `None` value
```

I'd prefer to keep `_extra` as-is and just add a [lints table](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-lints-table) to the workspace `Cargo.toml` to allow it.

jpculp:
- Rebased on #781 _(which updates `cargo-deny`)_.
- Added `allow(clippy::pub_underscore_fields)` to `tough/src/schema/mod.rs`.

tuftool/src/error.rs
```diff
@@ -373,12 +351,6 @@ pub(crate) enum Error {
         backtrace: Backtrace,
     },
 
-    #[snafu(display("Failed writing target data to disk: {}", source))]
-    WriteTarget {
-        source: std::io::Error,
-        backtrace: Backtrace,
-    },
```
bcressey, line 380: I am surprised that so many errors end up unused.

Over in the core kit, when I was working on signpost, I noticed that a lot of the new "unused" warnings were false positives. Seems to be an issue with a crate that's built as both a library and a binary - the error variants that aren't used in the library side get flagged even though the binary needs them.

Which is to say, please double check via `git grep` or `rg` that all of these are actually not used.

bcressey (approved):

tuftool/src/error.rs
jpculp: I think a couple of these might have just been copied from `tough` at some point. I wasn't able to find any references in `tuftool` itself.

cbgbt (approved):

PR #781, opened 2024-07-08
Update rust dependencies (including AWS SDK)

jpculp:
*Description of changes:*

Bumps various rust dependencies, including the AWS SDK for Rust and its **BehaviorVersion**.

Also bumps `cargo-deny` to **0.14.20**.

### DIFF
--- a/tough-kms/src/client.rs
+++ b/tough-kms/src/client.rs
@@ -8,6 +8,7 @@ use aws_sdk_kms::Client as KmsClient;
 
 /// Builds a KMS client for a given profile name.
 pub(crate) async fn build_client_kms(profile: Option<&str>) -> KmsClient {
+    #[allow(deprecated)] // pinning to v2023_11_09 for now
     let config = aws_config::defaults(BehaviorVersion::v2023_11_09());
     let client_config = if let Some(profile) = profile {
         let region = DefaultRegionChain::builder()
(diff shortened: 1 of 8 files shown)


sam-berning (approved):

tough-kms/src/client.rs
```diff
@@ -8,7 +8,7 @@ use aws_sdk_kms::Client as KmsClient;
 
 /// Builds a KMS client for a given profile name.
 pub(crate) async fn build_client_kms(profile: Option<&str>) -> KmsClient {
-    let config = aws_config::defaults(BehaviorVersion::v2023_11_09());
+    let config = aws_config::defaults(BehaviorVersion::v2024_03_28());
```
tquirk-aws, line 11: I notice that we have to make this same change in 3 or 4 different files.  Could we set this as a constant somewhere that we can import where needed, so we only have to change it in one place?

jpculp: It might not be worth it since `tough-kms` and `tough-ssm` are two separate crates. It might make sense to have `tough-kms` tests reference a version set in `client.rs`, which would bring you from editing 3 files to 2. I would recommend opening a separate issue on this to tackle in another PR.

tquirk-aws (approved):

bcressey (approved):

Issue #796, opened 2024-07-26
Feature Request: Allow using password-protected private signing keys


fghanmi:
This issue suggests implementing support for using password-protected private keys. This feature is generic enough to be valuable for a wide range of use cases.

PR #797, opened 2024-07-26
Allow using password-protected private keys

fghanmi:
*Issue #, if available:*
https://github.com/awslabs/tough/issues/796

*Description of changes:*
- Added support for signing TUF repositories with password-protected private keys.
- Implemented functionality to handle password protection during key management.
- add unit tests 

If the password is invalid, the code will throw the following error:
`Failed to sign repository: Unable to parse keypair: Unrecognized private key format`


fghanmi:
Hello @bcressey, @jpculp, 
Could you please take a look when you have a moment?
Thank you in advance!

jpculp:
Hello @fghanmi! Thank you for this. I'll share this with the rest of the team, but while we take a look at it would you be able to do a `rustfmt`? The actions workflow will fail as is.

tough/Cargo.toml
```diff
@@ -35,6 +35,7 @@ typed-path = "0.9"
 untrusted = "0.9"
 url = "2"
 walkdir = "2"
+openssl = "0.10"
```
jmt-lab: So we have in the near past worked to remove openssl from the dependencies of tough and would prefer to keep it that way. Is there a way to adjust this implementation to rely on rustls instead?

fghanmi: I believe that reading password encrypted private keys is not yet supported by rustls(pemfile) as per this [issue](https://github.com/rustls/pemfile/issues/15). They've suggested to use pkcs8 instead. 
So, I've done the same and removed openssl.

fghanmi:
> Hello @fghanmi! Thank you for this. I'll share this with the rest of the team, but while we take a look at it would you be able to do a `rustfmt`? The actions workflow will fail as is.

Hello @jpculp, 
I tried it and it failed. But, I tried it on `develop` branch and it fails as well. I am not sure if I am using a rust version that is incompatible with this project.

jpculp:
> ﻿I tried it and it failed.

I was able to run `cargo fmt` at the top level of the repo, but we can also push the adjustment if you'd rather we take care of it.

fghanmi:
> > I tried it and it failed.
> 
> I was able to run `cargo fmt` at the top level of the repo, but we can also push the adjustment if you'd rather we take care of it.

@jpculp 
Yes, `cargo fmt` is working fine without any errors on this PR branch.
**I've updated the PR**
Thanks!

tough/Cargo.toml
```diff
@@ -35,6 +35,8 @@ typed-path = "0.9"
 untrusted = "0.9"
 url = "2"
 walkdir = "2"
+pkcs8 = { version = "0.7", features = ["encryption", "pem", "std"] }
+base64 = "0.21"
```
jpculp: Would it be possible to alphabetize these?

tuftool/src/add_key_role.rs
```diff
@@ -66,8 +74,18 @@ impl AddKeyArgs {
...
-        for source in &self.new_keys {
-            let key_source = parse_key_source(source)?;
+        let default_password = String::new();
+        let new_passwords = match &self.new_passwords {
+            Some(pws) => pws,
+            None => &vec![],
+        };
+
+        if new_passwords.len() > self.new_keys.len() {
+            panic!("More new passwords provided than new key sources");
```
jpculp:
```suggestion
            error::MoreNewPasswordsSnafu.fail()?;
```

Would it be possible to replace any panics outside of tests with [snafu](https://docs.rs/snafu/latest/snafu/index.html) errors?

_errors.rs:_
```rust
    #[snafu(display("More new passwords provided than new key sources"))]
    MoreNewPasswords { backtrace: Backtrace },
```

tuftool/src/root.rs
```diff
@@ -303,6 +339,10 @@ impl Command {
         command.arg("-pkeyopt");
         command.arg(format!("rsa_keygen_pubexp:{exponent}"));
 
+        if let Some(password_str) = password.as_deref() {
+            command.args(["-aes256", "-pass"]);
+            command.arg(format!("pass:{}", password_str));
```
jpculp:
```suggestion
            command.arg(format!("pass:{password_str}"));
```

tuftool/src/root.rs
```diff
@@ -239,10 +260,24 @@ impl Command {
     }
 
     #[allow(clippy::borrowed_box)]
-    async fn add_key(path: &Path, roles: &[RoleType], key_source: &Vec<String>) -> Result<()> {
+    async fn add_key(
+        path: &Path,
+        roles: &[RoleType],
+        key_source: &Vec<String>,
```
jpculp:
```suggestion
        key_source: &[String],
```

jpculp (changes requested):
Thank you for the formatting fix! We are still evaluating this, but I left a couple of small changes in the code in the meantime.

fghanmi:
Hello @jpculp , 
any news regarding this PR, please ?

jpculp:
@fghanmi, apologies for the delays. It's still under review, but the gears are in motion. Something that might help is if you are able to include a small unit test for handling a password protected key. The same is true for adding a small test for SHA512 hashes in #814. Thank you for your patience in all of this.

jpculp:
Thank you for the quick follow-up! It looks like if we bump `pkcs8` to **0.10** the duplicate dependencies will drop in the `check-licenses` action, but looks like it requires a tweak in the code. I suspect it requires something close to: 
```rust
    let encrypted_private_key_document = pkcs8::EncryptedPrivateKeyInfo::from_der(pem.contents())?;
    let decrypted_private_key_document =
        encrypted_private_key_document.decrypt(password.as_bytes())?;
    let decrypted_key_bytes: Vec<u8> = decrypted_private_key_document.as_bytes().to_vec();
```

jpculp:
I think we're getting close here. @fghanmi, would you be able to include example output in the description if the password is incorrect?

fghanmi:
> Failed to sign repository: Unable to parse keypair: Unrecognized private key format

@jpculp , I hope I understood you correctly, I've added an error example to the PR description.

jku:
I'm not a maintainer, so I'll leave this as a drive by comment, not a review:

In my opinion "--password" is an anti pattern in any cli: It practically guarantees users are going to make a mistake and the password is then available in shell history and /proc/.

fghanmi:
Hello @jpculp, any update on this, please ?

fghanmi:
Hello @jpculp, any update on this, please ?

jpculp:
Hi @fghanmi, sorry for the radio silence on this. Thank you for all the work that you've done to provide this solution. I've brought this to the team a couple of times, but I think the general consensus is we'd rather not include this feature. In agreement with @jku's comment, it could be considered an anti-pattern for a CLI.

fghanmi:
> Hi @fghanmi, sorry for the radio silence on this. Thank you for all the work that you've done to provide this solution. I've brought this to the team a couple of times, but I think the general consensus is we'd rather not include this feature. In agreement with @jku's comment, it could be considered an anti-pattern for a CLI.

Hello @jpculp @jku , 
Thank you for your feedbacks! 
If I understood correctly the rejection was based on the fact that using password within the CLI is considered as a problem. If yes, I can work on another solution like:

1. reading from env var (tuftool ... --password-env PASSWORD).
2. reading from stdin (tuftool ... --password-stdin)

What do you think ?

PR #820, opened 2024-09-08
Fix 128bit integer serialization

aumetra:
Description of changes: When proptesting this implementation against my own, I noticed that the serialization of 128-bit integers is broken and will produce faulty output.

This PR fixes the serialization.

---


cbgbt (approved):
Nice, thank you for your contribution!

bcressey (approved):

PR #822, opened 2024-09-09
Updated aws-sdk-kms dependencies & use default tls for reqwest

sumukhballal:
*Description of changes:*

- Updated aws-sdk-kms to v1.41. It is required so we get the latest regions supported from the SDK.
- Updated `reqwest` to use default-tls for on-host ca's instead of the one packaged with `rust-tls`.

### DIFF
--- a/tuftool/Cargo.toml
+++ b/tuftool/Cargo.toml
@@ -26,7 +26,7 @@ log = "0.4"
 maplit = "1"
 olpc-cjson = { version = "0.1", path = "../olpc-cjson" }
 rayon = "1"
-reqwest = { version = "0.11", default-features = false, features = ["rustls-tls"] }
+reqwest = { version = "0.11", default-features = false, features = ["rustls-tls-native-roots"] }
 ring = { version = "0.17", features = ["std"] }
 serde = "1"
 serde_json = "1"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -205,9 +205,9 @@ dependencies = [
 
 [[package]]
 name = "aws-credential-types"
-version = "1.2.0"
+version = "1.2.1"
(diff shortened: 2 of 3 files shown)


sumukhballal:
Added some comments for `deny.toml`.

sumukhballal:
Added the `Cargo.lock` file.

sumukhballal:
Added a change suggested by @jpculp !

jpculp (approved):

bcressey (commented):
Can you split this into two commits? One for the dependency update and one for the `reqwest` feature flag change?

tuftool/Cargo.toml
```diff
@@ -26,7 +26,7 @@ log = "0.4"
 maplit = "1"
 olpc-cjson = { version = "0.1", path = "../olpc-cjson" }
 rayon = "1"
-reqwest = { version = "0.11", default-features = false, features = ["rustls-tls"] }
+reqwest = { version = "0.11", default-features = false, features = ["default-tls"] }
```
bcressey: Is this actually the feature you want?  For crates in the core kit, we use `rustls-tls-native-roots`.

bcressey: I think this is not the right feature flag and could result in using `native-tls` on some platforms. We don't want `tuftool` to link with openssl, even if that (indirectly) solves the problem of getting it to use the host's CA cert bundle.

sumukhballal: That makes sense. I have not tested with `rustls-tls-native-roots` but let me run a test.

sumukhballal:
Addressed @bcressey 's comments.

bcressey (approved):

mgsharm (approved):

Issue #824, opened 2024-09-12
Migrate crypto backend from `ring` to `aws-lc-rs`


ginglis13:
The maintainers of `tough` wish to support the `aws-lc-rs` crypto library, which wishes to be a "drop-in replacement for ring that provides FIPS support and is compatible with the ring API".

We have some rough ideas for an approach and would like to get community input before such a change:

1. Completely replace `ring` with `aws-lc-rs` as a drop-in replacement
2. Use a cargo feature to gate usage of `aws-lc-rs` as the backend behind a feature (see `rustls`, which uses `aws-lc-rs` by default and offers `ring` as a feature on the crate: https://github.com/rustls/rustls?tab=readme-ov-file#platform-support)
3. Expose a `CryptoProvider` interface for consumers of the library, with implementations for `ring` and `aws-lc-rs` (a la `rustls`: https://docs.rs/rustls/latest/rustls/crypto/struct.CryptoProvider.html)

We will keep this issue updated with our plans.

cc: @flavio @iliana @fghanmi and others in the community for your thoughts, questions, concerns, etc.

iliana:
Thanks for the ping!

In a vacuum I think **Option 1** makes perfect sense, except that `aws-lc-rs` doesn't currently build on illumos according to a coworker of mine. I think before I commit to saying this option works fine for us I want to chase that issue a bit and see if I can contribute a fix there (and maybe run a nightly illumos CI to catch issues early).

With my ex-maintainer hat on, for whatever that's still worth: I think the best option long-term is **Option 3**. I think it could be tricky to design a `CryptoProvider` interface if you're only building it for `ring` and `aws-lc-rs`, since one is a drop-in replacement for the other; I would recommend also trying to build a non-C-based implementation (e.g. [RustCrypto](https://github.com/RustCrypto) and friends) of that interface to make sure the design of that traits are sensible, perhaps in a separate crate.

iliana:
Update: I tested `aws-lc-rs` v1.9.0 (and also the current git HEAD) on my [Helios](https://github.com/oxidecomputer/helios) system and it built fine, so either it was an environment error or something's changed in the past few months. Either way, I would consider that concern dealt with for the time being.

ginglis13:
[...]

PR #825, opened 2024-09-13
Use `aws-lc-rs` in place of `ring`

ginglis13:
*Issue #, if available:*

#824 

*Description of changes:*

Migrate from `ring` to `aws-lc-rs`. Testing thus far has been simply `cargo test` and `make integ`.

The `untrusted` crate that `tough` depends on is downgraded to `0.7.1` as a consequence of `aws-lc-rs` pinning to this version of the crate; it is now required in `tough` by the `ring-io` default feature of `aws-lc-rs`, which provides `aws_lc_rs::io::der`.


ginglis13:
^ force pushed a rebase

ginglis13:
^ skip untrusted v0.7.1 in license check; also bring `Cargo.lock` up to speed

tough/Cargo.toml
```diff
@@ -11,6 +11,7 @@ edition = "2018"
 [dependencies]
 async-recursion = "1"
 async-trait = "0.1"
+aws-lc-rs = { version = "1", features = ["prebuilt-nasm"] }
```
bcressey, line 14: The docs only mention this feature for Windows - is it relevant for other platforms?

ginglis13: Correct, this is only a feature for the Windows platform

https://github.com/aws/aws-lc-rs/blob/8d428026db8a8d4434d3db50e982ad65a98fdeb3/book/src/requirements/windows.md?plain=1#L11-L12

> Windows requirements
> [...]
> The NASM assembler is recommended on `x86-64` platforms. NASM is required for `x86` and for "fips" builds. 

Setting the feature pulls in the project's NASM objects rather than falling back to those locally available in the environment: https://github.com/aws/aws-lc-rs/tree/main/aws-lc-sys/builder/prebuilt-nasm https://docs.rs/aws-lc-rs/latest/aws_lc_rs/#prebuilt-nasm

ginglis13: If you'd prefer we leave this up to consumers of `tough`, I could see about what's needed to add NASM to our CI (could split out the Windows CI and set `AWS_LC_SYS_PREBUILT_NASM=1`

jpculp: I believe on non-windows platforms the feature is a no-op.

> Enables the use of crate provided prebuilt NASM objects under certain conditions. This only affects builds for Windows x86-64 platforms. This feature is ignored if the "fips" feature is also enabled.

jpculp (approved):

bcressey (approved):

PR #828, opened 2024-09-26
Add FIPS support

ginglis13:
*Issue #, if available:*

*Description of changes:*

- add a `fips` feature to all crates in this project for toggling on/off `fips` features in dependent crates
- explicitly install a default `CryptoProvider` before constructing a `reqwest` HTTP client
- update `reqwest` to 0.12

This change is dependent on #826 (rustls upgrade included there: https://github.com/smithy-lang/smithy-rs/blob/37c1cc0b31fa85057bd294707bdc88d7bc91d627/rust-runtime/aws-smithy-experimental/Cargo.toml#L25-L26)

*Testing*

- [x] `make integ`
- [x] use a fips `tuftool` to pull down Bottlerocket TUF repos

```
$ cargo install --path tuftool --all-features
$ wget https://updates.bottlerocket.aws/2020-07-07/aws-k8s-1.30/aarch64/6.root.json
$ tuftool download -r 6.root.json \
  --targets-url https://updates.bottlerocket.aws/targets \
  --metadata-url https://updates.bottlerocket.aws/2020-07-07/aws-k8s-1.30/aarch64/ \
  test-fips-tuftool
...
        -> bottlerocket-aws-k8s-1.30-aarch64-1.21.1-82691b51-root.verity.lz4
        -> bottlerocket-aws-k8s-1.30-aarch64-v1.20.4-data.img.lz4
```

- [x] test fips `tuftool` against NGINX server using FIPS, and non-FIPS, ciphers, ensure success/failure

1. `$ cargo install --path tuftool --all-features`
2. Start an nginx server with the default configuration, with this minor addition:
```
ssl_conf_command Ciphersuites TLS_CHACHA20_POLY1305_SHA256;
```
3. Create a basic TUF repo:
```
ALIAS_SUFFIX=$(date "+%Y-%m-%d")

# Simple script to create RSA 3072 KMS keys
SCRIPT_PATH=create-new-kms-key.sh
$SCRIPT_PATH alias/bottlerocket-root-${ALIAS_SUFFIX}
$SCRIPT_PATH alias/bottlerocket-publication-${ALIAS_SUFFIX}                                                                                                                                                                                      $SCRIPT_PATH alias/bottlerocket-timestamp-${ALIAS_SUFFIX}

tuftool root init --version 1 1.root.json

expiration_date=$(date -d "${ALIAS_SUFFIX} + 1 year" --iso-8601=date -u)T00:00:00+00:00

tuftool root expire 1.root.json ${expiration_date}

[...]

PR #868, opened 2025-04-14
Bump AWS SDK version

sam-berning:
*Issue #, if available:*

*Description of changes:*

Bump version of AWS SDK to leverage the updated HTTPS stack with `hyper 1.x` and `aws-lc`. See https://github.com/awslabs/aws-sdk-rust/discussions/1257.

Also migrates from `aws-smithy-experimental` to `aws-smithy-https-client`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -470,7 +470,7 @@ dependencies = [
  "regex-lite",
  "roxmltree",
  "serde_json",
- "thiserror 1.0.69",
+ "thiserror",
 ]
 
 [[package]]
@@ -677,7 +677,7 @@ dependencies = [
  "proc-macro2",
  "quote",
  "regex",
- "rustc-hash 1.1.0",
+ "rustc-hash",
  "shlex",
  "syn 2.0.100",
  "which",
@@ -801,12 +801,6 @@ name = "cfg-if"
 version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "baf1de4339761588bc0619e3cbc0120ee582ebb74b53b4efbf79117bd2da40fd"
-
-[[package]]
-name = "cfg_aliases"
-version = "0.2.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "613afe47fcd5fac7ccf1db93babcb082c5994d996f20b8b159f2ad1658eb5724"
 
 [[package]]
 name = "chrono"
@@ -1116,7 +1110,7 @@ dependencies = [
  "anyhow",
  "axum",
  "noxious-client",
- "rand 0.8.5",
+ "rand",
  "serde_json",
  "tempfile",
  "tokio",
@@ -1287,10 +1281,8 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c4567c8db10ae91089c99af84c68c38da3ec2f087c3f82960bcdbf3656b6f4d7"
 dependencies = [
  "cfg-if",
- "js-sys",
  "libc",
  "wasi 0.11.0+wasi-snapshot-preview1",
- "wasm-bindgen",
 ]
 
 [[package]]
@@ -1300,11 +1292,9 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "73fea8450eea4bac3940448fb7ae50d91f034f941199fcd9d909a5a07aa455f0"
 dependencies = [
  "cfg-if",
- "js-sys",
  "libc",
  "r-efi",
  "wasi 0.14.2+wasi-0.2.4",
- "wasm-bindgen",
 ]
 
 [[package]]
@@ -2014,9 +2004,9 @@ dependencies = [
  "futures",
  "mockall_double",
  "pin-project-lite",
- "rand 0.8.5",
+ "rand",
  "serde",
- "thiserror 1.0.69",
+ "thiserror",
  "tokio",
  "tokio-util 0.6.10",
  "tracing",
@@ -2031,7 +2021,7 @@ dependencies = [
  "noxious",
  "reqwest 0.11.27",
  "serde",
- "thiserror 1.0.69",
+ "thiserror",
  "tokio",
 ]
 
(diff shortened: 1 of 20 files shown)


tuftool/Cargo.toml
```diff
@@ -12,14 +12,14 @@ edition = "2018"
 integ = []
 default = ["aws-sdk-rust"]
 aws-sdk-rust = ["aws-sdk-rust-rustls"]
-aws-sdk-rust-rustls = ["aws-config/rustls", "aws-sdk-ssm/rustls", "aws-sdk-kms/rustls", ]
+aws-sdk-rust-rustls = []
```
arnaldo2792, line 16: This isn't quite right, you are turning off `default-features`, which disables the `rustls` support turned on by default in the `aws-sdk-ssm`. You should add back `aws-smithy-runtime/tls-rustls` to this list so that by default, `tls` isn't enabled, but users can still enable it by turning on ` aws-sdk-rust-rustls`. 

The same applies for `aws-sdk-kms`.

sam-berning: The `default-https-client` feature enables the updated HTTPS stack with `hyper 1.x`, `rustls`, and `aws-lc-rs`, so disabling this `rustls` feature doesn't actually disable support for `rustls`. This feature currently just enables the legacy HTTPS stack with `hyper 0.14.x` and `ring`. For more info see this announcement from `aws-sdk-rust`: https://github.com/awslabs/aws-sdk-rust/discussions/1257

I want to disable this feature to avoid needing to build two versions of `hyper` and `rustls` for the updated `aws-sdk` crates.

deny.toml
```diff
@@ -64,35 +64,18 @@ skip = [
...
 skip-tree = [
     # noxious-client is using an older version of reqwest
     { name = "reqwest", version = "=0.11" },
-    # aws-smithy-runtime-experimental is using several newer dependencies
-    { name = "aws-smithy-experimental", version = "=0.1" },
+    # tokio and mio are bringing in older versions of windows libraries while other
+    # libraries, are bringing in newer versions
+    { name = "windows-sys" },
+    # failure-server is an integ test package that pulls in legacy http libraries
+    { name = "failure-server" },
```
cbgbt, line 78: In this case, we're only skipping it for bans but not for licenses, right? That seems like a workable solution.

ginglis13 (approved):
LGTM - verified FIPS testing following https://github.com/awslabs/tough/pull/828#issue-2551692693

tuftool/Cargo.toml
cbgbt: Perhaps we should add a comment similar to what we've [done in coldsnap](https://github.com/awslabs/coldsnap/blob/07ec73153bebda795bf8c19ff39e7bf6bbe6fde9/Cargo.toml#L13)?

deny.toml
cbgbt: If we're ignoring the `failure-server` branch, I wonder if we can add an explicit ban on `ring`?

We do an explicit ban on crates in [bottlerocket-os/bottlerocket](https://github.com/bottlerocket-os/bottlerocket) [here](https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/819a408c89c4023613ff67b04e8a823411af9031/sources/deny.toml#L49)

deny.toml
```diff
@@ -64,35 +64,18 @@ skip = [
...
-    { name = "windows_i686_gnu", version = "=0.48" },
-    { name = "windows_i686_msvc", version = "=0.48" },
-    { name = "windows_x86_64_gnu", version = "=0.48" },
-    { name = "windows_x86_64_gnullvm", version = "=0.48" },
-    { name = "windows_x86_64_msvc", version = "=0.48" },
 ]
 
 skip-tree = [
     # noxious-client is using an older version of reqwest
     { name = "reqwest", version = "=0.11" },
```
cbgbt: Might be able to remove this now that we're skipping `failure-server`?

Cargo.lock
```diff
@@ -3335,8 +3255,8 @@ dependencies = [
  "aws-config",
  "aws-lc-rs",
  "aws-sdk-kms",
- "aws-smithy-experimental",
- "aws-smithy-http",
+ "aws-smithy-http 0.60.11",
```
cbgbt: I see that we have multiple `aws-smithy-http`'s here but no corresponding `skip` in `deny.toml`. Any idea why?

tough/src/lib.rs
```diff
@@ -667,6 +667,7 @@ fn parse_url(url: Url) -> Result<Url> {
 /// Steps 0 and 1 of the client application, which load the current root metadata file based on a
 /// trusted root metadata file.
 #[expect(clippy::too_many_arguments)]
+#[allow(clippy::needless_continue)]
```
cbgbt, line 670: Can you use `expect` instead of `allow`?

deny.toml
sam-berning: > In this case, we're only skipping it for bans but not for licenses, right? That seems like a workable solution.

Yes, this only skips it for bans.

> If we're ignoring the `failure-server` branch, I wonder if we can add an explicit ban on `ring`?

The feature set of `reqwest` that we're using in `tuftool` still depends on `ring`: https://github.com/seanmonstar/reqwest/blob/master/Cargo.toml#L52. We'd need to remove that dependency before we can deny `ring`.

Cargo.lock
sam-berning: No, that's really strange, I'm not sure how that wasn't flagged by `cargo-deny`! The `0.60.x` version comes from `tough-kms`'s dev-dependencies, where we specifically use that version. I'll bump it to `0.62`.

jpculp: I've noticed sometimes the `aws-sdk-rust` crates will get ahead of themselves and pull multiple `aws-smithy-http`s. You might be able to shake it by pulling newer (but not the newest) SDK crates.

tough/src/lib.rs
sam-berning: Hm, tried this and apparently not. There seems to be something odd with the macos Github actions runner, it doesn't register a `needless_continue` in this function so it will fail the `expect` (and it's also had a bunch of other strange failures on this PR). I couldn't replicate this on my mac, but switching this back to `allow` to pass the workflows.

cbgbt: This suggests to me that the Mac actions runner must be using an older rust toolchain.

I think it points out to me that we probably don't want this to fail if you are using an older rust toolchain, so I'm OK to leave it as-is.

cbgbt (approved):

arnaldo2792 (approved):

PR #871, opened 2025-04-18
github-actions: update Mac runners to macos-15

sam-berning:
*Issue #, if available:*

*Description of changes:*

I've been observing some strange test failures on the Mac runners on some of my recent PRs. The `tuftool clone` tests fail about 1 in every 10 runs, and every time it's because the `1.root.json` that was copied is actually just an empty file.

This hasn't happened on runners for any other OS, and when I tried to replicate it on my Mac that's running MacOS 15, I wasn't able to. Attempting to bump the runners to macos-15 to see if it resolves this issue, otherwise I will continue investigating this as a potential bug.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -17,13 +17,13 @@ jobs:
       fail-fast: false
       matrix:
         make_target: ["check-licenses", "build", "integ", "integ-fips"]
-        os: [ubuntu-latest, windows-latest, macos-latest]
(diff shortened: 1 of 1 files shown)


sam-berning:
We just observed this failure on an ubuntu runner in another PR: https://github.com/awslabs/tough/actions/runs/14542260946/job/40802332867?pr=869

So we can rule out that this is a mac issue. Canceling this PR.

PR #873, opened 2025-04-18
Pin tokio to its LTS version

sam-berning:
*Issue #, if available:*

#872 

*Description of changes:*

In #868, we bumped `tokio` from `1.42.x` to `1.44.x`. Since the failing tests in #872 are using async file read, I believe our code may be incompatible with the new `tokio` version. Given that `1.43` is the latest LTS version [listed on tokio's repo](https://github.com/tokio-rs/tokio/tree/master/tokio#bug-patching-policy), I'm pinning to that version.


sam-berning:
This has been superseded by #874 and is no longer needed

Issue #766, opened 2024-04-26
Be less strict in keyid validity check ?


jku:
I'm wondering if you'd accept a patch that makes the key deserialization checks a little bit less strict...

I've tried to write down all of the the context in this spec issue https://github.com/theupdateframework/specification/issues/305 but the short of it is that
* clients calculating the keyid from the canonicalized key content (see `deserialize_keys()`) as the spec instructs seems to have no security benefit
* it does have significant potential for compatibility issues
* and can be a headache for repository implementations

The practical reason I'm filing this now is this:
* tuf-on-ci (the repository implementation I maintain) had a [bug](https://github.com/theupdateframework/tuf-on-ci/issues/292) that produced keys with keyids that are not strictly conformant: 
* I intend to fix this bug but in the mean time root-signing-staging  (Sigstores _staging_ TUF repository) already has metadata with these keyids
* none of the existing sigstore clients seem to verify the keyid calculation so we did not notice, and as documented in the linked issue there seems to be rough consensus in the TUF community that keyid calculation is not useful  

So I'm trying to clean up my mess, wondering if (in addition to fixing the bug in tuf-on-ci) part of that could be a patch to tough that removes the keyid calculation check from the key deserialization as a way to improve practical compatibility between the implementations.

rpkelly:
Hi, thanks for reaching out! We're reviewing this proposal with our security team and will be able to give you an answer once we've gone through that process.

webern:
Still reviewing. In reading the TAP12 proposal, as I understand it, I agree with the change. Most likely we would prefer to see the change incorporated into the spec before making the change here.

One thing not entirely clear to me @jku, are there downstream users of this library that are blocked by the enforcement of the KeyID calculation? I saw that sigstore uses this library in its "experiemental" Rust client. Is that causing blockers or failures now?

jku:
Cheers, I can understand that view point. I will likely try to push the linked spec issue forward before the TAP to make the spec intent clear but your point applies there as well: you'd rather see it in the spec first...

So the situation is that 
* sigstore staging TUF repo now has these incorrect keyids in their current root metadata versions, this is unlikely to revert
* this is annoying but not a complete blocker since A) this is a staging repo, we don't intend to reproduce the issue in production and B) once we figure out how to create new root metadata that is compliant in a smooth upgrade, the experimental rust client can again use the staging repo as long as it starts with root N+1

[...]

PR #886, opened 2025-06-21
tough: condition tap-12 check with feature

glebiller:
*Issue #, if available:*

https://github.com/awslabs/tough/issues/766

*Description of changes:*

Conditionally disable the keyid with a feature `tap-12`. The default behavior remain unchanged, but it can be disable easily.
Since the [TAP-12 is accepted](https://github.com/theupdateframework/taps/blob/master/tap12.md), and hopefully should be merged in a near future, the goal is to unblock some TUF repos from being used with tough (like the [GitHub TUF repo](https://tuf-repo.github.com/)).

### DIFF
--- a/tough/src/schema/de.rs
+++ b/tough/src/schema/de.rs
@@ -22,15 +22,18 @@ where
         key: Key,
         map: &mut HashMap<Decoded<Hex>, Key>,
     ) -> Result<(), error::Error> {
-        let calculated = key.key_id()?;
         let keyid_hex = hex::encode(&keyid);
-        ensure!(
-            keyid == calculated,
-            error::InvalidKeyIdSnafu {
-                keyid: &keyid_hex,
-                calculated: hex::encode(&calculated),
-            }
-        );
+        #[cfg(not(feature = "tap-12"))]
+        {
+            let calculated = key.key_id()?;
+            ensure!(
+                keyid == calculated,
+                error::InvalidKeyIdSnafu {
+                    keyid: &keyid_hex,
+                    calculated: hex::encode(&calculated),
+                }
+            );
+        }
         ensure!(
             map.insert(keyid, key).is_none(),
             error::DuplicateKeyIdSnafu { keyid: keyid_hex }
--- a/tough/Cargo.toml
+++ b/tough/Cargo.toml
@@ -50,6 +50,7 @@ tokio-test = "0.4"
 [features]
 fips = ["aws-lc-rs/fips", "rustls/fips"]
 http = ["reqwest"]
+tap-12 = []
 
 # The `integ` feature enables integration tests. These tests require `noxious-server` to be installed on the host.
 integ = []


cbgbt:
Thanks for the contribution!

I'd like to unblock `tough` users, but I am a bit torn on the conditional compilation approach. Feature unification could enable this in surprising cases.

I'm unfamiliar with the TAP process, do you have a sense of how long it takes for such proposals to merge with the spec?

It seems like the CI is failing due to an unrelated clippy lint. We can look to merge a fix for that separately.

glebiller:
> I'm unfamiliar with the TAP process, do you have a sense of how long it takes for such proposals to merge with the spec?

Thanks for your reply, unfortunately I am not really close to the TAP process too, however you can see from the previous TAP that it takes years to officially be added to the spec once approved.

However, I checked the official implementation [theupdateframework/python-tuf](https://github.com/theupdateframework/python-tuf) and they already implement TAP-12 as it do not enforce the keyid check, and the client can be used with the GitHub TUF repo without errors.

tough/src/schema/de.rs
```diff
@@ -22,15 +22,18 @@ where
...
+        {
+            let calculated = key.key_id()?;
+            ensure!(
+                keyid == calculated,
+                error::InvalidKeyIdSnafu {
+                    keyid: &keyid_hex,
+                    calculated: hex::encode(&calculated),
+                }
+            );
+        }
```
bcressey, line 36: My reading of TAP-12 suggests that it's not sufficient to just remove this check, because two different key IDs can now refer to the same key:
> To ensure that a client does not use the same key to verify a signature more than once, they must additionally check that all keys applied to a signature threshold are unique. So, the specification should additionally require that "Clients MUST use each key only once during a given signature verification."

The existing code may not implement this requirement because before there was a 1:1 mapping between key IDs and keys, so the `ensure!` check here was sufficient.

procr1337: @bcressey In the interest of figuring out if a change similar to this could potentially be landed, would you consider this issue to be addressed properly if only unique `key_id()` values were to be counted towards the thresholds in the relevant places?

* https://github.com/awslabs/tough/blob/0ad97ea46542c3f3377ded0753505df8b23f392c/tough/src/schema/verify.rs#L37-L48
* https://github.com/awslabs/tough/blob/0ad97ea46542c3f3377ded0753505df8b23f392c/tough/src/schema/verify.rs#L80-L90

This would seem to be in the spirit of the TAP-12 proposal, as long as `.key_id()` qualifies as a "standardized representation for keys" (which it should, otherwise the previous logic already allows for colliding keys under different key IDs) , and furthermore should retain the previous guarantees.

procr1337:
@cbgbt We are also interested in solving the same problem. What alternative to a feature flag would you suggest? The way that other libraries went is to unconditionally remove the key ID checks, but you may not want to go to that extent

tough/src/schema/de.rs
```diff
@@ -22,15 +22,18 @@ where
         key: Key,
         map: &mut HashMap<Decoded<Hex>, Key>,
```
emilyalbini, line 23: This PR still assumes the key ID will be hex-encoded, which is not mandated by TAP 12:

> The KEYID is an identifier for the key that is determined by the metadata owner and MUST be unique within the delegating metadata file (either root or delegating targets metadata).

PR #932, opened 2026-04-23
fix: update file writing

sky1122:
*Description of changes:*
update file writing

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #933, opened 2026-04-23
fix: update delegations code

rpkelly:
*Description of changes:*
Update delegated target handling.

### DIFF
--- a/tough/src/cache.rs
+++ b/tough/src/cache.rs
@@ -119,13 +119,12 @@ impl Repository {
 
         for name in self.targets.signed.role_names() {
             if let Some(filename) = self.delegated_filename(name) {
-                self.cache_file_from_transport(
-                    filename.as_str(),
-                    self.limits.max_targets_size,
-                    "max_targets_size argument",
-                    &metadata_outdir,
-                )
-                .await?;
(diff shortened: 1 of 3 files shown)


tough/src/cache.rs
```diff
@@ -119,13 +119,12 @@ impl Repository {
...
         for name in self.targets.signed.role_names() {
             if let Some(filename) = self.delegated_filename(name) {
-                self.cache_file_from_transport(
-                    filename.as_str(),
-                    self.limits.max_targets_size,
-                    "max_targets_size argument",
-                    &metadata_outdir,
-                )
-                .await?;
+                if let Some(bytes) = self.delegated_metadata_bytes.get(name) {
```
jmt-lab, line 122: This could be folded (surprised clippy didn't get ya for it)

```
if let Some(filename) = self.delegated_filename(name) && let Some(bytes) = self.delegated_metadata_bytes.get(name)
```

This might be tied to edition 2024 not sure

jmt-lab (approved):

tough/src/cache.rs
rpkelly: #932 fixes the too_many_lines error. I'll repush once that merges

qianxjcraig (approved):

ginglis13 (approved):

PR #936, opened 2026-05-03
Add client configuration to HttpTransportBuilder

NobodyXu:
Allow library users to override the client

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/tough/src/http.rs
+++ b/tough/src/http.rs
@@ -14,6 +14,7 @@ use snafu::Snafu;
 use std::cmp::Ordering;
 use std::future::Future;
 use std::pin::Pin;
+use std::sync::OnceLock;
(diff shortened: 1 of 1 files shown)


sky1122:
Hi @NobodyXu, thanks for the PR! Could you share more about your use case about what do you need the custom client for? We have a use case where we need to control the CryptoProvider to select different cipher suites depending on the deployment environment. Would a custom `CryptoProvider` solve your need as well(#940), or do you specifically need full client override for other reasons?

NobodyXu:
Hello, the motivation is to
 - share the `reqwest::Client` between different http requests so that the connection pool can be reused
 - Configuration of my `reqwest::Client` can apply to tough too

Issue #939, opened 2026-05-14
Add support for custom CryptoProvider in HttpTransportBuilder


sky1122:
Allow library consumers to pass a custom rustls::crypto::CryptoProvider to HttpTransportBuilder, enabling per-client
TLS configuration without relying on the global default provider.

PR #940, opened 2026-05-14
tough: add support for custom CryptoProvider in HttpTransportBuilder

sky1122:
**Issue** #939

**Description of changes:**
Allow library consumers to pass a custom rustls CryptoProvider to HttpTransportBuilder, enabling per-client TLS configuration without relying on the global default provider.

When a custom provider is set, the transport builds a rustls ClientConfig with native root certificates and ALPN protocols. When no custom provider is set, the existing behavior is preserved (global default provider via install_default).

**Test:**
testing with t[his PR](https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/935) and [this code chunk](https://gist.github.com/sky1122/cf0482d9328c0683e0b2d8c9719cbaf3)

```
 cargo test --features http -p tough --test http test_fips_crypto_provider_e2e             
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.14s
     Running tests/http.rs (target/debug/deps/http-e659a739e99ce22d)

running 1 test
test http_custom_provider::test_fips_crypto_provider_e2e ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 3 filtered out; finished in 0.00s
```


tough/src/http.rs
```diff
@@ -32,14 +33,15 @@ use url::Url;
 ///
 /// See [`HttpTransport`] for proxy support and other behavior details.
 ///
-#[derive(Clone, Copy, Debug)]
+#[derive(Clone, Debug)]
```
ginglis13, line 36: we should make sure to call out in release notes that `Copy` is being dropped for `HttpTransportBuilder`, `HttpTransport`, and `DefaultTransport`, so downstreams are aware

tough/src/http.rs
```diff
@@ -329,11 +342,27 @@ impl RetryStream {
...
+        // If a custom CryptoProvider was supplied, build a rustls ClientConfig with it
+        // and configure reqwest to use it. Otherwise, reqwest uses the global default.
+        if let Some(ref provider) = self.settings.crypto_provider {
+            let mut root_store = rustls::RootCertStore::empty();
+            for cert in rustls_native_certs::load_native_certs().certs {
+                let _ = root_store.add(cert);
+            }
+            let mut tls_config = rustls::ClientConfig::builder_with_provider(provider.clone())
+                .with_safe_default_protocol_versions()
+                .expect("valid TLS protocol versions")
```
ginglis13: let's add a snafu instead of potential panic here

tough/src/http.rs
```diff
@@ -329,11 +342,27 @@ impl RetryStream {
...
-            .context(HttpClientSnafu)?;
+            .connect_timeout(self.settings.connect_timeout);
+
+        // If a custom CryptoProvider was supplied, build a rustls ClientConfig with it
+        // and configure reqwest to use it. Otherwise, reqwest uses the global default.
+        if let Some(ref provider) = self.settings.crypto_provider {
+            let mut root_store = rustls::RootCertStore::empty();
+            for cert in rustls_native_certs::load_native_certs().certs {
+                let _ = root_store.add(cert);
+            }
```
ginglis13: you could make this

```
  let (added, ignored) = root_store.add_parsable_certificates(
      rustls_native_certs::load_native_certs().certs
  );
```

and warn about ignored certs. Shouldn't we error on 0 added certs?

sky1122:
force pushed to address feedback.

ginglis13 (approved):

tough/src/http.rs
```diff
@@ -329,11 +342,32 @@ impl RetryStream {
...
+                    "{ignored} native root certificate(s) could not be parsed and were ignored"
+                );
+            }
+            snafu::ensure!(added > 0, NoValidRootCertificatesSnafu);
+            let mut tls_config = rustls::ClientConfig::builder_with_provider(provider.clone())
+                .with_safe_default_protocol_versions()
+                .map_err(|_| HttpError::InvalidTlsConfig)?
+                .with_root_certificates(root_store)
+                .with_no_client_auth();
+            tls_config.alpn_protocols = vec![b"http/1.1".to_vec()];
```
jmt-lab: Hardcoding ALPN to only http/1.1 disables HTTP/2 for any user who supplies a custom provider. The default path (reqwest's built-in TLS) supports h2. Opting into a custom provider for FIPS shouldn't unexpectedly downgrade the protocol.

Does FIPS not support http/2 or something?

tough/Cargo.toml
```diff
@@ -48,8 +49,8 @@ tokio = { version = "1", features = ["macros", "rt", "rt-multi-thread"] }
 tokio-test = "0.4"
 
 [features]
-fips = ["aws-lc-rs/fips", "rustls/fips"]
-http = ["reqwest"]
+fips = ["aws-lc-rs/fips"]
```
jmt-lab: This removal of rustls/fips, is this because we are no longer using rustls at all? If not shouldn't we keep this turned on?

tough/src/http.rs
```diff
@@ -329,11 +342,32 @@ impl RetryStream {
...
+            let mut root_store = rustls::RootCertStore::empty();
+            let certs = rustls_native_certs::load_native_certs().certs;
+            let (added, ignored) = root_store.add_parsable_certificates(certs);
+            if ignored > 0 {
+                log::warn!(
+                    "{ignored} native root certificate(s) could not be parsed and were ignored"
+                );
+            }
+            snafu::ensure!(added > 0, NoValidRootCertificatesSnafu);
+            let mut tls_config = rustls::ClientConfig::builder_with_provider(provider.clone())
```
jmt-lab: Is there an opportunity to cache this setup instead of redoing it on every call?

AI Finding:
poll_new_request is called on every fetch attempt and every retry. Each time it:
• calls rustls_native_certs::load_native_certs() (touches the OS keychain / filesystem)
• builds a RootCertStore
• constructs a new ClientConfig
• builds a new reqwest Client

The reqwest client rebuild is pre-existing, but this PR piles expensive cert-store assembly on top. Cache the ClientConfig (or the Client) once at construction. Build failures should also surface at builder time, not at first request.

tough/src/http.rs
```diff
@@ -329,11 +342,32 @@ impl RetryStream {
...
+                .with_safe_default_protocol_versions()
+                .map_err(|_| HttpError::InvalidTlsConfig)?
```
jmt-lab: Discards the underlying rustls error. Inconsistent with the rest of the file, which uses .context(...) for snafu sources. Make InvalidTlsConfig { source: rustls::Error } and use .context(InvalidTlsConfigSnafu)?.

tough/src/http.rs
```diff
@@ -329,11 +342,32 @@ impl RetryStream {
...
-            .connect_timeout(self.settings.connect_timeout)
-            .build()
-            .context(HttpClientSnafu)?;
+            .connect_timeout(self.settings.connect_timeout);
+
+        // If a custom CryptoProvider was supplied, build a rustls ClientConfig with it
+        // and configure reqwest to use it. Otherwise, reqwest uses the global default.
+        if let Some(ref provider) = self.settings.crypto_provider {
+            let mut root_store = rustls::RootCertStore::empty();
+            let certs = rustls_native_certs::load_native_certs().certs;
```
jmt-lab: FInding:

The default path (reqwest with its bundled TLS stack) uses platform-aware verification. The custom-provider path uses rustls-native-certs — a one-shot snapshot of static roots. On macOS/Windows these are not equivalent (no dynamic trust evaluation, no revocation). Either document the difference clearly or use rustls-platform-verifier for the custom path too.

I'm not sure if this is really a concern here, but the callout on macOS/Windows does seem worrisome, but feel free to teach me where i'm wrong.

sky1122:
force pushed
- ALPN protocol support: Added h2 alongside http/1.1 in ALPN negotiation so that supplying a custom CryptoProvider does not unexpectedly downgrade connections to HTTP/1.1 only.
- Cached TLS configuration: Moved TLS setup from poll_new_request into build()
- Platform-aware certificate verification: Replaced rustls-native-certs (one-shot static root snapshot) with rustls-platform-verifier
- Restored rustls/fips feature

jmt-lab (approved):

PR #943, opened 2026-05-21
tough: make rustls optional via http-rustls-tls / http-native-tls features

wolfv:
## Summary

`tough` currently has `rustls = \"0.23\"` as an unconditional, non-optional
dependency, even though `rustls` is only used in `src/http.rs` — which is
already gated behind `#[cfg(feature = \"http\")]` in `lib.rs`. The single use
site is `HttpTransportBuilder::default()`, which calls
`aws_lc_rs::default_provider().install_default()` to set rustls' process-wide
crypto provider.

This forces `rustls` (and aws-lc-rs's full compile cost) on every consumer of
`tough`, including consumers using `reqwest/native-tls` who have no use for
rustls at all. For example, [rattler](https://github.com/conda/rattler) hits
this through the sigstore-rs dependency chain when building with `native-tls`.

### Why we can drop `rustls` entirely

reqwest 0.13 changed how the rustls crypto provider is selected.

- **reqwest 0.12** read `CryptoProvider::get_default()` and fell back to
  `ring` if nothing was installed
  ([client.rs#L763-L771](https://github.com/seanmonstar/reqwest/blob/v0.12.28/src/async_impl/client.rs#L763-L771)).
  Tough's pre-install was needed so reqwest would pick `aws-lc-rs` over `ring`.
- **reqwest 0.13** constructs `aws_lc_rs::default_provider()` directly when
  the `rustls` feature is enabled
  ([client.rs#L2458-L2465](https://github.com/seanmonstar/reqwest/blob/v0.13.3/src/async_impl/client.rs#L2458-L2465)).
  It no longer reads from `CryptoProvider::get_default()`, so tough's
  pre-install is dead code as of the 0.13 bump.

`tuftool` already installs its own `aws_lc_rs::default_provider()` provider
in `main.rs:74-78`, so it is unaffected.

## Changes

- Drop `rustls` from `tough/Cargo.toml`'s `[dependencies]`.
- Remove the `use rustls::crypto::{aws_lc_rs, CryptoProvider};` import and
  the `install_default()` block in `HttpTransportBuilder::default()`.
- Add new features mirroring reqwest's TLS naming:
  - `http-rustls-tls = [\"http\", \"reqwest/rustls\"]`
  - `http-native-tls = [\"http\", \"reqwest/native-tls\"]`
  - `http` itself no longer enables any TLS backend (matches the previous
    behavior where `reqwest = { default-features = false, features = [\"stream\"] }`
    already didn't pick a backend).
- `fips` drops the `rustls/fips` half — tough no longer depends on rustls,
  so downstream crates that bring in rustls themselves should enable
  `rustls/fips` from their own manifest.

> Note: in reqwest 0.13 the rustls feature was renamed from `rustls-tls`
> to `rustls`. The Cargo.toml comment flags this for anyone migrating.

## Migration

- Users on `features = [\"http\"]` who relied on tough configuring rustls
  for them should switch to `features = [\"http-rustls-tls\"]`.
- Users with TLS already configured elsewhere (e.g., their own
  `reqwest` dep with `rustls`/`native-tls`) need no changes — `\"http\"`
  alone is sufficient, just like before.

## Test plan

- [x] `cargo check -p tough` with each of: default, `http`, `http-rustls-tls`,
      `http-native-tls`, `fips` — all pass
- [x] `cargo tree -i rustls --features http-native-tls` confirms no rustls
      in the tree
- [x] Full workspace builds (`cargo build --workspace`)
- [x] `cargo test -p tough --features http-rustls-tls` — all tests pass

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -784,6 +784,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9330f8b2ff13f34540b44e946ef35111825727b38d33286ef986142615121801"
 
 [[package]]
+name = "cfg_aliases"
+version = "0.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "613afe47fcd5fac7ccf1db93babcb082c5994d996f20b8b159f2ad1658eb5724"
+
+[[package]]
 name = "chacha20"
 version = "0.10.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1284,8 +1290,10 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ff2abc00be7fca6ebc474524697ae276ad847ad0a6b3faa4bcb027e9a4614ad0"
 dependencies = [
  "cfg-if",
+ "js-sys",
  "libc",
  "wasi",
+ "wasm-bindgen",
 ]
 
 [[package]]
@@ -1295,9 +1303,11 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "899def5c37c4fd7b2664648c28120ecec138e4d395b459e5ca34f9cce2dd77fd"
 dependencies = [
  "cfg-if",
+ "js-sys",
  "libc",
  "r-efi 5.3.0",
  "wasip2",
+ "wasm-bindgen",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


Sparksssj (approved):

wolfv:
Not really relevant for us anymore as we have our own TUF impl now: https://github.com/sigstore/sigstore-rust/tree/main/crates/sigstore-tuf

PR #944, opened 2026-05-29
Implement TAP 12

emilyalbini:
*Issue #, if available:* https://github.com/awslabs/tough/issues/766

*Description of changes:*

This PR implements optional support for the accepted [TAP 12](https://github.com/theupdateframework/taps/blob/master/tap12.md), removing the requirements that Key IDs must be a hash of the key. Support for arbitrary Key IDs can be enabled by passing `.key_id_format(KeyIdFormat::Any)` during repository loading.

This PR is best reviewed commit-by-commit: the first commit switches the codebase to use a `KeyId` newtype to represent Key IDs (I would've had to otherwise switch everything to `String`, as there is no requirement in TAP 12 that Key IDs must be hex-encoded), the second commit deduplicates the validation logic, and the third commit actually implements the feature.

Allowing users to choose between the TUF 1.0.0 and TAP 12 behavior regarding Key IDs means the `KeyIdFormat` enum has to be passed around the codebase a lot, if you all prefer I can switch the PR to unconditionally use the TAP 12 behavior instead.

Support for TAP 12 is required to load [GitHub's TUF repo](https://tuf-repo.github.com/). Note that #886 was also opened to support TAP 12, but it used a crate feature to toggle the behavior.

### DIFF
--- a/tough/src/editor/mod.rs
+++ b/tough/src/editor/mod.rs
@@ -14,8 +14,7 @@ use crate::editor::targets::TargetsEditor;
 use crate::error::{self, Result};
 use crate::fetch::fetch_max_size;
 use crate::key_source::KeySource;
-use crate::schema::decoded::{Decoded, Hex};
-use crate::schema::key::Key;
+use crate::schema::key::{Key, KeyId};
 use crate::schema::{
     Hashes, KeyHolder, Metafile, PathSet, Role, RoleType, Root, Signed, Snapshot, Target, Targets,
     Timestamp,
@@ -669,7 +668,7 @@ impl RepositoryEditor {
         paths: PathSet,
         terminating: bool,
         threshold: NonZeroU64,
-        keys: Option<HashMap<Decoded<Hex>, Key>>,
+        keys: Option<HashMap<KeyId, Key>>,
     ) -> Result<&mut Self> {
         let limits = self.limits.context(error::MissingLimitsSnafu)?;
         let transport = self
--- a/tough/src/editor/mod.rs
+++ b/tough/src/editor/mod.rs
@@ -82,6 +82,8 @@ pub struct RepositoryEditor {
 
     transport: Option<Box<dyn Transport>>,
     limits: Option<Limits>,
+
+    key_id_format: KeyIdFormat,
 }
 
 impl RepositoryEditor {
@@ -136,6 +138,7 @@ impl RepositoryEditor {
             signed_targets: None,
             transport: None,
             limits: None,
+            key_id_format: KeyIdFormat::HashedKey,
         })
     }
 
@@ -153,6 +156,7 @@ impl RepositoryEditor {
         editor.timestamp(repo.timestamp.signed)?;
         editor.transport = Some(repo.transport.clone());
         editor.limits = Some(repo.limits);
+        editor.key_id_format = repo.key_id_format;
         Ok(editor)
     }
 
@@ -561,7 +565,7 @@ impl RepositoryEditor {
                     })?;
             (KeyHolder::Delegations(parent), &mut targets.signed)
         };
-        parent.verify_role(&role, name, KeyIdFormat::Any)?;
+        parent.verify_role(&role, name, self.key_id_format)?;
         // Make sure the version isn't downgraded
         ensure!(
             role.signed.version >= current_targets.version,
--- a/tough/src/editor/mod.rs
+++ b/tough/src/editor/mod.rs
@@ -20,7 +20,7 @@ use crate::schema::{
     Timestamp,
 };
 use crate::transport::{IntoVec, Transport};
-use crate::{encode_filename, Limits};
+use crate::{encode_filename, KeyIdFormat, Limits};
 use crate::{Repository, TargetName};
 use aws_lc_rs::digest::{SHA256, SHA256_OUTPUT_LEN};
 use aws_lc_rs::rand::SystemRandom;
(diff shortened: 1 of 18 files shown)


wolfv:
I wanted to voice my support here. We're rolling out sigstore to our clients and would need that.

Additionally "mise" also switched to our sigstore implementation and would benefit from TUF loading of GitHub roots.

jmt-lab:
Looks like you have some cargo fmt issues going on, if you can fix those we can approve. Other than that the code looks solid. Thanks for your contribution!

emilyalbini:
Fixed rustfmt and clippy (`make build` now works locally!).

tough/src/schema/key.rs
```diff
@@ -230,6 +230,29 @@ impl FromStr for Key {
     }
 }
 
+/// Unique identifier representing a key.
+///
+/// TUF version 1.0.0 (section 4.2) mandates that KEYID must be hexdigest of the SHA-256 hash of
+/// the canonical JSON form of the key. TAP 12 changes the definition to allow arbitrary strings to
+/// be key IDs.
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq, Hash)]
+#[serde(transparent)]
```
ginglis13, line 288: Previously with hex decoding, comparisons were case insensitive: https://github.com/awslabs/tough/blob/466c10462d4fc4060d4e6628d67c3c8cebdece35/tough/src/schema/decoded.rs#L220-L224 now, key comparisons will be case sensitive. Doubtful there's many out there using uppercase hex, but I believe this would be a regression

tough/src/editor/mod.rs
```diff
@@ -561,7 +561,7 @@ impl RepositoryEditor {
                     })?;
             (KeyHolder::Delegations(parent), &mut targets.signed)
         };
-        parent.verify_role(&role, name)?;
+        parent.verify_role(&role, name, KeyIdFormat::Any)?;
```
ginglis13: Shouldn't `RespositoryEditor` respect a user's `KeyIdFormat` ? with this hardcoded, I could:

1. `RepositoryLoader::new(...).key_id_format(KeyIdFormat::HashedKey).load()`
2. `RepositoryEditor::from_repo(root_path, repo)`
3. Editor internally calls `verify_role(..., KeyIdFormat::Any)`,  accepts arbitrary key IDs
4. Delegated targets with non-hash key IDs pass the editor's check

tough/src/schema/verify.rs
```diff
@@ -77,6 +91,34 @@ fn verify_common<T: Role + Serialize>(
...
+                                calculated: canonical_key_id,
+                            }
+                        );
+                    }
+                    KeyIdFormat::Any => {
+                        // TAP 12 proposes the wording "Clients MUST use each key only once during a
+                        // given signature verification". As the canonical key ID returned by the
+                        // key_id() method is a hash of the underlying key, we can check that no
+                        // other role has the same canonical key ID.
+                        if let Some(other) = contained_canonical_keyids
```
ginglis13: `key_id()` produces different outputs for the same public key when `_extra` fields differ, since it hashes the entire serialized `Key` including `_extra`. This means `contained_canonical_keyids` won't detect the collision, which could allow one physical key to count multiple times toward the signature threshold.

If I were to list the same key twice in the keys map with different extra JSON fields to get distinct canonical IDs, both entries verify against the same private key.

Consider deduplicating on actual key material (keytype + scheme + keyval bytes) rather than `key_id()`.

ginglis13 (changes requested):
Thanks for the contribution!

tough/src/schema/verify.rs
emilyalbini: Good catch. I also applied the same check to `KeyIdFormat::HashedKey`, since it's possible to have two keys with the same underlying material there too.

tough/src/editor/mod.rs
emilyalbini: I did put `KeyIdFormat::Any` thinking that the editor shouldn't fail regardless of which key ID you give it, but your reasoning makes sense. Changed the code!

tough/src/schema/key.rs
emilyalbini: I'm not sure we should *always* treat key IDs as case insensitive, as TAP 12 doesn't mention IDs should be compared ignoring case. It's also not possible to pass `KeyIdFormat` here (or in the `Hash` implementation).

Where I do see value adding a comparison ignoring case is in the check for `KeyIdFormat::HashedKey`, where we compare the ID against the one generated by the library, as otherwise uppercase IDs would always be rejected.

*As far as I know*, in every other case this would only fail if a repository contains a mix of lowercase and uppercase for the same key ID, which I'd argue is user error? Let me know if there are other places I missed where the casing might matter.

emilyalbini:
Any update on this PR?

wolfv:
We finally had to create our own TUF client: https://github.com/sigstore/sigstore-rust/tree/main/crates/sigstore-tuf

If it is useful for you we could turn it into a standalone repository. You should be able to use it as a standalone crate. Works really well so far and passes the whole conformance test suite.

emilyalbini:
I don't think there is capacity at $dayjob to switch to a different TUF client, we're just depending on our tough fork to use this PR. Still, we would love for the PR to eventually be merged (git dependencies are not the best thing in the world).

emilyalbini:
Rebased on top of develop. In the latest commit, I also changed the check that prevents multiple keys with the same underlying material to skip the signature rather than erroring out. This matches the behavior of [python-tuf](https://github.com/theupdateframework/python-tuf/pull/1091).

Sparksssj:
Hi! 
Blocker — CI is red on cargo fmt --check. The July rebase reintroduced a formatting violation.

emilyalbini:
Pushed a fix!

emilyalbini:
Fixed some unrelated Clippy failures that showed up.

PR #947, opened 2026-06-03
tuftool: truncate parsed expires datetimes to whole seconds

### LINKED ISSUE #447: tough: dates with microseconds may be non-compliant
> What you quoted from the specification might imply that fractions of a second must not be included, but it doesn't say that explicitly.

Spec is explicit about time formats ( this is not linked to from expiry sections but it is in there):

> Metadata [date-time](https://theupdateframework.github.io/specification/latest/#date-time) follows the ISO 8601 standard. The expected format of the combined date and time string is "YYYY-MM-DDTHH:MM:SSZ". Time is always in UTC, and the "Z" time zone designator is attached to indicate a zero UTC offset. An example date-time string is "1985-10-21T01:21:00Z".

This is very clear: timezones and microseconds are not allowed

__Originally posted by @jku in https://github.com/awslabs/tough/pull/426#issuecomment-1037071702__

arpitjain099:
## Issue

Fixes #447.

`tuftool` can write TUF metadata `expires` timestamps that carry a sub-second component, which the TUF specification does not allow. The spec requires the date-time format `YYYY-MM-DDTHH:MM:SSZ` (whole seconds, UTC, `Z` offset; for example `1985-10-21T01:21:00Z`).

## Root cause

`parse_datetime` in `tuftool/src/datetime.rs` is the shared `value_parser` for every subcommand's `expires` argument (`create`, `update`, `add-role`, `create-role`, `add-key-role`, `remove-role`, `remove-key-role`, `transfer-metadata`, `update-targets`) and for the `root expire` `time` argument. It accepted RFC 3339 input and returned it unchanged, so a fractional second such as `2026-01-01T00:00:00.5Z` flowed straight through to the `expires` field. The relative `"in N days"` branch had the same issue because `Utc::now()` always carries a sub-second component.

`expires` is typed as `chrono::DateTime<Utc>` and serializes with chrono's default `Serialize` impl (there is no custom serde on that field), which renders the fractional second. A quick check confirms it:

- `serde_json::to_string` of the parsed `2026-01-01T00:00:00.5Z` yields `"2026-01-01T00:00:00.500Z"` (non-compliant)
- after truncation it yields `"2026-01-01T00:00:00Z"` (compliant)

So the value has to be truncated before it reaches the metadata; doing it at parse time is the single choke point that covers all of the subcommands above.

## Change

Truncate the parsed datetime to whole seconds with `with_nanosecond(0)` in both branches of `parse_datetime`. This mirrors the existing `round_time` helper in `tuftool/src/root.rs`, which already applies the same `with_nanosecond(0)` truncation to the `init`/`expire` defaults but not to the user-supplied value that went through `parse_datetime`.

Input that is already whole-second precision is unchanged. Non-UTC offsets continue to be normalized to UTC as before, just without the sub-second part.

## Tests

Added unit tests in `datetime.rs`:

- sub-second RFC 3339 input is truncated and serializes without a fractional second
- whole-second input round-trips unchanged
- a non-UTC offset with a sub-second component normalizes to whole-second UTC
- the relative `"in N days"` branch produces a whole-second value

Verification on `develop`:

- `cargo test -p tuftool --bin tuftool` - 20 passed, 0 failed (16 pre-existing + 4 new)
- `cargo fmt -p tuftool --check` - clean
- `cargo clippy -p tuftool --all-targets` - no warnings

I left the `tuftool/CHANGELOG.md` untouched to keep the diff focused, since the existing entries are grouped under released versions and reference a merged PR number. Happy to add a changelog line under the next version if you would prefer it in this PR.

### DIFF
--- a/tuftool/src/datetime.rs
+++ b/tuftool/src/datetime.rs
@@ -7,11 +7,16 @@ use jiff::{SignedDuration, Timestamp};
 use snafu::{ensure, ResultExt};
 
 /// Parses a user-specified datetime, either in full RFC 3339 format, or a shorthand like "in 7
-/// days"
+/// days".
+///
+/// The TUF specification requires metadata `expires` timestamps in whole seconds, UTC, with no
+/// sub-second component (`YYYY-MM-DDTHH:MM:SSZ`). RFC 3339 input can carry fractional seconds (for
+/// example `2030-01-01T00:00:00.5Z`), and `Timestamp::now()` always has a sub-second component, so
+/// the returned value is truncated to whole seconds.
 pub(crate) fn parse_datetime(input: &str) -> Result<Timestamp> {
     // If the user gave an absolute date in a standard format, accept it.
     if let Ok(ts) = input.parse::<Timestamp>() {
-        return Ok(ts);
+        return Ok(truncate_to_seconds(ts));
     }
 
     // Otherwise, pull apart a request like "in 5 days" to get an exact datetime.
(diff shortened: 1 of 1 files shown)


arpitjain099:
Rebased onto develop. Since the datetime module moved from chrono to jiff in the meantime, I re-implemented the truncation on jiff: a small `truncate_to_seconds` (via `Timestamp::from_second(ts.as_second())`) applied to both the parsed-RFC3339 and the relative ("in N days") paths, since `Timestamp::now()` and fractional RFC3339 input still carry sub-second precision that the TUF `expires` format doesn't allow. Added unit tests for the fractional, whole-second, and relative cases; `cargo test -p tuftool datetime` passes.

Issue #942, opened 2026-05-20
Consider replacing chrono with jiff?


ww-oai:
Hi folks,

We're looking at integrating `tough` into uv, by way of `sigstore-rust` (which has various crates, some of which depend on tough for TUF operations against the Sigstore root of trust).

As part of that integration, we're looking for ways to reduce our dependency tree/de-proliferate our transitive deps. We're currently avoiding a dependency on `chrono` entirely, but tough introduces it.

So, the request: would you all consider (optionally) replacing your dependency on `chrono` with `jiff`? This has some direct advantages on your end (a smaller and more misuse-resistant API), but admittedly it's primarily an advantage for us as a downstream 😅. I can see at least two possible routes here, although others may exist:

1. Remove `chrono` entirely and replace it with `jiff`. This shouldn't be externally disruptive (FWICT, your usage of `chrono` is entirely internal), but it does mean changing every downstream's dependencies.
2. Keep `chrono` as a default-but-optional dependency, while allowing downstreams to opt into a `jiff` via a separate optional feature. This would allow uv (and other interested downstreams) to selecitvely opt into `jiff` while not disturbing the default resolution.

Please let me know if either of these sounds interesting or appealing to you all! I'm happy to implement whatever works best.

ww-oai:
(For context, https://github.com/sigstore/sigstore-rust/pull/90 is tracking a similar request, but one level up the dependency tree from uv's perspective.)

jmt-lab:
Hello there, thanks for reaching out. We understand the desire to reduce dependency trees and requirements. However, datetimes are a pretty critical part of TUF. While using jiff might just be drop in replacement with some minor changes, this sort of change would require a lot more of manual evaluation and testing on our side, so we cannot say whether or not we can do this soon. I do want to explore this and create an analysis of this.

CathalMullan:
Worth noting that `chrono` has been soft-deprecated, with `jiff` recommended as an alternative: https://github.com/chronotope/chrono/issues/1768

ww-oai:
Thanks @CathalMullan, I forgot to mention that! Indeed, I think that's worth noting.

(I'm happy to help with whatever validation would clear your bar here, since the use of chrono in tough is a blocker for our adoption. Anecdotally, I can say that uv's internal use of jiff has been rock-solid, and we similarly make use of time APIs in various load bearing parts of uv's functionality.)

woodruffw:
[...]

PR #948, opened 2026-06-03
feat!: migrate from chrono to jiff

jmt-lab:
*Issues*: #942 

*Description of changes:*
Due to chrono being soft deprecated, we need to migrate tough from chrono -> jiff. This PR performs this migration as a hard cutover. The commit is marked as breaking as it does cause API breakage for consumers who consume tough and still use chrono. Upon merge we should put a sticky issue calling this out.

*Testing done*

- [ ] Local twoliter that has also been migrated to jiff, building a variant and creating a tuf repo and publishing

Please comment with what more testing we should do here

### DIFF
--- a/tough/tests/repo_editor.rs
+++ b/tough/tests/repo_editor.rs
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
 use crate::test_utils::{days, dir_url, read_to_end, test_data};
-use chrono::Utc;
+use jiff::Timestamp;
 use std::collections::HashMap;
 use std::num::NonZeroU64;
 use std::path::PathBuf;
@@ -78,11 +78,11 @@ async fn load_tuf_reference_impl(paths: &mut RepoPaths) -> Repository {
 
 async fn test_repo_editor() -> RepositoryEditor {
     let root = root_path();
-    let timestamp_expiration = Utc::now().checked_add_signed(days(3)).unwrap();
+    let timestamp_expiration = Timestamp::now() + days(3);
     let timestamp_version = NonZeroU64::new(1234).unwrap();
-    let snapshot_expiration = Utc::now().checked_add_signed(days(21)).unwrap();
+    let snapshot_expiration = Timestamp::now() + days(21);
     let snapshot_version = NonZeroU64::new(5432).unwrap();
-    let targets_expiration = Utc::now().checked_add_signed(days(13)).unwrap();
+    let targets_expiration = Timestamp::now() + days(13);
     let targets_version = NonZeroU64::new(789).unwrap();
     let target3 = targets_path().join("file3.txt");
     let target_list = vec![target3];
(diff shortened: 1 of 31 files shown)


Sparksssj (approved):
LGTM!

tough/tests/repo_editor.rs
```diff
@@ -326,11 +326,11 @@ async fn create_role_flow() {
...
     let root_key = key_path();
     let key_source = LocalKeySource { path: root_key };
-    let timestamp_expiration = Utc::now().checked_add_signed(days(3)).unwrap();
+    let timestamp_expiration = Timestamp::now().checked_add(days(3)).unwrap();
     let timestamp_version = NonZeroU64::new(1234).unwrap();
-    let snapshot_expiration = Utc::now().checked_add_signed(days(21)).unwrap();
+    let snapshot_expiration = Timestamp::now().checked_add(days(21)).unwrap();
     let snapshot_version = NonZeroU64::new(5432).unwrap();
-    let targets_expiration = Utc::now().checked_add_signed(days(13)).unwrap();
+    let targets_expiration = Timestamp::now().checked_add(days(13)).unwrap();
```
BurntSushi: You should just be able to do `Timestamp::now() + days(13)` here. (And for other places too.) I'd only suggest it where you're already doing `timestamp.checked_add(...).unwrap()`.

tough/tests/target_path_safety.rs
```diff
@@ -18,8 +18,8 @@ use tough::{Prefix, RepositoryLoader, TargetName};
 /// Returns a date in the future when Rust programs will no longer exist. `MAX_DATETIME` is so huge
 /// that it serializes to something weird-looking, so we use something that is recognizable to
 /// humans as a date.
-fn later() -> DateTime<Utc> {
-    Utc.with_ymd_and_hms(2999, 1, 1, 0, 0, 0).unwrap()
+fn later() -> Timestamp {
+    "2999-01-01T00:00:00Z".parse::<Timestamp>().unwrap()
```
BurntSushi: If you want to avoid parsing here, I'd do `civil::date(2999, 1, 1).to_zoned(TimeZone::UTC).unwrap().timestamp()`. It's a little more wordy than `chrono`, but it's pretty intentional.

tuftool/src/datetime.rs
```diff
@@ -43,24 +39,9 @@ pub(crate) fn parse_datetime(input: &str) -> Result<DateTime<Utc>> {
...
-        }
-        "week" | "weeks" => {
-            TimeDelta::try_weeks(i64::from(count)).context(error::DateArgInvalidSnafu {
-                input: count.to_string(),
-                msg: format!("unable to convert {count} to a number of weeks"),
-            })?
-        }
+        "hour" | "hours" => SignedDuration::from_hours(i64::from(count)),
+        "day" | "days" => SignedDuration::from_hours(i64::from(count) * 24),
+        "week" | "weeks" => SignedDuration::from_hours(i64::from(count) * 24 * 7),
```
BurntSushi: Note that your multiplications here for days and weeks can overflow and panic depending on how the binary is built. Previously you would return an error.

tuftool/src/root.rs
```diff
@@ -429,9 +429,9 @@ impl Command {
     }
 }
 
-fn round_time(time: DateTime<Utc>) -> DateTime<Utc> {
-    // `Timelike::with_nanosecond` returns None only when passed a value >= 2_000_000_000
-    time.with_nanosecond(0).unwrap()
+fn round_time(time: Timestamp) -> Timestamp {
+    // Truncate sub-second precision; jiff guarantees `from_second(as_second())` round-trips.
+    Timestamp::from_second(time.as_second()).unwrap()
```
BurntSushi, line 434: I guess the [second example here](https://docs.rs/jiff/latest/jiff/struct.TimestampRound.html#example) is the more declarative approach, but I agree your approach here is also correct and just fine.

BurntSushi (commented):
I found one bug I think, but this otherwise looks good to me!

jmt-lab:
Sorry for taking so long on this, we had some emergent priorities come up that got in the way.

sky1122 (approved):

